PR #4, opened 2016-01-02
EKF: Fix 'Dizzy on Startup' behaviour

priseborough:
Fixes incorrect initialisation of vehicle orientation that contributed to large angular HUD movement for the first few seconds after startup.

Uses the gravity vector to estimate the initial roll and pitch angle
Uses the projection of the magnetic field measurement onto the earth axis horizontal plane to calculate the initial heading

TODO: implement an alignment state machine that does tilt -> yaw -> GPS with health checks at each state transition.


LorenzMeier:
Thanks!

PR #9, opened 2016-01-09
added interface for parameters

tumbili:


EKF/estimator_base.cpp
```diff
@@ -259,27 +258,6 @@ void EstimatorBase::initialiseVariables(uint64_t time_usec)
 	_state.quat_nominal.setZero();
 	_state.quat_nominal(0) = 1.0f;
 
-	_params.mag_delay_ms = 0;
```
LorenzMeier: Why are you now not initialising the values to sane values? The filter as a software component is unusable without those sane defaults.

EKF/estimator_base.h
```diff
@@ -58,6 +58,33 @@ struct gps_message {
 	bool vel_ned_valid;		// GPS ground speed is valid
 };
 
+struct parameters {
+	float mag_delay_ms;
```
LorenzMeier: They should be all initialized to sane defaults. Try if you can do that here, C++11 added support for some of that.

tumbili:
@LorenzMeier I added default values for the parameters. This should be good to go.

PR #12, opened 2016-01-20
fix buffer overflow in constrain of wind velocity

nickolasrossi:


tumbili:
@nickolasrossi Many thanks!

PR #14, opened 2016-01-25
Add comprehensive user selectable GPS checks

priseborough:


EKF/gps_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+    double lat = gps->lat * 1.0e-7;
+    double lon = gps->lon * 1.0e-7;
+    if (_posRef.init_done) {
+        map_projection_project(&_posRef, lat, lon, &deltaPosN, &deltaPosE);
+    } else {
+        map_projection_init(&_posRef, lat, lon);
+        _gps_alt_ref = gps->alt * 1e-3f;
+    }
+
+    // Calculate time lapsesd since last update, limit to prevent numerical errors and calculate the lowpass filter coefficient
```
tumbili: typo

EKF/estimator_base.h
```diff
@@ -320,5 +341,25 @@ class EstimatorBase
 
 	uint64_t _last_gps_origin_time_us = 0;
 	struct map_projection_reference_s _posRef = {};
-	float _gps_alt_ref = 0.0f;
+    float _gps_alt_ref = 0.0f;
```
tumbili: @priseborough Do those variables need to be public. Otherwise I would move them up. They are a bit lonely down there.

EKF/estimator_base.cpp
```diff
@@ -170,7 +171,7 @@ void EstimatorBase::setGpsData(uint64_t time_usec, struct gps_message *gps)
 		gps_sample_new.time_us -= FILTER_UPDATE_PERRIOD_MS * 1000 / 2;
 		_time_last_gps = time_usec;
 
-		_last_valid_gps_time_us = hrt_absolute_time();
+        _last_valid_gps_time_us = _time_last_imu;
```
tumbili: @priseborough Do we really need this? We already have _time_last_gps above. The two variables have exactly the same value I believe.

tumbili:
@priseborough I reviewed this and it looks great! Please see my comments.

EKF/gps_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+    float velN = fminf(fmaxf(deltaPosN / dt, -velLimit), velLimit);
+    float velE = fminf(fmaxf(deltaPosE / dt, -velLimit), velLimit);
+
+    // Apply a low pass filter
+    _gpsDriftVelN = velN * filterCoef + _gpsDriftVelN * (1.0f - filterCoef);
+    _gpsDriftVelE = velE * filterCoef + _gpsDriftVelE * (1.0f - filterCoef);
+
+    // Calculate the horizontal drift speed and fail if too high
+    // This check can only be used if the vehicle is stationary during alignment
+    if(_vehicleArmed) {
```
tumbili: @priseborough Sorry to bother you about this but I think we should agree on code style. I tried to used google coding style until now
https://google.github.io/styleguide/cppguide.html#Variable_Names
Maybe we can discuss this sometime it's probably a good idea to do this at this stage.

EKF/estimator_base.cpp
priseborough: Agree. I'll fix it.

EKF/estimator_base.h
priseborough: These are is used in ekf2_main. I'll comment the altitude reference

EKF/gps_checks.cpp
priseborough: Will fix

EKF/gps_checks.cpp
priseborough: Will change naming convention to match previous

priseborough:
These changes have been made in response to review comments

PR #17, opened 2016-01-27
EKF: Add magnetometer fusion error handling

priseborough:
Modifies the magnetometer 3-axis fusion so that the 3 components are fused on the one time step. This is necessary because all three components need to pass innovation consistency checks before sequential fusion is performed. This means the magnetometer fusion is no longer spread out across three time steps which improves numerical stability, but increases timing jitter.

Adds a fusion gate parameter for the 3-axis fusion and fixes a scaling error on the exisiting heading fusion gate.

Adds a numerical error check to the 3-axis fusion innovation variance calculations.

Indoor flight test with3-axis fusion here: http://logs.uaventure.com/view/qwawJmkkUMbKJWm3uttDzc

Magnetomter innovation plots (note innovation offsets at start and end of flight when taking off and landing on the steel reinforced floor):
mag innov x
mag innov y
mag innov z

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -49,7 +49,6 @@ Ekf::Ekf():
 	_fuse_pos(false),
 	_fuse_hor_vel(false),
 	_fuse_vert_vel(false),
-	_mag_fuse_index(0),
 	_time_last_fake_gps(0),
 	_vel_pos_innov{},
 	_mag_innov{},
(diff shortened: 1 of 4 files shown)


LorenzMeier:
@priseborough @tumbili I assume you will work out the PRs between yourselves?

priseborough:
That's the plan

LorenzMeier:
FYI: I did review the change on the logic, results and architecture side and it looks good. I can't speak to the derivation though.

priseborough:
The mathematics of the Kalman gains and observation Jacobians should be unchanged. Unfortunately the diff tool doesn't reflect this. I think I will do a separate text compare with the original to make sure that no cut and paste errors have been introduced (unlikely given block text operations, but not impossible).

PR #28, opened 2016-01-29
Ekf Multiple Instance

bugobliterator:
Just started to invoke discussions.


priseborough:
I think we can close this now that the restructure has gone in

PR #39, opened 2016-02-06
Doxygen Documentation - Added doxy style documentation to header files in the EKF Directory

mcsauder:
Hello,

I am very excited to have discovered your Estimation Control Library.

In reading through the files I've found that I don't understand a few of the function argument variables, so I thought a good place to start might be to add as much commenting as I could, then submit this pull request to see if the commenting I've added is correct and ask the authors to help me finish the commenting.

If this is a good start and deemed worthwhile, I can continue similar work in the other files.  If this kind of documentation is unwanted or if there is a different documentation style preferred that I could help with please let me know what is preferred, give me an example to work from and I will get back to work. 

Thanks for reviewing my PR!

-Mark


mcsauder:
I've gone ahead and finished adding doxygen style comments to the remaining header files in the "ecl" directory and I've placed @TODO's everywhere a bit of additional description remains needed.

Hope this PR can be useful; if you like this modification, excellent, and if not please let me know if there is anything I can do to improve things.

Thanks!

-Mark

priseborough:
I believe this is something that should be done and having doxygen compatible headers is part of the PX4 style guide recommendations. The question is when.  The code will not be feature complete for multi-rotor use until end of Q1 and for all other use cases by end Q2. The next few weeks will see a fair amount of churn in the code. There is an argument that documenting now and updating as we go will be better in the longer term. There is also an argument that it is better to wait until the churn slows before doing this.

@bugobliterator  and @tumbili what do you think?

mcsauder:
Hi @priseborough,

I can certainly appreciate the concern you raise.  If the group would like, I could close this PR and resubmit it later when the dust settles a bit.  I'll be keeping up with the work as best I can in the meantime.  Let me know what is the most useful and I'll help with what I can!

Thanks!

-Mark

PR #41, opened 2016-02-08
EKF2: Prevent declination/heading drift during non-GPS operation

priseborough:
This PP adds a method to fuse the declination which allows 3-axis fusion of magnetometer measurements to be used when there are no external measurements to make the heading and declination observable. It has been flight tested with and without GPS aiding.

Magnetometer fusion results with GPs aiding - note the published declination value at the test location is 11.5 deg
mag dec with gps
mag bias with gps
mag innov with gps

Magnetometer fusion results without  GPS aiding. The estimated declination stays around the parameter value of 0 deg, but the magnetometer bias offsets are still learned
mag dec without gps
mag bias without gps
mag innov without gps

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -110,12 +110,13 @@ void Ekf::controlFusionModes()
 		}
 	}
 
-        // if we are using 3-axis magnetometer fusion, but without external aiding, then the declination needs to be fused as an observation to prevent long term heading drift
-        if(_control_status.flags.mag_3D && _control_status.flags.gps) {
-            _control_status.flags.mag_dec = false;
-        } else {
-            _control_status.flags.mag_dec = true;
-        }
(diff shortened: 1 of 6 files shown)


EKF/control.cpp
```diff
@@ -109,6 +109,14 @@ void Ekf::controlFusionModes()
 			_control_status.flags.mag_3D = false;
 		}
 	}
+
+	// if we are using 3-axis magnetometer fusion, but without external aiding, then the declination needs to be fused as an observation to prevent long term heading drift
+	if (_control_status.flags.mag_3D && _control_status.flags.gps) {
+		_control_status.flags.mag_dec = false;
+
+	} else {
```
tumbili: Do not fuse declination if we don't fuse 3D mag.

tumbili:
@priseborough Reviewed, looks great!

PR #42, opened 2016-02-09
EKF: Improve control of magnetometer fusion

priseborough:
Allow the different magnetometer fusion methods to be selected (heading or 3-axis). The default behaviour is auto selection. This is controlled by the EKF2_MAG_TYPE parameter.

Allow user to specify use of declination from geo_lookup library or EKF2 parameter. The default behaviour is to use the declination from the lookup. This is controlled by the EKF2_DECL_TYPE bitmask parameter.

Allow the value obtained from the table lookup to be saved to EKF2_MAG_DECL parameter in non-volatile memory when the EKF origin is initialised. The default behaviour is to save the parameter for the next start-up, this can be disabled via a parameter setting. This is controlled by the EKF2_DECL_TYPE bitmask parameter.

### DIFF
--- a/EKF/gps_checks.cpp
+++ b/EKF/gps_checks.cpp
@@ -40,6 +40,7 @@
  */
 
 #include "ekf.h"
+#include <mathlib/mathlib.h>
 
 // GPS pre-flight check bit locations
 #define MASK_GPS_NSATS  (1<<0)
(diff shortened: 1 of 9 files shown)


EKF/gps_checks.cpp
```diff
@@ -66,6 +67,13 @@ bool Ekf::collect_gps(uint64_t time_usec, struct gps_message *gps)
 			_gps_alt_ref = gps->alt / 1e3f;
 			_NED_origin_initialised = true;
 			_last_gps_origin_time_us = _time_last_imu;
+			// set the magnetic declination returned by the geo library using the current GPS position
+			_mag_declination_gps = math::radians(get_mag_declination(lat, lon));
+
+			// save the declination to the relevant parameter
+			if (_params.mag_declination_source & MASK_SAVE_GEO_DECL) {
+				param_set(param_find("EKF2_MAG_DECL"), &_mag_declination_gps);
```
tumbili: @priseborough I can do an interface for setting and saving params. We probably don't want to have these dependencies in here

priseborough: OK - that would be good.

priseborough:
There are been instances of instability of magnetometer fusion after arming introduced by these patches which need to be investigated.

PR #43, opened 2016-02-10
Correct for sensor noise and baro offset during alignment

priseborough:
The baro is averaged during alignment and an offset calculated that is used subsequently to ensure that the reported height of the EKF origin is correct, even for in-air GPs gain and that the filter correctly reports both absolute and relative height.

Some additional averaging is used during initialisation of the states to improve accuracy.

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -196,73 +196,89 @@ bool Ekf::update()
 
 bool Ekf::initialiseFilter(void)
 {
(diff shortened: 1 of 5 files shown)


priseborough:
I think the height reset needs to be pulled out from the position reset function as we should not be doing a height reset when transitioning to GPS aiding

LorenzMeier:
Thanks!

PR #45, opened 2016-02-12
moved fuse function to ekf_helper.cpp

CarlOlsson:
Moved it since it is a general function used not only for velocity and position fusion


tumbili:
@CarlOlsson Thanks!

PR #48, opened 2016-02-13
Major cleanup (mostly comments)

tumbili:
-removed unused print functions
-removed false _imu_time_last variable (correct is _time_last_imu)
-added lots of comments

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -82,13 +82,8 @@ public:
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


tumbili:
@priseborough I would like to merge this before anything else so that we can finally get in.
Quick look would be appreciated.

tumbili:
@bugobliterator FYI

tumbili:
Merged!

PR #49, opened 2016-02-16
EKF: Fix magnetometer fusion and covariance handling

priseborough:
Poor heading stability when stationary for longer time periods had been observed. During the investigation a number of deficiencies were identified:

1) A bug in the original derivation that meant a tan function was being used instead of an atan function in the derivation of the observation Jacobians and Kalman gains for the direct heading and declination fusion.
2) The covariance matrix terms for inactive states were not being explicitly controlled
3) The various fusion operations did not have explicit protections against modifiying inactive states
4) The covariance terms for inactive states were not being explicitly set when changing modes
5) The method of preventing uncontrolled covariance growth on magnetic field states relied on hard variance constraints and did not control the covariance terms. This could result in numerical instability

These deficiencies have been rectified, and placeholder to enable management of wind state covariance terms has been added to reduce the likelihood of similar errors being introduced when airspeed and sideslip fusion is added.

Ground testing has been completed and flight testing will be completed shortly  for operation using the three available magnetometer fusion mode parameter settings. The following figure shows the declination for the estimated NE field after 10 minutes of static operation with noisy GPS. The declination drift is acceptable, so the declination fusion during aiding has been turned off by default (it can be turned on by setting EKF2_DECL_TYPE bit position 2 to true).

without declination fusion

The following figures are taken from a flight with EKF2_MAG_TYPE = 1 which turns off use of the magnetic field states and uses a simple magnetic heading observation model. and velocity and position innovations were zero mean throughout with no unexpected vehicle movement observed during the flight.

GPS velocity and position innovations remained within limits and were centred about zero other than towards the end of flight when it was flown into an area with significant GPS masking and the GPS solution was glitching

heading fusion - vel innov

heading fusion - pos innov

yaw gyro bias and scale factor variation was normal

heading fusion - dangz bias

heading fusion - dangz scale

The following figures were taken with EKF2_MAG_TYPE = 2 which forces use of the 3-axis magnetometer fusion and EKF2_DECL_TYPE = 3 which means that fusion of the declination was not performed after GPS fusion started. This was done to test the stability of the earth field estimation. The flight was performed in a continuous hover with occasional yaw orientation changes to try and maximise declination drift.

The estimated declination during the flight was relatively constant with abut 5deg of change at the end of flight when the yaw orientation was changed.

mag3d fusion - declination

Velocity, position and mag innovations were normal

mag3d fusion - vel innov

mag3d fusion - pos innov

mag3d fusion - mag innov

IMU Z gyro bias and scale factor variation was normal

[...]

PR #50, opened 2016-02-16
Address C99 compatibility issues by relocating variable initialization.

mcsauder:
Hello, 

This pull request is being submitted to provide C99 compatibility so that C11 incompatible microprocessor IDEs are able to compile the library.

Please let me know if you have any comments/questions/concerns and I will work to address them.  Thanks! 

-Mark

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -141,7 +141,7 @@ struct parameters {
 	float mag_declination_deg = 0.0f;   // magnetic declination in degrees
 	float heading_innov_gate = 3.0f;    // heading fusion innovation consistency gate size in standard deviations
 	float mag_innov_gate = 3.0f;        // magnetometer fusion innovation consistency gate size in standard deviations
-	int mag_declination_source = 7;     // bitmask used to control the handling of declination data
+	int mag_declination_source = 3;     // bitmask used to control the handling of declination data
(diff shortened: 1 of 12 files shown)


priseborough:
Mark, we were about to pull in PR #49, which will create a number of conflicts with PR #50.

I was assessing the option of doing #50 first and tried rebasing #49 onto #50, but got into a mess as I didn't understand the criteria you were using when deciding which variables needed to be initialised by the constructor. 

If you think rebasing PR#49 on PR#50 is less work than vice-versa and are prepared to help sort out the conflicts, then let us know. Otherwise we will pull in PR#49 first.

mcsauder:
Hi Paul,

Thanks so much!  No worries, I'm happy to comb through and tidy things up in this pull request after you bring in PR#49!  I will make another check at that point to also verify that we've gotten all variables initialized and haven't missed any.  Thanks for all of the work you do!

-Mark

mcsauder:
Hi Paul and others,

Paul were right, this one did turn into a bit of mess... The only files we needed to change from current master were estimator_interface.cpp/h, ekf.cpp/h., and common.h, but it looks like my attempt at syncing/rebasing the fork from upstream/master didn't turn out so well. 

I think its probably best just to close this PR and start a new one so that it is more apparent what is being requested for change.

-Mark

PR #51, opened 2016-02-17
Optflow Support

bugobliterator:
Current working PX4 Firmware branch:
https://github.com/bugobliterator/PX4Firmware/tree/pr-optflowekf

Basic testing pointers:
optflow sensor (PX4Flow) +Y axis should point in the +X axis of Controller board
On the Firmware side I have disabled use of PX4Flow's distance sensor, but you can remove the commit. Just make sure you have precise range measurement, as this PR doesn't have terrain estimate.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -119,11 +119,31 @@ struct flowSample {
 	uint64_t    time_us;
 };
 
+enum fusion_mode_t {
+	FUSE_GPS,
+	FUSE_OF,
+	FUSE_GPS_OF,
+	FUSE_NONE
+};
+
+enum vdist_sensor_type_t {
+	VDIST_SENSOR_RANGE,
+	VDIST_SENSOR_BARO,
+	VDIST_SENSOR_GPS,
+	VDIST_SENSOR_NONE
+};
+
 struct parameters {
+	int fusion_mode = FUSE_OF;
+	int vdist_sensor_type = VDIST_SENSOR_RANGE;
+	float rng_gnd_clearance = 0.1f;
+
 	float mag_delay_ms = 0.0f;          // magnetometer measurement delay relative to the IMU
 	float baro_delay_ms = 0.0f;         // barometer height measurement delay relative to the IMU
 	float gps_delay_ms = 200.0f;        // GPS measurement delay relative to the IMU
 	float airspeed_delay_ms = 200.0f;   // airspeed measurement delay relative to the IMU
+	float flow_delay_ms = 200.0f;
+	float range_delay_ms = 200.0f;
 
 	// input noise
 	float gyro_noise = 1.0e-3f;         // IMU angular rate noise used for covariance prediction
@@ -141,8 +161,12 @@ struct parameters {
 	float pos_noaid_noise = 10.0f;      // observation noise for non-aiding position fusion
 	float baro_noise = 3.0f;            // observation noise for barometric height fusion
 	float baro_innov_gate = 3.0f;       // barometric height innovation consistency gate size in standard deviations
+	float range_noise = 0.1f;
+	float flow_innov_gate = 3.0f;
+	float flow_noise = 0.001f;
 	float posNE_innov_gate = 3.0f;      // GPS horizontal position innovation consistency gate size in standard deviations
 	float vel_innov_gate = 3.0f;        // GPS velocity innovation consistency gate size in standard deviations
+	float range_innov_gate = 5.0f;
 
 	float mag_heading_noise = 1.7e-1f;  // measurement noise used for simple heading fusion
 	float mag_noise = 5.0e-2f;          // measurement noise used for 3-axis magnetoemeter fusion
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -113,6 +113,7 @@ struct airspeedSample {
 };
 
 struct flowSample {
+	uint8_t quality;
 	Vector2f    flowRadXY;
 	Vector2f    flowRadXYcomp;
 	uint64_t    time_us;
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -63,6 +63,13 @@ typedef matrix::Vector<float, 3> Vector3f;
 typedef matrix::Quaternion<float> Quaternion;
 typedef matrix::Matrix<float, 3, 3> Matrix3f;
 
+struct flow_message {
+	uint8_t quality;			// Quality of Flow data
+	Vector2f flowdata;			// Flow data received
+	Vector2f gyrodata;			// Gyro data from flow sensor
+	uint32_t dt;				// integration time of flow samples
+};
+
 struct outputSample {
 	Quaternion  quat_nominal;	// nominal quaternion describing vehicle attitude
 	Vector3f    vel;	// NED velocity estimate in earth frame in m/s
--- a/EKF/optflow_fusion.cpp
+++ b/EKF/optflow_fusion.cpp
@@ -198,7 +198,18 @@ void Ekf::fuseOptFlow()
 		}
 		// by definition our error state is zero at the time of fusion
 		_state.ang_error.setZero();
-		
+	    // only update magnetic field states if we are fusing 3-axis observations
+        if (!_control_status.flags.mag_3D) {
+            for (int row = 16; row <= 21; row++) {
+                Kfusion[row] = 0.0f;
+            }
+        }
+        // only update wind states if we are doing wind estimation
+        if (!_control_status.flags.wind) {
+            for (int row = 22; row <= 23; row++) {
+                Kfusion[row] = 0.0f;
+            }
+        }	
         fuse(Kfusion,_flow_innov(index));
 
 		Quaternion q_correction;
--- a/EKF/optflow_fusion.cpp
+++ b/EKF/optflow_fusion.cpp
@@ -41,7 +41,7 @@
  */
 
 #include "ekf.h"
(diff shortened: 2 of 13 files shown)


EKF/optflow_fusion.cpp
```diff
@@ -0,0 +1,249 @@
...
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ ****************************************************************************/
+
+/**
+ * @file vel_pos_fusion.cpp
```
tumbili: wrong file

EKF/optflow_fusion.cpp
```diff
@@ -0,0 +1,249 @@
...
+ * Function for fusing gps and baro measurements/
```
tumbili: Also this

EKF/common.h
```diff
@@ -133,8 +161,12 @@ struct parameters {
 	float pos_noaid_noise = 10.0f;      // observation noise for non-aiding position fusion
 	float baro_noise = 3.0f;            // observation noise for barometric height fusion
 	float baro_innov_gate = 3.0f;       // barometric height innovation consistency gate size in standard deviations
+	float range_noise = 0.1f;
+	float flow_innov_gate = 3.0f;
+	float flow_noise = 0.001f;
```
priseborough: If this is a 1-sigma noise and not a variance, then it is far too small for a default parameter setting and could drive the filter unstable or cause rejection of flow measurements.

EKF/optflow_fusion.cpp
```diff
@@ -0,0 +1,280 @@
...
... - SH_LOS[0]*SH_LOS[3]*(sq(q0) - sq(q1) + sq(q2) - sq(q3))*(P[8][4]*SH_LOS[0]*SH_LOS[2]*sq(SH_LOS[3]) - P[0][4]*(SH_LOS[0]*SH_LOS[3]*SH_LOS[4] - SH_LOS[2]*SH_LOS[3]*SH_LOS[6]) + P[3][4]*SH_LOS[0]*SH_LOS[3]*(2*q0*q3 - 2*q1*q2) + P[1][4]*SH_LOS[2]*SH_LOS[3]*SH_LOS[5] + P[2][4]*SH_LOS[0]*SH_LOS[1]*SH_LOS[3] - P[5][4]*SH_LOS[0]*SH_LOS[3]*SH_LOS[6] - P[4][4]*SH_LOS[0]*SH_LOS[3]*(sq(q0) - sq(q1) + sq(q2) - sq(q3))) + SH_LOS[0]*SH_LOS[2]*sq(SH_LOS[3])*(P[8][8]*SH_LOS[0]*SH_LOS[2]*sq(SH_LOS[3]) - P[0][8]*(SH_LOS[0]*SH_LOS[3]*SH_LOS[4] - SH_LOS[2]*SH_LOS[3]*SH_LOS[6]) + P[3][8]*SH_LOS[0]*SH_LOS[3]*(2*q0*q3 - 2*q1*q2) + P[1][8]*SH_LOS[2]*SH_LOS[3]*SH_LOS[5] + P[2][8]*SH_LOS[0]*SH_LOS[1]*SH_LOS[3] - P[5][8]*SH_LOS[0]*SH_LOS[3]*SH_LOS[6] - P[4][8]*SH_LOS[0]*SH_LOS[3]*(sq(q0) - sq(q1) + sq(q2) - sq(q3))) + SH_LOS[0]*SH_LOS[3]*(2*q0*q3 - 2*q1*q2)*(P[8][3]*SH_LOS[0]*SH_LOS[2]*sq(SH_LOS[3]) - P[0][3]*(SH_LOS[0]*SH_LOS[3]*SH_LOS[4] - SH_LOS[2]*SH_LOS[3]*SH_LOS[6]) + P[3][3]*SH_LOS[0]*SH_LOS[3]*(2*q0*q3 - 2*q1*q2) + P[1][3]*SH_LOS[2]*SH_LOS[3]*SH_LOS[5] + P[2][3]*SH_LOS[0]*SH_LOS[1]*SH_LOS[3] - P[5][3]*SH_LOS[0]*SH_LOS[3]*SH_LOS[6] - P[4][3]*SH_LOS[0]*SH_LOS[3]*(sq(q0) - sq(q1) + sq(q2) - sq(q3))));
```
priseborough: Split this calculation so that we calculate the innovation variance, then ensure that it is >=R_LOS before calculating the inverse. if it isn't, the the covariance matrix has contributed a negative component and the calculation is badly conditioned. if this occurs, then the covariance matrix will need to be reset with no fusion peerformed on that time step

priseborough: Split this calculation so that we calculate the innovation variance, then ensure that it is >=R_LOS before calculating the inverse. It is possible that the innovation variance could be zero at this step.

EKF/optflow_fusion.cpp
```diff
@@ -0,0 +1,280 @@
...
+			Kfusion[14] = SK_LOS[0]*(P[14][8]*SH_LOS[0]*SH_LOS[2]*sq(SK_LOS[4]) - P[14][0]*(SH_LOS[0]*SH_LOS[4]*SK_LOS[4] - SH_LOS[2]*SH_LOS[6]*SK_LOS[4]) + P[14][3]*SH_LOS[0]*SK_LOS[4]*(2*q0*q3 - 2*q1*q2) + P[14][2]*SH_LOS[0]*SH_LOS[1]*SK_LOS[4] + P[14][1]*SH_LOS[2]*SH_LOS[5]*SK_LOS[4] - P[14][5]*SH_LOS[0]*SH_LOS[6]*SK_LOS[4] - P[14][4]*SH_LOS[0]*SK_LOS[3]*SK_LOS[4]);
+			Kfusion[15] = SK_LOS[0]*(P[15][8]*SH_LOS[0]*SH_LOS[2]*sq(SK_LOS[4]) - P[15][0]*(SH_LOS[0]*SH_LOS[4]*SK_LOS[4] - SH_LOS[2]*SH_LOS[6]*SK_LOS[4]) + P[15][3]*SH_LOS[0]*SK_LOS[4]*(2*q0*q3 - 2*q1*q2) + P[15][2]*SH_LOS[0]*SH_LOS[1]*SK_LOS[4] + P[15][1]*SH_LOS[2]*SH_LOS[5]*SK_LOS[4] - P[15][5]*SH_LOS[0]*SH_LOS[6]*SK_LOS[4] - P[15][4]*SH_LOS[0]*SK_LOS[3]*SK_LOS[4]);
+			Kfusion[16] = SK_LOS[0]*(P[16][8]*SH_LOS[0]*SH_LOS[2]*sq(SK_LOS[4]) - P[16][0]*(SH_LOS[0]*SH_LOS[4]*SK_LOS[4] - SH_LOS[2]*SH_LOS[6]*SK_LOS[4]) + P[16][3]*SH_LOS[0]*SK_LOS[4]*(2*q0*q3 - 2*q1*q2) + P[16][2]*SH_LOS[0]*SH_LOS[1]*SK_LOS[4] + P[16][1]*SH_LOS[2]*SH_LOS[5]*SK_LOS[4] - P[16][5]*SH_LOS[0]*SH_LOS[6]*SK_LOS[4] - P[16][4]*SH_LOS[0]*SK_LOS[3]*SK_LOS[4]);
```
priseborough: The Kalman gain should be set to zero for states that are not active. For example if we are not doing 3axis mag fusion, then Kfusion[16] to Kfusion[21] should be set to zero to ensure that the corresponding terms in the covariance matrix are not inadvertently modified.

EKF/optflow_fusion.cpp
```diff
@@ -0,0 +1,280 @@
...
+	float q3 = _state.quat_nominal(3);
+
+	// get latest velocity in earth frame
+	float vn = _state.vel(0);
+	float ve = _state.vel(1);
+	float vd = _state.vel(2);
+	float pd = _state.pos(2);
+	Vector3f vel_body;
+
+	float R_LOS = sq(_params.flow_noise);
```
priseborough: This parameter needs to have a safe lower limit enforced. i would recommend a lower limit of 0.05 rad initially.

EKF/common.h
```diff
@@ -106,16 +113,37 @@ struct airspeedSample {
...
 struct parameters {
+	int fusion_mode = FUSE_OF;
+	int vdist_sensor_type = VDIST_SENSOR_RANGE;
+	float rng_gnd_clearance = 0.1f;
+
 	float mag_delay_ms = 0.0f;          // magnetometer measurement delay relative to the IMU
 	float baro_delay_ms = 0.0f;         // barometer height measurement delay relative to the IMU
 	float gps_delay_ms = 200.0f;        // GPS measurement delay relative to the IMU
 	float airspeed_delay_ms = 200.0f;   // airspeed measurement delay relative to the IMU
+	float flow_delay_ms = 200.0f;
```
priseborough: How did we arrive at a delay of 0.2sec? It seems high.

priseborough:
Replaced by https://github.com/PX4/ecl/pull/74

PR #54, opened 2016-02-18
Address C99 compatibility issues by relocating variable initialization. - Replaces PR#50

mcsauder:
Hello,

This pull request replaces PR#50. 

It is submitted to make provision for C99 compatibility such that C11 incompatible microprocessor IDEs may compile the library.

Thanks for your feedback!

-Mark

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -31,8 +31,6 @@
  *
  ****************************************************************************/
 
-#include "Matrix.hpp"
-
 /**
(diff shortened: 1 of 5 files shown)


mcsauder:
The latest commit added a number of variable initializations to the EstimatorInterface constructor.  The single variable that causes runtime errors when it is not initialized in the constructors is: 

    _inititialised(false)

-Mark

PR #56, opened 2016-02-23
EKF: Fix bad behaviour when aligning from inverted

priseborough:
The  previous code did not work when inverted and would /0 when pitch = +-90 deg


priseborough:
Needs to be included with some additional commits to provide a more reliable initial alignment

PR #57, opened 2016-02-23
EKF: Improvements to accuracy reporting and alignment

priseborough:
Note: needs to used in conjunction with https://github.com/PX4/Firmware/pull/3826 for full benefit

Use reported GPS accuracy to set observation noise, but keeps noise between limits set by parameters. This reduces the sensitivity to variations in GPS accuracy

Adjust initial diagonal elements in the covariance matrix to provide a more consistent alignment

Publish the EKF horizontal and vertical position 1-sigma error estimates

Monitor tilt error decay during alignment and don't align yaw and declare the EKF output as valid until tilt errors have decayed.

Change to use of the most recent IMU data from the buffer to do the initial tilt alignment because the non-buffered data being used was vulnerable to a race condition that could make it zero.

Fix a bug that caused an incorrect sign in initial roll when aligning from inverted.

Alignment has been tested from a number of starting orientations - upright and inverted. A single flight has been performed to verify that the normal post alignment prediction and fusion functions operate as expected. The position innovation variances can now be seen to adjust as the GPS EPH increases.

GPS EPH and EPV
gps accuracy

GPS position innovations
pos innov

GPS velocity innovations
vel innov

Magnetometer innovations
mag innov

Barometer innovations
baro innov

A bench test with large orientation changes and GPS signal blockage leading to large  glitches was performed to check the GPS error scaling. The position and velocity innovations for that test are shown here.

pos innov - bad gps

vel innov - bad gps

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -155,8 +155,8 @@ struct parameters {
 	float req_gdop;         // maximum acceptable geometric dilution of precision
 	float req_hdrift;       // maximum acceptable horizontal drift speed
 	float req_vdrift;       // maximum acceptable vertical drift speed
-	
-    // Initialize parameter values.  Initialization must be accomplished in the constructor to allow C99 compiler compatibility.
+
+	// Initialize parameter values.  Initialization must be accomplished in the constructor to allow C99 compiler compatibility.
 	parameters()
 	{
 		mag_delay_ms = 0.0f;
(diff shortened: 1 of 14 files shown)


tumbili:
@priseborough Tested initialising inverted while shaking, seemed to handle it nicely!

priseborough:
Travis is unhappy about a mathlib include, so i will take a look at that tomorrow

bugobliterator:
@priseborough use "mathlib.h" directly instead, after rebasing on top of latest master

priseborough:
I've rebased on master and fixed the Travis message

priseborough:
Chasing up an observed behaviour with horizon errors during bench testing - will re-open when resolved

priseborough:
I will break this up into two PR's - one for the essential alignment fixes, and one for the status reporting.

PR #59, opened 2016-02-24
Improvements to EKF startup - partially replaces PR#57

priseborough:
Fixes a bug that causes alignment failure when inverted
Improves initial covariance values for smoother alignment


priseborough:
Thats a dumb error from Travis.

PR #62, opened 2016-02-25
use in_air flag from landing detector

tumbili:
This enables the use of the PX4 landing detector to get information if the vehicle is in the ground or not. This feature is only available if we are using gps.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -198,12 +198,13 @@ void Ekf::calculateVehicleStatus()
 	}
 
 	// Transition to in-air occurs when armed and when altitude has increased sufficiently from the altitude at arming
-	if (!_control_status.flags.in_air && _control_status.flags.armed && (_state.pos(2) - _last_disarmed_posD) < -1.0f) {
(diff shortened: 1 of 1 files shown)


tumbili:
The figure below shows the results of an indoor flight without gps. The ekf uses heading fusion until it switches to full mag fusion when it's in the air. Since in this case there was no gps, the change in vertical position was used to detect in air state.

in_air

priseborough:
It is important that the switch to 3D mag fusion occurs sufficiently high from the ground so that the effect of the ground level magnetic interference is reduced to a level that will not cause 'toilet bowling' on a copter. Tests taking off from a car roof wiht EKF1 on APM showed that to be about 1.5m, and EKF2 is less sensitive to alignment errors, so the height increase has been set to 1.0m.

What height does the in-air flag trigger?

tumbili:
@priseborough You made me think and I think I'd rather just use the landing detector to force heading fusion when it says "landed". But I don't want to use it for switching from heading to full mag fusion.
Will correct that.

tumbili:
Merged.

PR #63, opened 2016-02-25
Attitude reset after heading reset

tumbili:
When doing a test flight in the lab with the Pixracer using latest ecl master the quad suddenly twitched some seconds after it was in the air. The log showed that the attitude jumped considerable (see picture below). This jump occurred when the heading reset happened (switch from heading fusion to 3D mag fusion). I believe that the problem was that we were setting the output attitude equal to the yaw corrected ekf attitude which is of course delayed and thus now necessarily a good choice.


tumbili:
bad_att_init

priseborough:
Yes, that makes sense. if the attitude was varying at the time of the reset, then we are setting roll and pitch to a value that is old. The solution in the is PR is the simplest way around the issue and will let the output observer smooth out the step in yaw angle.

PR #64, opened 2016-02-25
Additional filter variable initializations required to fully reset the filter variables

mcsauder:
Hi there,

There were a few additional variable re-initializations required in order to get everything restarted without NaN's if the filter happens to diverge.


priseborough:
There are likely other variables that will also need to be initialised for reliable re-starts, so I need to do an audit, however this is a good start.

PR #65, opened 2016-02-26
Move variable initialization from header file to constructor for C99 compilers

mcsauder:
This PR follows PR #54 for new variables added since #54.


tumbili:
Merged, thanks!

PR #67, opened 2016-02-29
EKF2: Reduce effect of magnetic disturbances on tilt

priseborough:
Thes changes add a method to fuse in magnetic heading using both a 321 and 312 Euler yaw observation and automatically chooses the best option to eliminate problems associated with use of an Euler angle observation when the second angle in the rotation set is near +-90 degrees.

This has enabled the simple yaw fusion method to be the default method for on-ground operation, which significantly reduces the effect of magnetic field disturbances on the roll and pitch angle.

The 3-axis fusion remains as the default method for up and away flight.

TODO - the required 312 Euler to DCM and DCM to Euler conversions were not available in the library so have been inlined. They will need to be added to the ECL math library.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -183,7 +183,7 @@ struct parameters {
 		posNE_innov_gate = 3.0f;
 		vel_innov_gate = 3.0f;
 
-		mag_heading_noise = 1.7e-1f;
+		mag_heading_noise = 5.0e-1f;
 		mag_noise = 5.0e-2f;
 		mag_declination_deg = 0.0f;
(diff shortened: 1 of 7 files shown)


priseborough:
Found a bug in the 312 yaw calculation that needs to be fixed

priseborough:
Uneventful bench test and in-door flight test with EKF2_MAG_TYPE = 1

priseborough:
The code has been cleaned up and the various patches consolidated. Rather than force push I have opened another PR.

PR #70, opened 2016-03-01
EKF: Reduce effect of magnetic field disturbances on roll and pitch

priseborough:
**Information**

Replaces #67 

The previous default method of fusing magnetometer data when on the ground used the azimuth angle of the projection of the measured field onto the horizontal. Unfortunately this method did not decouple the observation from the roll and pitch angle in the filter observation equations with the result that magnetic field interference could create large errors in the estimated roll and pitch.

The alternative provided as an option was fusion of the Euler yaw angle obtained from a separate magnetic compass calculation which had much smaller levels of roll and pitch error due to magnetic field disturbances. However this could not be used as a default method because it suffered from singularities at +-90 deg pitch.

The solution was to enable fusion of magnetic heading using either a 321 and 312 Euler yaw observation and automatically chooses the best option to eliminate problems associated with use of an Euler angle observation when the second angle in the rotation set is near +-90 degrees.

This has enabled the simple yaw fusion method to be the default method for on-ground operation, which significantly reduces the effect of magnetic field disturbances on the roll and pitch angle.

The 3-axis fusion remains as the default method for up and away flight.

The required 312 Euler to DCM and DCM to Euler conversions were not available in the library so have been inlined. They will need to be added to the ECL math library.

** Test Results**

The susceptibility to magnetic interference was tested on the bench with the board stationary and a magnetised steel rod used to create large levels of magnetic interference. The 2D fusion as can be seen next suffers from large level of coupling between magnetic field disturbance and roll and pitch errors:

mag_2d

The new heading fusion method reduced the disturbances by an order of magnitude:

mag_hdg

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -183,7 +183,7 @@ struct parameters {
 		posNE_innov_gate = 3.0f;
 		vel_innov_gate = 3.0f;
 
-		mag_heading_noise = 1.7e-1f;
+		mag_heading_noise = 5.0e-1f;
 		mag_noise = 5.0e-2f;
 		mag_declination_deg = 0.0f;
 		heading_innov_gate = 3.0f;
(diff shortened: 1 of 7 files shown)


tumbili:
Merged!

PR #71, opened 2016-03-02
initialize ECL_L1 variables

dagar:
This is a possible fix for https://github.com/PX4/Firmware/issues/3902, but should be good regardless.


tumbili:
Reviewed and merged!

PR #72, opened 2016-03-04
Fix bug in declination fusion and standardise covariance update

priseborough:
This also moves the large 24x24 arrays to static memory as they are used by all fusion processes.
TODO - move covariance updates for all fusion processes into a function that uses a mask to indicate when elements in the observation Jacobian are to be used.


priseborough:
Flight testing completed

PR #74, opened 2016-03-07
Add fusion of optical flow LOS rate measurements

priseborough:
Replaces #51 

Uses a new formulation for the optical flow Jacobians that makes the partial derivative from vertical position to optical flow zero. This improves the stability of the estimates.

Adds a simple terrain vertical position estimator to improve robustness to range finder glitches. Has been flight tested  in both GPS and OF mode.

Outputs the distance to terrain using the local position topic. Although use of range finder for the primary height measurement is available as an option, this has been turned off by default as moving the NED origin up and down with the terrain can have adverse effects on angle estimates. If terrain following is required, then it is recommended that the control loops use the distance to ground estimate published as part of the vehicle local position.

Enables the simultaneous use of GPS and optical flow to be tested.

Must be used in conjunction with https://github.com/priseborough/Firmware/tree/pr-optflowekf-rebased . By default the optical flow mode is turned off. To test optical flow in isolation, set EKF2_AID_MASK to 2.

**Test Results - EKF2_AID_MASK = 2 (Optical Flow)**

Optical Flow Innovations (indoor flight over level floor and obstacles up to 70cm height)

flow x innov
flow y innov

Range finder innvovations and local position ground clearance estimate (indoor flight over level floor and obstacles  up to 70cm height)

hagl innovations
hagl estimate

**Flight Test Results - EKF2_AID_MASK = 3 (GPS & Optical Flow)**

This test was conducted outdoors at low altitude across grass and concrete surfaces.
GPS was poor due to interference from the range finder and flow sensor. This caused height disturbances due to errors in the vertical GPS velocity.
Later in the flight the vehicle was flown at speed and low altitude to deliberately induce false readigs in the flow sensor. This can be seen in the innovations.
Attitude control and position control remained stable.

Optical Flow Innovations
flow x innov
flow y innov

GPS Position Innovations
pos innov

GPS Velocity Innovations
vel innov
vel z innov

Range Finder innovations
hagl innov

Baro innovations
pos z innov


priseborough:
This has been flight tested with OF only, GPS only and OF+GPS.

PR #78, opened 2016-03-16
EKF: Add option to use GPS for height and improve height fall-back behaviour

priseborough:
- Add GPS accuracy data to GPS buffer so that the accuracy and measurement information used by various processes is from the same point in time.
- Define maximum measurement intervals for different sensor types so that timeouts can be detected using consistent thresholds
- Use a combined time and height uncertainty criteria to determine if the height state needs to be reset
- Maintain a filtered baro height correction so that the baro data can be used as a fall-back if either range or GPS height fusion fails
- Use  low pass filtered data during alignment instead of averaged data to reduce the effect of time delays and allow for longer averaging periods when using range finder data.
- Implement a consistent fall-back behaviour following a height fusion timeout.

**Flight Test Results**

http://logs.uaventure.com/view/7Cc2Jtrrarxc7RKCqfFX4Y

A flight test was conducted indoors with EKF2_HGT_MODE = 1 to use GPs altitude. The board used was a pixracer. The rc.logging file was modified to that data logging started in power-up. Three flights were performed. Flight 1 started before the GPS origin was set, so the origin was set and the switch to GPS height use occurred in-flight.
The copter was landed and disarmed and the GPs connector was disconnected which caused it to revert to Baro height and static mode for position aiding. Flight 2 was performed without the GPs connected, the copter was landed, disarmed and the GPS connector reattached.  This caused the EKF to switch back to GPs aiding for position and velocity, but remain with Baro aiding as per the design intent.

Innovation Plots

There were two large glitches in baro measurements shortly after power-up - the cause of these glitches is not known and they have not been observed in other logs. The switch to GPS height and back again was smooth

pos d innov
vel d innov
pos ne innov
vel ne innov

Local and Global Height

height

NE Position - note the divergence in position after GPs loss which continues until the filter times-out and switches to a non-GPs mode. Note the reset back to the correct position when GPs is re-acquired.

pos ne

NE Position 1-Sigma Uncertainty - note the growth in estimated error which matches the actual error growth

pos ne uncertainty

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -158,10 +158,85 @@ void Ekf::controlFusionModes()
 		}
 	}
 
-	// Handle the case where we have rejected height measurements for an extended period
-	// This excessive vibration levels can cause this so a reset gives the filter a chance to recover
-	// After 10 seconds without aiding we reset to the height measurement
-	if (_time_last_imu - _time_last_hgt_fuse > 10e6) {
+	/*
+	 * Handle the case where we have not fused height measurements recently and
+	 * uncertainty exceeds the max allowable. Reset using the best available height
+	 * measurement source, continue using it after the reset and declare the current
+	 * source failed if we have switched.
+	*/
+	if ((P[8][8] > sq(_params.hgt_reset_lim)) && ((_time_last_imu - _time_last_hgt_fuse) > 5e6)) {
+		// handle the case where we are using baro for height
+		if (_control_status.flags.baro_hgt) {
+			// check if GPS height is available
+			gpsSample gps_init = _gps_buffer.get_newest();
+			bool gps_hgt_available = ((_time_last_imu - gps_init.time_us) < 2 * GPS_MAX_INTERVAL);
+			bool gps_hgt_accurate = (gps_init.vacc < _params.req_vacc);
+			baroSample baro_init = _baro_buffer.get_newest();
+			bool baro_hgt_available = ((_time_last_imu - baro_init.time_us) < 2 * BARO_MAX_INTERVAL);
+
+			// use the gps if it is accurate or there is no baro data available
+			if (gps_hgt_available && (gps_hgt_accurate || !baro_hgt_available)) {
+				// declare the baro as unhealthy
+				_baro_hgt_faulty = true;
+				// set the height mode to the GPS
+				_control_status.flags.baro_hgt = false;
+				_control_status.flags.gps_hgt = true;
+				_control_status.flags.rng_hgt = false;
+				// adjust the height offset so we can use the GPS
+				_hgt_sensor_offset = _state.pos(2) + gps_init.hgt - _gps_alt_ref;
+				printf("EKF baro hgt timeout - switching to gps\n");
+			}
+		}
+
+		// handle the case we are using GPS for height
+		if (_control_status.flags.gps_hgt) {
+			// check if GPS height is available
+			gpsSample gps_init = _gps_buffer.get_newest();
+			bool gps_hgt_available = ((_time_last_imu - gps_init.time_us) < 2 * GPS_MAX_INTERVAL);
+			bool gps_hgt_accurate = (gps_init.vacc < _params.req_vacc);
+			// check the baro height source for consistency and freshness
+			baroSample baro_init = _baro_buffer.get_newest();
+			bool baro_data_fresh = ((_time_last_imu - baro_init.time_us) < 2 * BARO_MAX_INTERVAL);
+			float baro_innov = _state.pos(2) - (_hgt_sensor_offset - baro_init.hgt + _baro_hgt_offset);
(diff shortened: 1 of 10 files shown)


EKF/control.cpp
```diff
@@ -158,10 +158,85 @@ void Ekf::controlFusionModes()
...
+	 * uncertainty exceeds the max allowable. Reset using the best available height
+	 * measurement source, continue using it after the reset and declare the current
+	 * source failed if we have switched.
+	*/
+	if ((P[8][8] > sq(_params.hgt_reset_lim)) && ((_time_last_imu - _time_last_hgt_fuse) > 5e6)) {
+		// handle the case where we are using baro for height
+		if (_control_status.flags.baro_hgt) {
+			// check if GPS height is available
+			gpsSample gps_init = _gps_buffer.get_newest();
+			bool gps_hgt_available = ((_time_last_imu - gps_init.time_us) < 2 * GPS_MAX_INTERVAL);
```
julianoes: Why do you add a factor of 2 before the MAX_INTERVAL. Isn't the max already saying that it is a threshold?

EKF/control.cpp
```diff
@@ -158,10 +158,85 @@ void Ekf::controlFusionModes()
...
+			if (gps_hgt_available && (gps_hgt_accurate || !baro_hgt_available)) {
+				// declare the baro as unhealthy
+				_baro_hgt_faulty = true;
+				// set the height mode to the GPS
+				_control_status.flags.baro_hgt = false;
+				_control_status.flags.gps_hgt = true;
+				_control_status.flags.rng_hgt = false;
+				// adjust the height offset so we can use the GPS
+				_hgt_sensor_offset = _state.pos(2) + gps_init.hgt - _gps_alt_ref;
+				printf("EKF baro hgt timeout - switching to gps\n");
```
julianoes: This is a rather critical warning. Is there a chance to do a ```mavlink_log_critical()```? Have you done it anywhere else in this module?

EKF/control.cpp
```diff
@@ -158,10 +158,85 @@ void Ekf::controlFusionModes()
...
+			rangeSample rng_init = _range_buffer.get_newest();
+			bool rng_data_available = ((_time_last_imu - rng_init.time_us) < 2 * RNG_MAX_INTERVAL);
+			// check if baro data is available
+			baroSample baro_init = _baro_buffer.get_newest();
+			bool baro_data_available = ((_time_last_imu - baro_init.time_us) < 2 * BARO_MAX_INTERVAL);
+			// check if baro data is consistent
+			float baro_innov = _state.pos(2) - (_hgt_sensor_offset - baro_init.hgt + _baro_hgt_offset);
+			bool baro_data_consistent = sq(baro_innov) < (sq(_params.baro_noise) + P[8][8]) * sq(_params.baro_innov_gate);
+			// switch to baro if necessary or preferable
+			bool switch_to_baro = (!rng_data_available && baro_data_available) || (baro_data_consistent && baro_data_available);
```
julianoes: I'm not sure about this check. Basically you always want to switch to the baro if the baro is ok, even if the range finder is still available?

julianoes:
I had a read through the code changes and made some comments here and there. However, I'm not really in the loop and haven't looked at the ekf2 before, so I can't really comment on logic or math. Just from reading the diff it didn't get how and when the logic of the switching works. 

How do you go about testing all the different cases? Can you deactivate sensors in SITL?

Also, I was a bit confused about what ```_control_status``` is. Is it just for logging, or where is it used?

EKF/control.cpp
priseborough: We allow for one missed message

EKF/control.cpp
priseborough: I'm not familiar wth mavlink_log_critical(). Printf statements are used in other places in the EKF code to assist with monitoring important events.

EKF/control.cpp
priseborough: The range data has proven to have poor reliability, so if it has tied out on the innovation consistency checks it is better to switch to using baro provided it is producing consistent data as attempting to reset the height to the range finder reading can produce large jumps in vertical position and the controllers are currently unable to accomodate large height resets.

Using range finder as the primary height reference has a very limited use case (primarily indoor operation over flat surfaces) and is generally not recommended. It would  be better to use baro as the height reference within the EKF and for the controller to use the height above ground estimate in the local position  message.

priseborough:
Thanks for having a look at it. Testing was done  by bench testing (disconnect range finder or disconnect GPS. I couldn’t find the functionality required in SITL to fail specific sensors like there is in APM.

The control _control status variable contains the status of the various filter modes (type of height sensor being used, aiding, magnetometer fusion control, etc). Most of the control loging is implemented here: https://github.com/PX4/ecl/blob/pr-addGpsHgtOption/EKF/control.cpp#L44

PR #79, opened 2016-03-25
fw multiply integrator gain in loop

dagar:
Otherwise the transition can be quite abrupt when tuning.


dagar:
Can we merge this? I can create a before/after log if desired.

dagar:
@tumbili Can we merge this? I can create a before/after log if desired.

tumbili:
@dagar Thanks, looks like that I gain was there for nothing :-D

tumbili:
@dagar Could you do a PR for the PX4 side as well?

dagar:
Sure, do you just keep it pointed at the latest commit? Do you perform any basic testing for accepting a new version of ecl on the PX4 side?

PR #82, opened 2016-04-03
activate airspeed fusion

CarlOlsson:
Fuse airspeed if the data has fallen behind the fusion time horizon, we are airborne and the measured airspeed is above a defined threshold, now hardcoded to 5 m/s


CarlOlsson:
This pr is not relevant anymore

PR #83, opened 2016-04-04
Fix bugs in covariance prediction and alignment

priseborough:
Fixes an error in the covariance prediction derivation that meant the IMU noise parameter was not being squared when generating the control input variance matrix. This change has required the filter to be retuned.

Fixes a bug in the initialisation of states that could cause the zero height reference to be offset.

**Test Results**

The changes were tested in replay using data logged on a snapdragon board. The replay was performed using EKF2_HGT_MODE = 1 and with EKF2_MAG_TYPE = 1 for the first test and EKF2_MAG_TYPE = 2 for the second test

The log used by the replay is here:

http://logs.uaventure.com/view/wnC4hNng8PyWcJQtbcCAc5

**EKF2_MAG_TYPE = 1 (mag heading fusion)**

velned
posned
dvel bias
dang scale
dang bias
velne innov
yaw innov
posne innov
vert innov

**EKF2_MAG_TYPE = 2 (3-axis mag fusion)**

velne innov
vert innov
posne innov
posd innov
magx innov
magz innov
magy innov
magy bias
magned
magxyz bias
dvel bias
dang scale
dang bias
velned
posned


tumbili:
http://logs.uaventure.com/view/Fti2MySJja93ho76NTsx73

tumbili:
@priseborough Looks good flies good but I think something with init on Snapdragon is not yet ok.

priseborough:
The reason for the bad height init with Snapdragon was that the data buffers were not being initialised correctly. I have included a patch that memsets the buffers to zero immediately after they are created. My snapdragon board now initialises local position altitude close to zero each startup.

tumbili:
@priseborough Cool, thanks! Will give it a last test and then merge!

tumbili:
Merged

PR #88, opened 2016-04-06
ekf2 airspeed fusion:

tumbili:
- finished logic for fusion
- fixed bug where previous control status was set in the wrong location

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -113,6 +113,7 @@ struct rangeSample {
 
 struct airspeedSample {
 	float       true_airspeed;	// true airspeed measurement in m/s
+	float 		eas2tas;		// equivalent to true airspeed factor 
 	uint64_t    time_us;	// timestamp in microseconds
 };
 
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -43,6 +43,8 @@
 
 void Ekf::controlFusionModes()
 {
+	// Store the status to enable change detection
+	_control_status_prev.value = _control_status.value;
 	// Determine the vehicle status
 	calculateVehicleStatus();
 
@@ -339,14 +341,13 @@ void Ekf::controlFusionModes()
 		_control_status.flags.rng_hgt = true;
 	}
 
-	// Placeholder for control of wind velocity states estimation
-	// TODO add methods for true airspeed and/or sidelsip fusion or some type of drag force measurement
-	if (false) {
+	// if the airspeed measurements have timed out for 10 seconds we declare the wind estimate to be invalid
+	if (_time_last_imu - _time_last_arsp_fuse > 10e6 || _time_last_arsp_fuse == 0) {
 		_control_status.flags.wind = false;
+	} else {
+		_control_status.flags.wind = true;
 	}
 
-	// Store the status to enable change detection
-	_control_status_prev.value = _control_status.value;
 }
 
 void Ekf::calculateVehicleStatus()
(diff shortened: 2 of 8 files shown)


EKF/common.h
```diff
@@ -113,6 +113,7 @@ struct rangeSample {
 
 struct airspeedSample {
 	float       true_airspeed;	// true airspeed measurement in m/s
+	float 		eas2tas;		// equivalent to true airspeed factor
```
tumbili: @priseborough @CarlOlsson I added the eas2tas factor to the airspeed sample

EKF/control.cpp
```diff
@@ -341,14 +343,13 @@ void Ekf::controlFusionModes()
 		_control_status.flags.rng_hgt = true;
 	}
 
-	// Placeholder for control of wind velocity states estimation
-	// TODO add methods for true airspeed and/or sidelsip fusion or some type of drag force measurement
-	if (false) {
+	// if the airspeed measurements have timed out for 10 seconds we declare the wind estimate to be invalid
```
tumbili: @priseborough @CarlOlsson At the moment I have a hardcoded factor of 10 seconds for the airspeed sensor to timout until we declare wind states to be invalid.

EKF/covariance.cpp
```diff
@@ -172,7 +172,7 @@ void Ekf::predictCovariance()
 	float wind_vel_sig;
 
 	// Don't continue to grow wind velocity state variances if they are becoming too large or we are not using wind velocity states as this can make the covariance matrix badly conditioned
```
tumbili: @priseborough What do you think, should we increase wind variances even if our wind flag is set to false? Will it lead to a badly conditioned matrix?

EKF/covariance.cpp
```diff
@@ -641,10 +641,47 @@ void Ekf::predictCovariance()
 
 	// Don't do covariance prediction on wind states unless we are using them
 	if (_control_status.flags.wind) {
```
tumbili: @priseborough This is temporary for testing, no idea if an initial estimate of wind states based on an airspeed measurement makes sense. We could improve this by not assuming zero side slip.

EKF/covariance.cpp
```diff
@@ -725,6 +762,9 @@ void Ekf::predictCovariance()
 
 	// copy variances (diagonals)
 	for (unsigned i = 0; i < _k_num_states; i++) {
```
tumbili: @priseborough @CarlOlsson Do we really want to write nothing back here?

EKF/control.cpp
```diff
@@ -43,6 +43,8 @@
 
 void Ekf::controlFusionModes()
 {
+	// Store the status to enable change detection
```
tumbili: @priseborough I believe there was a bug in the old code where the old status flags were set to the current status flags at the end of the control file which does not make any sense in my opinion. Do you agree or what what the intention to do this?

tumbili:
@priseborough It would be nice if you could have a look at the comments and give me feedback.
I would also like to talk about covariance init / reset/ growth handling with you.

LorenzMeier:
@priseborough @tumbili Can we finish this up soonish so we can enter flight testing with this?

priseborough:
The PR as it is is safe' to go in IMO, however for flight testing to be useful we need to firstly verify that the EKF setup works acceptably with airspeed fusion turned off and that we log replay both EST and RPL logs  from the flights so that problems can be solved off-line. The problem is that flight logs from this branch I have been sent to date were for a setup with high vibration levels and absence of RPL messages in the log.

tumbili:
@LorenzMeier @priseborough Flight testing of this branch have until now revealed other bugs which we are fixing at the moment. As soon as we have fixed all issues I can do a replay log.

tumbili:
@LorenzMeier @priseborough Merged, but fusion is not activated yet.

Issue #91, opened 2016-04-11
EKF2 vertical height diverged


tumbili:
Happened on a test flight with wing wing and Pixracer

http://logs.uaventure.com/view/EDXp7BmVQsSsboydZMKVWW

First insights from discussion with @priseborough:
- [ ] when local position state is reset the corresponding velocity states needs to be reset as well (not the case now)
- [ ] the high innovations and the resulting divergence of the vertical position could have been caused by timing issues (need to check the performance results)

priseborough:
https://github.com/PX4/ecl/pull/92

priseborough:
I have tested its ability to recover from temporary large errors on IMU z accel data. However we need to give more thought to the desired system behaviour during a height sensor timeout and immediately following a reset.

priseborough:
I have incorporated furthers changes to height reset behaviour here and am i the process of testing it:

https://github.com/priseborough/ecl/tree/pr-ekf2HeightReset

We need a means of inducing sensor faults in SITL

PR #92, opened 2016-04-11
EKF: Fix vulnerability in height reset

priseborough:
Addresses https://github.com/PX4/ecl/issues/91

If bad inertial data causes a large vertical velocity, resetting the height without also resetting the vertical velocity is ineffective, because the bad velocity will pull the height away from the measurement resulting in repeating height resets.

This PR adds code to reset the vertical velocity to the GPS if available, following a vertical position reset.
The variances for the vertical position and velocity states are also reset using known sensor error characteristics.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -184,9 +184,7 @@ void Ekf::controlFusionModes()
 				_control_status.flags.rng_hgt = false;
 				// adjust the height offset so we can use the GPS
 				_hgt_sensor_offset = _state.pos(2) + gps_init.hgt - _gps_alt_ref;
-				if (!baro_hgt_available) {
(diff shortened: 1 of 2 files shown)


priseborough:
I will pull this in now as it fixes a major vulnerability.

The EKF needs to do a reset without offsetting the measurement as is currently done to avoid large height jumps as this will be problematic for flight that relies on the absolute height for safe operation. It is proposed that this be handled by:

1) Publishing the amount that the vertical position and velocity changed during the last height reset and the time of the reset.
2) In the shorter term correcting the output for the reset on the PX4/Firmware side of the interface in the ekf2 module. In the longer term this should be handled within the controllers by changing the set-point if a relative height without jumps is important or not at all if absolute height accuracy is important.

PR #98, opened 2016-04-19
EKF: improve reporting of filter status

priseborough:


EKF/ekf_helper.cpp
```diff
@@ -316,6 +316,12 @@ void Ekf::get_heading_innov_var(float *heading_innov_var)
 	memcpy(heading_innov_var, &_heading_innov_var, sizeof(float));
 }
 
+// get GPS check status
+void Ekf::get_gps_check_status(uint16_t *val)
+{
+	memcpy(val, &_gps_check_fail_status, sizeof(uint16_t));
```
LorenzMeier: the better solution here is: ```*val = _gps_check_fail_status```

priseborough: Done

priseborough:
Any reason why this should not be pulled in?

LorenzMeier:
No, let's do it!

PR #99, opened 2016-04-19
limit maximum rates of all measurement at which they are stored into …

tumbili:
…the buffers


tumbili:
@priseborough Good to go?

priseborough:
Looks good to me

tumbili:
Merged

PR #101, opened 2016-04-20
Pr vision support

devbharat:
[WIP] 
Tested to track external vision positions X Y and Z correctly. 

ekf2 x
ekf2 y
ekf2 z

Issue:
LPOS.Vz is constantly not 0 when standing but around -2.5 cm.

ekf2 vz

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -56,6 +56,7 @@ Ekf::Ekf():
 	_time_last_vel_fuse(0),
 	_time_last_hgt_fuse(0),
 	_time_last_of_fuse(0),
+	_time_last_arsp_fuse(0),
 	_last_disarmed_posD(0.0f),
 	_airspeed_innov(0.0f),
 	_airspeed_innov_var(0.0f),
@@ -326,7 +327,7 @@ bool Ekf::update()
 		}
 
 		// TODO This is just to get the logic inside but we will only start fusion once we tested this again
-		if (_airspeed_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_airspeed_sample_delayed) && false) {
+		if (_airspeed_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_airspeed_sample_delayed)) {
 			fuseAirspeed();
 		}
 	}
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -327,7 +327,8 @@ bool Ekf::update()
 		}
 
 		// TODO This is just to get the logic inside but we will only start fusion once we tested this again
-		if (_airspeed_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_airspeed_sample_delayed)) {
+		//if (_airspeed_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_airspeed_sample_delayed)) {
+		if (false) {
 			fuseAirspeed();
 		}
 	}
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -292,6 +292,13 @@ bool Ekf::update()
 			if (_control_status.flags.ev_pos) {
 				_fuse_pos = true;
 				_fuse_height = true;
+				
+				// correct position and height for offset relative to IMU
+				Vector3f pos_offset_body = _params.visn_pos_body - _params.imu_pos_body;
+				Vector3f pos_offset_earth = _R_to_earth * pos_offset_body;
+				_ev_sample_delayed.posNED(0) -= pos_offset_earth(0);
+				_ev_sample_delayed.posNED(1) -= pos_offset_earth(1);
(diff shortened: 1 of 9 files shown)


LorenzMeier:
@devbharat Could you propose a new message for a vision estimate including the required covariance so we can properly fuse things here?

devbharat:
Will do!

EKF/vel_pos_fusion.cpp
```diff
@@ -119,7 +120,7 @@ void Ekf::fuseVelPosHeight()
 	}
 
 	if (_fuse_height) {
-		if (_control_status.flags.baro_hgt) {
+		if (_control_status.flags.baro_hgt && !_control_status.flags.gps && !_control_status.flags.ev_pos && !(_control_status.flags.rng_hgt && (_R_to_earth(2, 2) > 0.7071f))) {
```
priseborough: Can you please explain what yo are trying to achieve here?

LorenzMeier:
@priseborough Note the vision / altitude fusion PR here

EKF/vel_pos_fusion.cpp
devbharat: Nowhere in control.cpp or ekf.cpp is the logic for setting "flags.baro_hgt = false" available yet when using external vision sensor for height. This, is if(baro height) else if(vision height) fails. 

The solution above gives higher priority to external vision height updates  when available over baro height, even though both "_control_status.flags.baro_hgt" and "_control_status.flags.ev_pos" are true.  This is tested to work, otherwise height updates from external  vision were neglected. 

A nice addition would be to reset the "_control_status.flags.baro_hgt" when external vision height is available.

devbharat:
@priseborough I am having trouble understanding why altitude rate( vz) is non-zero, even though altitude itself is constant! (see the Vz and LPOS.Z graphs above, vz stays at -2.5 cm/s when aircraft is nearly static ie hanging from strings)

priseborough:
@devbharat It is possibly due to IMU bias errors that are not learning fast enough. If you provide a log that includes all of the EST messages, i might be able to spot something.

priseborough:
Do you have the changes for PX4 Firmware that will enable this functionality to be tested in replay? This will be essential for tuning of the estimator.

devbharat:
@priseborough Should I mail you the log or would you like me to upload it somewhere ?

EKF/ekf.cpp
```diff
@@ -389,6 +389,8 @@ bool Ekf::initialiseFilter(void)
 		}
 	}
 
+	// warnx("_params.fusion_mode & MASK_USE_EVPOS %d", (int)(_params.fusion_mode & MASK_USE_EVPOS));
```
priseborough: Can you please replace these with print statements or something else that does not crete a dependance on the PX4 library. We are trying to keep ecl platform independant

EKF/vel_pos_fusion.cpp
```diff
@@ -153,6 +154,7 @@ void Ekf::fuseVelPosHeight()
 			// innovation gate size
 			gate_size[5] = fmaxf(_params.range_innov_gate, 1.0f);
 		} else if (_control_status.flags.ev_pos) {
+			// warnx("fusing ext_visn_hight");
```
priseborough: similarly here, using warnx does not allow us to be platform inpependant

EKF/ekf.cpp
devbharat: I am sorry, those weren't supposed to be in. I don't have a laptop now, but I'll clean it and push tomorrow morning.

EKF/vel_pos_fusion.cpp
priseborough: I would prefer to fix the logic  that causes flags.baro_hgt to not go false when using external vision. I will have a look at that and push a fix for you to test.

priseborough:
I have fixed the bug in the height source selection that was causing flags.baro_hgt to remain true after ev data arrived, removed the warnx statements and added a couple of prints to the console when it starts fusing EV data for position and yaw.

https://github.com/priseborough/ecl/tree/wingtra-pr-visionSupport

priseborough:
The log you uploaded to log muncher appears to have stalled
screen shot 2016-04-21 at 3 22 36 pm

devbharat:
@priseborough yes the log file, i thought it'll process it in some time but looks like not, but the raw log file can still be downloaded from the webpage. Also shared on skype!

priseborough:
@devbharat have you had a chance to check the fixes I pushed here?

https://github.com/priseborough/ecl/tree/wingtra-pr-visionSupport

priseborough:
@devbharat I have rebased on master and pushed a branch to origin with the height selection logic and status print fixes:

https://github.com/PX4/ecl/tree/pr-visionSupport

If this passes your testing and you have a tested PR for PX4/Firmware that includes the additional logging and replay for vision messages, then I will pull this branch in.

priseborough:
Reopening with modified and rebased branch

Issue #96, opened 2016-04-15
Height sensor offset resetting


tumbili:
After switching from baro to gps we initialize the primary height sensor offset (in this case gps) like this
https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L186

However, we should not be using the current height state value in this calculation.
@priseborough Please have a look.

LorenzMeier:
@priseborough Could you look into this during this week?

PR #102, opened 2016-04-20
EKF: Improve height reset behaviour

priseborough:
Addresses #96 and #91 

Has been flown in SITL for normal operation and bench tested for a GPS height timeout

Height reset recovery from each of the three modes (Baro, GPS and range finder) needs to by tested. I have tried using SITL, but it does not support adding offsets to sensor outputs.

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -160,6 +160,30 @@ public:
 		return _buffer[index];
 	}
 
+	// return data at the specified index
+	data_type get_from_index(unsigned index)
+	{
+		if (index >= _size) {
+			index = _size-1;
+		}
+		return _buffer[index];
+	}
+
+	// push data to the specified index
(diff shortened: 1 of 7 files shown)


priseborough:
Failed bench testing with a manual offset added to the Z accelerometer offset to induce a loss of height fusion.

tumbili:
@priseborough Looked at this and generally looks good. I haven't seen any use of "_hgt_sensor_offset", is that missing? I would expect this value to be set when we are switching between the height sensors to maintain consistency.

priseborough:
_hgt_sensor_offset is included  to allow for the height to be maintained after a reset if required. Currently there are no resets that use it, however I have left it in because the requirements for reset behaviour when using different sensors could change when we do more testing.

priseborough:
I have completed testing using the replay and confirmed sensible switching and reset behaviour when using GPS and baro.

PR #112, opened 2016-04-27
ekf: compile fix for Ubuntu 16.04

julianoes:
Another fix like https://github.com/PX4/Firmware/pull/4374.


priseborough:
What about that travis compiler error?

julianoes:
@priseborough this one is now fixed in my opinion.

PR #114, opened 2016-04-27
CMakeLists: use find_package Eigen3

julianoes:
Instead of hardcoding the Eigen3 path which only works on Mac, use
find_package(Eigen3).

Tested on Ubuntu 16.04.


julianoes:
Works in travis again. @priseborough let me know what you think.

priseborough:
@mcsauder Does this work for you?

mcsauder:
Hi @priseborough,

The short answer is yes, but only on my machine with a fresh install of 16.04.  On my machine that is upgraded from 15.10 as long as I keep the sympbolic links it works properly, but If I remove the sym links it remains broken like before.  Everything appears correct to me, so although I haven't discovered the remedy for an upgraded system without the sym links as opposed to a fresh install, I think the issue is with the OS upgrade, not with the PR.

My other projects were utilizing slightly different syntax in the CMakeLists.tx,

    SET(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
(as opposed to the current PR)

    SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR})
and I have tried both, each of which seem to perform identically, just with the same issue I have with the upgrade.

Now that I've verified it on the fresh install machine, I think this PR is the right way, and I'll continue using the sym-link work-around until I get a fresh install on the other machine as well.  

Thanks @julianoes for all of the work/feedback/tips!  Thanks @priseborough and everyone else for all of the work!!!  Awesome job everyone!

-Mark

julianoes:
@mcsauder: is the cmake script actually found and run, or is that the problem that you're having on 15.10?

mcsauder:
Hi @julianoes,

The cmake script is found, (and the compile script does progress as one would otherwise expect given the error), however the error I get without the sym links is:

    fatal error: Eigen/Dense: No such file or directory
    compilation terminated.

Again, I observe no errors with the machine with the 16.04 fresh install, only on the machine with the 15.10->16.04 upgrade.  It still has me scratching my head, but everything does compile fine for me with sym links in place on the upgraded machine.

(Also, when I have removed the sym links, I have followed with an apt-get remove libeigen3-dev, then apt-get install libeigen3-dev, then I have been checking the compile, which still fails, then I re-add the sym links again and compile passes.)  I think this must be an issue local to my machine somehow.  Please don't let it slow down the PR.  Please let me know if there is anything else I can check that might be useful to anyone else.

-Mark

julianoes:
@mcsauder: and I presume you tried to do a fresh cmake check using:
```
rm -rf Build
mkdir Build
cd Build
cmake ../EKF
```

What do you get if you add this in CMakeLists.txt after `find_package`:

```
message("Found Eigen3 include dir: ${EIGEN3_INCLUDE_DIR}")
```

mcsauder:
@julianoes, typically I've been using a make clean command:

    cd build/
    make clean
    cd ..

but I have also wiped the build directory manually as well just as you've described.

Thanks for asking about adding that line... it appears that might reveal an issue:

    Found Eigen3 include dir: /usr//usr/include/eigen3

notice the extra "/"

julianoes:
That looks odd. And that's with all symlinks removed? I wonder if this is the cmake's script's bug...

mcsauder:
Indeed, that's with the sym links removed.  With the sym links added there are no issues, just re-checked that, everything's compiling for me with the links.  I think this has to be an issue with my local machine that I can't explain... It wasn't there before upgrading, but it's not an issue on my other machine with a clean install... so weird! I think I am convinced I'll do a fresh install on this machine though. :)

Thanks for looking at all of that with me!

-Mark

julianoes:
Ok odd. Thanks for testing this though.

julianoes:
@LorenzMeier merge please if you agree with the change.

priseborough:
@LorenzMeier Bump

julianoes:
Fixed merge conflicts. Tested on Ubuntu 15.10. Good to merge.

julianoes:
Ok, checks fail now because `-Werror` is on.

julianoes:
@LorenzMeier I think this is good to merge now.

LorenzMeier:
Awesome, thanks!

PR #116, opened 2016-04-30
after a height reset zero the cross covariances for height and vertical

tumbili:
after a height reset zero the cross covariances for height and vertical velocity states. this should prevent accel bias and height states covariances
to get negative after a height reset.


EKF/ekf_helper.cpp
```diff
@@ -188,6 +188,12 @@ void Ekf::resetHeight()
 
 	}
 
+	// zero the cross covarinace entries for height and vertical velocity
```
priseborough: Row and column 8 should only be zeroed if vert_pos_reset is true

Similarly for Row and column 5 and vert_vel_reset.

priseborough:
i have reworked the patch.

PR #119, opened 2016-05-03
Changed airspeed innovation gate to 10

CarlOlsson:
Will test this later in the week

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -197,7 +197,7 @@ struct parameters {
 	int mag_fusion_type;		// integer used to specify the type of magnetometer fusion used
 
 	// airspeed fusion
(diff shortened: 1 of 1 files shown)


priseborough:
If the gate needs to be this big, then there is something wrong elsewhere in the filter tuning. We want this to be able to reject measurements if the pitot tube blocks and the airspeed goes close to zero. Changes to this and the other defaults airspeed fusion  parameters need to be validated using replay of flight logs with airspeed failure injected to ensure that it will not cause loss of position or attitude reference.

CarlOlsson:
Thanks for the comments, I changed it because I picked the first value arbitrary. What innovation gate would you start with as default?

priseborough:
5-sigma is a good starting point

LorenzMeier:
@CarlOlsson Could you please address the comments here or close?

CarlOlsson:
innovation gate changed to 5 and rebased

PR #87, opened 2016-04-04
removed reinitialization of intermediate variables

CarlOlsson:
They are already initialized in ekf.h

PR #120, opened 2016-05-04
enable Wshadow

dagar:
I'm trying to enable -Wshadow in PX4 and found another declaration of KH and KHP when I'm assuming the version in the Ekf class was intended to be used.

### DIFF
--- a/EKF/airspeed_fusion.cpp
+++ b/EKF/airspeed_fusion.cpp
@@ -196,9 +196,6 @@ void Ekf::fuseAirspeed()
 		_state.quat_nominal.normalize();
 
 		// update covariance matrix via Pnew = (I - KH)P = P - KHP
-		float KH[_k_num_states][_k_num_states] = {};
-		float KHP[_k_num_states][_k_num_states] = {};
-
 		for (unsigned row = 0; row < _k_num_states; row++) {
 			for (unsigned column = 0; column < _k_num_states; column++) { // Here it will be a lot of zeros, should optimize that...
 				KH[row][column] = Kfusion[row] * H_TAS[column];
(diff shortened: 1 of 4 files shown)


EKF/airspeed_fusion.cpp
```diff
@@ -196,9 +196,6 @@ void Ekf::fuseAirspeed()
 		_state.quat_nominal.normalize();
 
 		// update covariance matrix via Pnew = (I - KH)P = P - KHP
-		float KH[_k_num_states][_k_num_states] = {};
```
LorenzMeier: @CarlOlsson @tumbili This looks like Daniel fixed a real bug here - could you review this change?

EKF/vel_pos_fusion.cpp
```diff
@@ -246,8 +246,6 @@ void Ekf::fuseVelPosHeight()
 		_state.quat_nominal.normalize();
 
 		// update covarinace matrix via Pnew = (I - KH)P
-		float KHP[_k_num_states][_k_num_states] = {};
```
LorenzMeier: @priseborough This looks like it could have been an actual bug here.

LorenzMeier:
@dagar Excellent! Two of the changes look like actual potential bugs.

CarlOlsson:
Good catch!
It seems like there was a problem with the merging of [this](https://github.com/PX4/ecl/pull/87) pr?

EKF/vel_pos_fusion.cpp
priseborough: Agree, there is no need to declare a local variable for the Kalman gain and the class variable should be used. Note that changing this will have no effect on functionality as the Kalman gain does not need to be persistent.

EKF/airspeed_fusion.cpp
priseborough: There was no need to declare the Kalman gain as a local variable and the class variable should be used to be consistent with usage elsewhere. This will have no effect on the functionality as the Kalman gain does not need to be persistent. A class variable was used to avoid time lost allocating an array each time.

PR #123, opened 2016-05-08
added bad_airspeed reporting

CarlOlsson:
_fault_status.bad_airspeed was never set


EKF/airspeed_fusion.cpp
```diff
@@ -96,8 +96,10 @@ void Ekf::fuseAirspeed()
 
 		if (_airspeed_innov_var_temp >= R_TAS) { // Check for badly conditioned calculation
 			SK_TAS[0] = 1.0f / _airspeed_innov_var_temp;
+			_fault_status.bad_airspeed = false;
```
priseborough: This has been invalidated by recent changes as fault status is now a union so should be referenced as _fault_status.flags.bad_airspeed

LorenzMeier:
@CarlOlsson Can you please update?

CarlOlsson:
Adopted code and rebased

PR #127, opened 2016-05-13
Add Sideslip function

CarlOlsson:
Adds the sideslip function but no functionality regarding when to fuse it etc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ px4_add_module(
 		EKF/mag_fusion.cpp
 		EKF/gps_checks.cpp
 		EKF/airspeed_fusion.cpp
+		EKF/sideslip_fusion.cpp
 		EKF/optflow_fusion.cpp
 		EKF/control.cpp
 		EKF/terrain_estimator.cpp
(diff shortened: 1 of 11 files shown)


CarlOlsson:
@tumbili is this ok to merge? It just adds the sideslip function but not the logic, would be nice to merge it and avoid future rebasing

CarlOlsson:
Goes together with this
https://github.com/PX4/Firmware/pull/4481

LorenzMeier:
@priseborough Good to go?

priseborough:
Pull requests for that contain significant changes (which includes new fusion methods) should be accompanied with test evidence (flight or SITL logs).

CarlOlsson:
Hi Paul. I definitely agree, I will supply flight tests and post logs etc. However, since the function isn't called anywhere yet I thought that we could merge it even without that. I can also close this one and create a new when everything is ready if that is preferable?
And thanks for reviewing!

priseborough:
Carl, I would prefer to wait until we have something that can be tested. Let's organise a Skype call to discuss the control logic requirements.

CarlOlsson:
Ok perfect. Then I will close this pr and create a new cleaner one with logs + plots

Issue #125, opened 2016-05-10
Snapdragon: Poor altitude estimation with range sensor


ndepal:
After following the advice given in https://github.com/PX4/ecl/issues/111, we have come to the conclusion that altitude estimation performance of the complementary filter is still not satisfactory.

This can be seen in the following image:
![alt_est](https://cloud.githubusercontent.com/assets/8755300/15151138/4384a140-16d0-11e6-8220-21554184a8ea.png)

`EST0.s8` tracks the range data (the sign of which is inverted in the shown plot) well, while `LPOS.Z` does not. `LPOS.Z` often overshoots in tracking changes in altitude. At 55s, `LPOS.Z` does something completely different than `EST0.s8`. Between 20s and 40s, the quad is placed on the ground (the range sensor saturates at its minimum value) and is disarmed (thus, IMU values are not noisy). Still, `LPOS.Z` overshoots significantly and appears to saturate at an offset value.

The depicted log is here: http://logs.uaventure.com/view/CYBGUXRipknC9Rr5Ug9kN4

Here is another replay log of my current setup: http://logs.uaventure.com/view/KphLtTVUAttaQhPtUf5XBn

The firmware is on ffb0d37 with all submodules updated.

On top of that I have added a low pass filter for the acceleration measurements fed into the `control_status` with a cutoff frequency of 30Hz, as suggested in the above-mentioned issue. I have also added the `control_status` acc measurements to the log file. However, it appears as though we have an estimation problem and not (or not only) a control problem.

@priseborough can you help us with this?

priseborough:
Yes i can - I will do some analysis on the replay log.

priseborough:
I'm away on travel for the rest of the week, so will not be able to resume work on this until Sat

priseborough:
I have some requests for further tests:

1) Can you please provide a short replay log where the copter is placed upright and held still in a level orientation for 5 seconds and then rotated upside down and held still and level in an inverted orientation for another 5 seconds. It is possible that the accel calibration may not be correctly calibrating the delta velocity outputs.

2) Can you please generate a replay log where the copter is held at 50cm, 100cm and 150cm above the floor with the heights determined using a tape measure or other mechanical means. it is possible the range finder has a scale factor error.

3) Can you please test the following PR and report back: https://github.com/PX4/ecl/pull/129. This should improve the performance with inconsistent sensor data, but the root cause of the data inconsistency should still be determined.

ndepal:
[...]

PR #129, opened 2016-05-14
EKF:  Improve output observer position and velocity tracking

priseborough:
Enables smaller output complementary filer time constants to be used for position and velocity states in response to https://github.com/PX4/ecl/issues/125. 

Replace the delayed time feedback mechanism used by the translational states with a direct feedback method.
Time constants for velocity and position convergence can be separately adjusted with tunable parameters.
The method is more computationally more expensive because it requires modification of the output buffer history but is acceptable because updating the buffer only requires 6 FLOP per buffer index for a total of 30*6 = 180 FLOP per output update. EKF2 processing utilistion on a Pixracer board has going up from 15.1 to 15.7 % as a result of this change. These changes may not be necessary if the issues associated with the snapdragon board and range finder sensor data quality can be resolved.
The method was not applied to the attitude states because the quaternion operations required at each buffer index would have been computationally prohibitive.

The change was tested on a problem replay log with inconsistent range finder and IMU data: 

**Vertical Tracking Before Change**
posd before
veld before

**Vertical Tracking After Change**
posd after
veld after

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -229,6 +229,10 @@ struct parameters {
 	Vector3f rng_pos_body;	// xyz position of range sensor in body frame (m)
 	Vector3f flow_pos_body;	// xyz position of range sensor focal point in body frame (m)
 
+	// output complementary filter tuning
+	float vel_Tau;	// velocity state correction time constant (1/sec)
+	float pos_Tau;	// postion state correction time constant (1/sec)
+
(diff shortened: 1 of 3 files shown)


priseborough:
Output observer tracking results from flight test:

vn tracking
ve tracking
vd tracking
pn tracking
pe tracking
pd tracking

LorenzMeier:
Nice!

PR #131, opened 2016-05-15
Fixed compiler issues.

waltjohnson:
We addressed a few items that prevent the ECL from compiling with our compiler.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "matrix"]
-	path = matrix
-	url = git@github.com:inertialsense/Matrix.git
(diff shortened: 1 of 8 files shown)


LorenzMeier:
You removed a few initialisations without accounting for them properly somewhere else. This PR cannot be merged in its current state. All initialisers that got removed need to be called somewhere else. If they already were and I missed that please let me know.

waltjohnson:
Lorenz.  I double checked the initialization that removed and they are accounted for elsewhere.  So there shouldn't be a problem.  The five that I removed from estimator_interface.h (lines 259-263) where already in estimator_interface.cpp (lines 55-58 and 72).  The two from ekf.h (lines 136-137) were addressed by adding to ekf.cpp (lines 60-61).  I don't see any other initialization removed, so I think the pull request is kosher.   Let me know if I'm missing something.

LorenzMeier:
Is there a reason you closed it now?

Issue #128, opened 2016-05-13
Initialization in SITL is slow


julianoes:
The initialization in SITL is painfully slow and consumes a lot of time for developers.

The two delays to wait for are:
- ekf2 initialized
- home position initialized

Presumably both are due to the rather slow startup of ekf2.

It would be nicer to have this faster, or at least use a hack to speed it up for SITL.

priseborough:
https://github.com/PX4/ecl/pull/134

priseborough:
@julianoes If this PR is able to resolve your issues (in combination with parameter changes) then I will pull in the PR and close this issue.

priseborough:
Fixed in master

PR #134, opened 2016-05-17
EKF: Improved tilt alignment control

priseborough:
Improves control over the initial alignment process. Used in conjunction with https://github.com/PX4/Firmware/pull/4554.

Addresses #128 if combined with following tuning parameter changes to take advantage of 'ideal' SITL environment:

EKF2_GBIAS_INIT = 0.01
EKF2_ANGERR_INIT = 0.01

The quaternion covariances are converted to rotation vector variances so that the tilt uncertainty can be isolated from the yaw uncertainty when checking filter convergence.
The initial IMU bias uncertainty can be tuned to allow faster alignment for boards with accurate sensors (eg SITL)
The initial angular alignment uncertainty can be tuned to allow faster alignment for applications where the vehicle is not subject to accelerations on startup.
The initial quaternion state co-variances are set from the tilt uncertainty using the correct transformation from rotation vector to quaternions.

Has been bench tested with a range of  initial orientations.


priseborough:
Alignment from +90 pitch and +90 yaw is taking longer than it should - investigating now.

priseborough:
Alignment failure at 90deg pitch fixed by: https://github.com/PX4/ecl/pull/134/commits/e4b2e9c93dfad03aec14e974991a309d732ba1ee

PR #136, opened 2016-05-17
added flag for airspeed fusion control

CarlOlsson:


CarlOlsson:
@tumbili 
I added two fields to the control bitmask. What do you think?

EKF/common.h
```diff
@@ -414,6 +414,8 @@ union filter_control_status_u {
 		uint16_t gps_hgt     : 1; // 11 - true when range finder height is being fused as a primary height reference
 		uint16_t ev_pos      : 1; // 12 - true when local position data from external vision is being fused
 		uint16_t ev_yaw      : 1; // 13 - true when yaw data from external vision measurements is being fused
+		uint16_t fuse_airspeed : 1; // 14 - true when we want to fuse airspeed sensor data
+		uint16_t fuse_beta : 1; // 15 - true when we want to fuse syntetic sideslip measurements
```
priseborough: 'synthetic'

EKF/estimator_interface.h
```diff
@@ -155,6 +155,12 @@ class EstimatorInterface
 	// set vehicle landed status data
 	void set_in_air_status(bool in_air) {_control_status.flags.in_air = in_air;}
 
+	// set flag if airspeed data should be fused
+ 	void set_fuse_airspeed_flag(bool fuse_airspeed) {_control_status.flags.fuse_airspeed = fuse_airspeed;}
+
+ 	// set flag if syntetic sideslip measurement should be fused
```
priseborough: 'synthetic'

priseborough:
Looks good - moving the decision to use airspeed outside the EKF makes sense given the range of platform types and limitations of different airspeed measurement methods.

CarlOlsson:
Thanks for reviewing, no will see if I can do a pr for the fw now

tumbili:
@CarlOlsson Let me know if you have the PR for the firmware. We then only have to add flags for the replay indicating when airspeed / beta are fused.

priseborough:
@CarlOlsson I've rebased here: https://github.com/priseborough/ecl/tree/add_control_flag_2

It needs the following updates in PX4/Firmware due to the increased size of the status message:

https://github.com/priseborough/Firmware/tree/add_control_flag_2

priseborough:
@CarlOlsson did you want to continue with this PR?

PR #138, opened 2016-05-18
EV testing

pickledgator:
added ev_pos flag check to valid_lpos check.
need to disable fake gps when ev_pos is valid, not just optical flow.


LorenzMeier:
@priseborough Can we merge this?

PR #144, opened 2016-05-19
Addressed C99 compiler issues of initializing variables in header and…

waltjohnson:
unused included statements.  Also 	isfinite() is not part of the math class, rather the math.h header file.

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -57,6 +57,9 @@
 #endif
 
 
+const float Ekf::_k_earth_rate = 0.000072921f;
+const float Ekf::_gravity_mss = 9.80665f;
+
 Ekf::Ekf():
 	_filter_initialised(false),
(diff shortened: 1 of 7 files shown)


waltjohnson:
Lorenz.  The variable initialization I removed from the estimator_interface.h (lines 263-267) already exist in estimator_interface.cpp (lines 55-58 and 71).

LorenzMeier:
Thanks! Could we generalise the isfinite handling into a header, using this logic?

https://github.com/PX4/ecl/blob/e3b9800cac459e7c279395c18191a872f465f848/EKF/ekf.cpp#L45-L52

waltjohnson:
Yes, we can generalize isfinite()...

LorenzMeier:
@priseborough This looks good to me, its just changing the API from C++11 (which is kinda new) to C99.

PR #146, opened 2016-05-22
updated logic of calculation of body field variance

CarlOlsson:
Fixed typo in the check when updating the body field variance


LorenzMeier:
Nice!

PR #148, opened 2016-05-22
EKF2 provide alternative output velocity prediction

priseborough:
This PR introduces an optional method of predicting the velocity forward from the fusion to the output time horizon. It is activated by setting EKF2_TAU_VEL to a negative value. This causes the state history for the output predictor states to be corrected using the position difference instead of the velocity difference. The end result is a velocity that more closely matches the first derivative of the position. This may help in situations where control loops rely on kinematic consistency between position and velocity.

Tested with https://github.com/PX4/Firmware/tree/pr-ekf2OutputPredictor and EKF2_TAU_VEL = -1

Flight test log here: http://logs.uaventure.com/view/C5gVya4qXoKuHXPXEY5AxJ

Plots of the vertical position and velocity for the EKF and predictor:

vertical comparison

vertical comparison zoomed 1

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -161,7 +161,7 @@ public:
 	}
 
 	// return data at the specified index
-	data_type get_from_index(unsigned index)
+	inline data_type get_from_index(unsigned index)
 	{
 		if (index >= _size) {
 			index = _size-1;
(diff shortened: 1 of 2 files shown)


tumbili:
@paul The  code  looks  ok  to  me. I  think  you  can  bring  it  in, I'll try  to  give  it  a test soon.

LorenzMeier:
@julianoes Could you please flight-test this and drive the Snapdragon flight testing with it to completion?

julianoes:
@LorenzMeier So I should make test flights as follows?

- EKF2_TAU_VEL = 0
- EKF2_TAU_VEL = -1

LorenzMeier:
Question for Paul, but basically yes.

PR #150, opened 2016-05-24
EKF: Publish full reset and innovation consistency check status data

priseborough:
Position and velocity state reset deltas and times are captured and are published. This data allows control loops to compensate for resets via set-point changes if required.
Innovation consistency check failures are captured and published as an integer bit-mask. This data in combination with the innovations and innovation variances will assist with monitoring sensor health.

Indoor flight test with GPS using a pixracer board
http://logs.uaventure.com/view/XWF2ovMGg4sr2mKexoGd9c

### DIFF
--- a/EKF/airspeed_fusion.cpp
+++ b/EKF/airspeed_fusion.cpp
@@ -138,12 +138,10 @@ void Ekf::fuseAirspeed()
 
 		// If the innovation consistency check fails then don't fuse the sample and indicate bad airspeed health
 		if (_tas_test_ratio > 1.0f) {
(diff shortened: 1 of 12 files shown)


priseborough:
@tumbili are you happy with the way the reset and innovation gate rejection data is being presented at the public interface?

priseborough:
Changed to event counters after discussion with @tumbili

priseborough:
@tumbili we need the reset and innovation test fail status for integration with another project. Are you happy for this to go in, or is there something else you would like adjusted?

Issue #151, opened 2016-05-25
Specifying desired height sensor


pickledgator:
Per [control.cpp#L452-L470](https://github.com/PX4/ecl/blob/e2d9e19a5dde27e8db0d8f838ddd614bfd4b0f1b/EKF/control.cpp#L452-L470), what is the motivation for adding the _control_status.flags.XX as logical ors inside the if/else statements for baro/gps? Are we trying to limit the height sensor fusion to one sensor only or fuse all that are available? 

If the prior, seems to make sense to add add a param check for VDIST_SENSOR_EV (which exists in the enum already) to the first if statement and remove the _control_status.flags checks for baro/gps. If the later is the case, what purpose does the _params.vdist_sensor_type serve?

priseborough:
The original intention was to provide a structure to allow fusion of multiple height sources, but the height offsets were problematic unless additional states were added

Issue #141, opened 2016-05-19
Oscillations and settling time of EKF using EV


pickledgator:
Ran some more motion tests today using the EV fusion updates that were pulled into master a few days ago (and the new alignment PR from yesterday -- alignment seems much faster). I'm passing in vision updates at 100Hz and they are pretty clean. There's quite of a bit of overshoot and settling time in the ekf (as compared to the vision measurements) that I'm struggling to tune out. Eg., as I'm walking with the drone, when I physically stop motion, the vision measurements flatline, but the ekf oscilates twice or so and eventually settles. I've played around with the noise parameters for both EV and accel/accel bias, with no clear trends to reduce the overshoot. Not sure what other nobs I can turn. Could use some advice. Log file with replay enabled here (not sure why the LPOS data isn't show up on logmuncher?): http://logs.uaventure.com/view/SWMESe69SkMENaTSPwHwcZ

Thanks!

priseborough:
I can't get you log to work on the replay. Will investigate further.

priseborough:
It's because https://github.com/PX4/Firmware/pull/4382 was not finished. If you want help with the EV feature, then you must get that PR cleaned up, tested and pulled into master. @LorenzMeier for info.

pickledgator:
Thanks, @priseborough. Will get an updated log out when #4382 is resolved.

priseborough:
@pickledgator Can you please get a replay log using the following branch and send it too me as @devbharat is currently unable to generate logs. This will help me get #4382 resolved.

https://github.com/priseborough/Firmware/tree/px4_push_ev_Ekf2_replay

TXS

pickledgator:
@priseborough I pulled in your branch compiled and flash, but apparently my logs are still corrupt. They open in FlightPlot, but are missing tons of fields like, LPOS but have fields like attitude. When the px4log is loaded, it says there are errors but I have no idea how to troubleshoot this. My process was to pull down PX4/master, then merge in your branch and add the few lines of modifications that I need to get my vision data piped in. All data shows up fine live in QGC, but somehow isn't being logged correctly. Any help is appreciated so that I can get you some data.

devbharat:
Are you talking about the replay log ? I does not log LPOS etc, only replay stuff (test RPL1.t) and attitude. So that is OK.

priseborough:
I think I know what the problem could be - the changes to the field descriptor for the RPL5 message did not make it into the PR. I will fix that now.

[...]

PR #153, opened 2016-05-27
EKF: external vision bug fixes

priseborough:
Addresses: #141 and #151. Note turning accel bias learning off via the EKF2_AID_MASK parameter may still be beneficial depending on tuning for external vision operation with accurate sensing.

Required for https://github.com/PX4/Firmware/pull/4647

Replayed log fusing external vision data here (experimental data courtesy of @pickledgator from #141)

http://logs.uaventure.com/view/eZryf2znyV9stqokLeWkwj

Replayed log fusing external vision and range finder data here (experimental data courtesy of @pickledgator from #141)

http://logs.uaventure.com/view/aGBegC39R5Rdx33GazzuvF

EKF plots from second log

posd innov
posne innov
yaw innov
posned
vert tracking

log from normal quad flight with GPS here:

http://logs.uaventure.com/view/a5JUpD5uXC9thJV9kD6KRK

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -231,9 +231,7 @@ struct parameters {
 	float rng_gnd_clearance;	// minimum valid value for range when on ground (m)
 
 	// vision position fusion
-	float ev_noise;		// observation noise for vision sensor estimates (m)
 	float ev_innov_gate;		// vision estimator fusion innovation consistency gate size (STD)
-	float ev_gnd_clearance;	// minimum valid value for vision based height estimate when on ground (m)
 
 	// optical flow fusion
 	float flow_noise;		// observation noise for optical flow LOS rate measurements (rad/sec)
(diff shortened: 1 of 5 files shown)


priseborough:
@tumbili can you please have a look at this with me when you start work - I would like to do a walk through wiht you on the height reset changes and this needs to go in to close out the EV issues. TXS.

pickledgator:
@priseborough we should consider a fix for the buffer size probably as well [estimator_interface.h#L227](https://github.com/PX4/ecl/blob/e41524ac9d7cc6cc11ffc0f02561cec248985431/EKF/estimator_interface.h#L227). Perhaps, support buffer lengths for measurements up to 100Hz since that is common for range/ev?

priseborough:
The input rate into the buffer now automatically limits.
On May 27, 2016 14:50, pickledgator <notifications@github.com> wrote:@priseborough we should consider a fix for the buffer size probably as well estimator_interface.h#L227. Perhaps, support buffer lengths for measurements up to 100Hz since that is common for range/ev?

PR #176, opened 2016-07-12
Tiny PR: added getter function for wind states

yyassa:
Small interface to wind states, so we can call them from ekf2_main in a clean way.
(we use this for publishing estimated airspeed to controllers)


julianoes:
Applied via rebase, thanks.

PR #178, opened 2016-07-18
[WIP] Synthetic sideslip support ecl

CarlOlsson:
@priseborough 
This one goes together with https://github.com/PX4/Firmware/pull/5093


LorenzMeier:
You will want to rebase on master.

priseborough:
Do we have flight test results for this?

CarlOlsson:
This have been flight tested today.
With the current implementation synthetic sideslip measurements are fused 5 seconds after the airspeed sensor have initially been fused. If the airspeed sensor is found faulty the wind states are continued to be updated by the beta fusion. Below is a plot of the beta innovations and variance.

The airspeed sensor fusion was turned of in mid flight as seen in the plot below when the airspeed innovations stopped updating

LorenzMeier:
@CarlOlsson @priseborough Where is this standing?

priseborough:
It's on the todo list, but as it wasn't  blocking anything it kept getting pushed back by higher priority tasks. I'm happy with the fusion processing, but the logic of when to activate it needs to be reviewed. 

Based on experience with APM that uses an equivalent method, this fusion method is of primary value when flying fixed wing planes without airspeed sensing or when GPS is lost. For normal flight with an airspeed sensor, it does not make much difference, although if used in conjunction with airspeed sensing it will speed up convergence of the initial wind estimates in the situation where the plane does a sustained climb-out from launch without turning.

@CarlOlsson  and @tumbili  Did we get the replay with normal airspeed fusion working? If not I think that should be done before we re-start work on this PR, as it will be hard to check logic and tuning changes without replay.

tumbili:
@priseborough @CarlOlsson I couldn't find anything suspicious in the replay . At one point in the log you see massive excursions in z acceleration and also the airspeed various relatively fast between 35 and 5 m/s.
Eventually, fusing airspeed leads to a position reset which is event looking at the position innovations.
carl_fw_flight

CarlOlsson:
@tumbili hm that is interesting. I will do more flights with the airframe, fusing airspeed with standard logging and then also replaying to compare. I always thought that since the filter did not reset without airspeed fusion there was a problem with the fusion itself not that it helped to actually reset when it was a good thing to do.

priseborough:
I will start looking at this PR in conjunction with the replay. @CarlOlsson can you provide a link to a replay log wth airspeed data that you would like me to use to do the testing? Thanks.

priseborough:
I have made some changes and tested using replay. The updated PR is here: https://github.com/PX4/ecl/pull/207

PR #180, opened 2016-08-05
fixed wrong condition on which gps height would be fused

tumbili:
- this solves the issue where height innovations would jump
between two regimes


LorenzMeier:
Can you send an update PR to firmware as well? Thanks!

PR #182, opened 2016-08-11
Fixed wind estimation

vvranjek:
Wind estimation wasn't working for me so I fixed it.  I am not completely sure if I've done the whole thing correctly, since I am new to this code and only spent a day on it, but I am sure that the math/physics behind wind estimation is now correct. For example, I am not sure what is  `_control_status_prev.flags.wind` and what this does:
```
// initialise diagonal of covariance matrix for the wind velocity states
		for (uint8_t index = 22; index <= 23; index++) {
			P[index][index] = sq(5.0f);
		}
```
But wind estimation works for me on the pixhawk.

Cheers!

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -564,7 +564,7 @@ void Ekf::predictCovariance()
 	// Don't do covariance prediction on wind states unless we are using them
 	if (_control_status.flags.wind) {
 		// Check if we have just transitioned to using wind states and set the variances accordingly
-		if (!_control_status_prev.flags.wind) {
+		if (_control_status_prev.flags.wind) {
 			// simple initialisation of wind states: calculate wind component along the forward axis
 			// of the plane.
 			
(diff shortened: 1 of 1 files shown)


tumbili:
@vvranjek The code you have modified is just taking care of the initialization of the wind states.
The actual estimation of wind speed is happening when airspeed measurements are fused in.

vvranjek:
What does ' initialization of the wind states' mean? Where is the continuous wind calculation done then?

I dont't understand, was wind estimation working for you before? It didn't for me but now it does.

tumbili:
@vvranjek Everything that you need is implemented in #183 
Once wind estimation starts the wind states are initialized with the same formula you have  been using (zero side slip assumption). After that wind states are updated via fusion of airspeed measurements.
You can find the relevant code here https://github.com/PX4/ecl/blob/master/EKF/airspeed_fusion.cpp#L71

tumbili:
@vvranjek It would be cool if you could test fly #183 and give feedback on that (and share the logs).
@CarlOlsson Has already been flying it but his flight was very short.

PR #183, opened 2016-08-11
Fix wind state  and wind variance initialization

CarlOlsson:
@tumbili @priseborough 
This was tested successfully today. Very short flight and some pitot offset but the data looks good to me.

log is [here](http://logs.uaventure.com/view/rYeLRcDS4X9FUTJvDACTjH)


LorenzMeier:
@CarlOlsson Have you seen this PR?

https://github.com/PX4/ecl/pull/182

Would be good to make sure everything done there is also implemented here.

CarlOlsson:
@LorenzMeier @priseborough 
We have flown this around 20 times now and it looks good. Below are some sample plots. Please merge or update with what is left to do.

PR #184, opened 2016-08-12
fixed computation of range measurement timestamp

tumbili:

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -241,9 +241,7 @@ void EstimatorInterface::setRangeData(uint64_t time_usec, float *data)
 		rangeSample range_sample_new = {};
 		range_sample_new.rng = *data;
 		rng = *data;
-		range_sample_new.time_us -= _params.range_delay_ms * 1000;
-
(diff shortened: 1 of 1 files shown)


priseborough:
The additional delay (1/2 sampling interval) added assumes that range data is down-sampled using an a averaging technique. Is this confirmed as correct?

tumbili:
@priseborough Sorry for the delay. The assumption of an averaging technique does not apply to the range finder. Additionally, I don't think it applies for the mag and baro data.

priseborough:
So as far as you know we are using the most recent sensor sample with no downsampling technique that would introduce a delay?

tumbili:
@priseborough Yes, as far as I know. We need to see if this also applies to the baro and the mag samples because we have the delay in there now. @julianoes Is there any down-sampling happening for baro / mag sensor data?

julianoes:
@tumbili I don't think so but I'd have to check in source.

tumbili:
@priseborough Updated this according to the discussion. I've also opened another issue for the other two sensor #192

PR #190, opened 2016-08-18
Update matrix.

jgoppert:


jgoppert:
The last time matrix was updated was February, so this brings axis angle representation and another of other improvements. Also, it is way behind the matrix version on PX4 and this brings them in sync.

jgoppert:
@LorenzMeier @priseborough @tumbili @bkueng 

This also brings up the issue that we are using multiple versions of the same library onboard, which has to be a waste of space, so it would be nice if we could remove this submodule here.

bkueng:
I completely agree, there should only be one matrix library. @jgoppert can you remove one?

priseborough:
The original intention was that ECL should be useable as a stand-alone library for non-PX4 projects (ArduPilot, Paparazzi, etc), so we avoided calls to functions inside other PX4 libraries. Provided we maintain the ability for for it to be integrated into non-PX4 projects, then I don't have a preference.

PR #191, opened 2016-09-14
ecl attitude_fw fix excessive body fixed turn offset

### LINKED ISSUE #181: ECL_PitchController body_fixed_turn_offset issue
The ecl pitch controller body fixed turn offset is an additional term added to the rate setpoint. 

https://github.com/PX4/ecl/blob/master/attitude_fw/ecl_pitch_controller.cpp#L153

    float body_fixed_turn_offset = (fabsf((CONSTANTS_ONE_G / airspeed) *
				  		tanf(constrained_roll) * sinf(constrained_roll)));

With large roll angles this term becomes enormous and causes problems like in this recent flight. Note the small pitch setpoint, but full elevator deflection.

ecl

I've scaled the roll, pitch, and pitch setpoints to degrees.

Clearly this should be limited, but to what? Controller roll limit?

dagar:
-limit body fixed turn offset contrained roll to possible
roll setpoint range
-fixes #181

### DIFF
--- a/attitude_fw/ecl_pitch_controller.cpp
+++ b/attitude_fw/ecl_pitch_controller.cpp
@@ -127,7 +127,7 @@ float ECL_PitchController::control_bodyrate(const struct ECL_ControlData &ctl_da
 	/* roll is used as feedforward term and inverted flight needs to be considered */
 	if (fabsf(ctl_data.roll) < math::radians(90.0f)) {
 		/* not inverted, but numerically still potentially close to infinity */
-		constrained_roll = math::constrain(ctl_data.roll, math::radians(-80.0f), math::radians(80.0f));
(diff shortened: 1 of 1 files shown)


dagar:
This fixes the current usage in PX4, where inverted flight isn't yet possible and the max roll is going to be set to something reasonable.

LorenzMeier:
Good to merge / tested?

tumbili:
@dagar While you are at it, could you please check if this constrain is correct?
https://github.com/PX4/ecl/blob/81d891a58c7e38296f0b22e7e2d29fd29cbab3c8/attitude_fw/ecl_pitch_controller.cpp#L142
Shouldn't the arguments be swapped?

dagar:
@tumbili Yes, but now I'm thinking I should probably fix the same numerical problem if I'm going to touch the inverted section at all.

PR #193, opened 2016-09-20
Pr hagl init

tumbili:
- assign the flag storing the initialization state of the terrain estimator to the output of the init function in the ekf main init function. The estimator was initializing twice before.

- when initializing the terrain estimator use the projection of the range measurement rather than the raw value. Also do not allow initialization if the vehicle is tilted too much.


priseborough:
Looks OK to me.

PR #195, opened 2016-09-27
Remove optimization flags

julianoes:
This removes the optimization flag (optimize for size) out of the module because it should be taken care of on a platform level rather than a module level.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,15 +30,12 @@
 # POSSIBILITY OF SUCH DAMAGE.
 #
 ############################################################################
-set(MODULE_CFLAGS )
-if(${OS} STREQUAL "nuttx")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ endif()
 px4_add_module(
 	MODULE lib__ecl
 	STACK_MAIN 6000
-	COMPILE_FLAGS ${MODULE_CFLAGS} -Os
+	COMPILE_FLAGS ${MODULE_CFLAGS}
```
bkueng: Agreed. And while you're at it, line 33-36 can be removed and `STACK_MAX 6000` can be added instead

julianoes: Done in 4fefe5c6c66844838f3d0923c3403d5ddabeb4fe

PR #203, opened 2016-10-18
EKF: Add IMU vibration calculation and reporting

priseborough:
Add calculation and reporting of IMU delta angle and velocity coning and high frequency vibration.

These will need to be published on the estimator_status uORB message which is currently un-used.

This has been flight tested and the messages checked using this branch: https://github.com/PX4/Firmware/pull/5682

screen shot 2016-10-18 at 5 31 10 pm

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -121,6 +121,14 @@ public:
 	// error magnitudes (rad), (m/s), (m)
 	void get_output_tracking_error(float error[3]);
 
+	/*
+	Returns  following IMU vibration metrics in the following array locations
+	0 : Gyro delta angle coning metric = filtered length of (delta_angle x prev_delta_angle)
+	1 : Gyro high frequency vibe = filtered length of (delta_angle - prev_delta_angle)
+	2 : Accel high frequency vibe = filtered length of (delta_velocity - prev_delta_velocity)
+	*/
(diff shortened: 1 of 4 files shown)


priseborough:
Roman, can you take a quick look at this? It's only for data logging, but I need this data to complete the estimator health assessment tool. Thanks.

tumbili:
@priseborough Sure!

tumbili:
@priseborough The implementation looks good! Do you have some literature about those metrics?

tumbili:
@priseborough Is the high frequency delta angle vibration measure supposed to capture aliasing effects?
In the upper plot you have shown the value (middle plot) stays more or less constant which I guess is good.
I assume if you had aliasing you would see it growing, right?

priseborough:
It's very hard to detect aliasing, however large levels of higher frequency vibration that degrade accuracy of the numerical integration will show up. Normal The limits for these can be set for each platform which means that abnormal vibration levels can be caught during factory acceptance testing.

priseborough:
@tumbili and @bkueng Is there a wiki with information on interpreting logs that i could add explanations to for the various estimator status messages?

bkueng:
Not that I know of. You can add it to the msg topic, and from there we could generate a wiki page eventually.

PR #204, opened 2016-10-24
ekf reset: more granular reset / timeout strategy for gps fusion

tumbili:
- if both gps position and velocity measurements are rejected for 7 seconds
do a reset
- if only gps position measurements are rejected then wait for 14 seconds
as we still have velocity measurements to constrain the drift in position
- introduced ecl internal parameter for the timeout

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -261,6 +261,9 @@ struct parameters {
 	float vel_Tau;	// velocity state correction time constant (1/sec)
 	float pos_Tau;	// postion state correction time constant (1/sec)
 
+	unsigned no_gps_timeout_max;	// maximum time we allow dead reckoning while both gps position and velocity measurements are being
+									// rejected
+
 	// Initialize parameter values.  Initialization must be accomplished in the constructor to allow C99 compiler compatibility.
 	parameters()
 	{
@@ -350,6 +353,8 @@ struct parameters {
 		// output complementary filter tuning time constants
 		vel_Tau = 0.25f;
 		pos_Tau = 0.25f;
+
+		no_gps_timeout_max = 7e6;	// maximum seven seconds of dead reckoning time for gps
 
 	}
 };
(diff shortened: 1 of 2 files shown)


tumbili:
@priseborough As partly discussed this PR does the following:

1) If we get no gps data for longer than 5 seconds stop gps aiding.

2) If  gps position measurements are rejected but the velocities are not we wait for **2*gps_fusion_timeout** seconds before we do a reset.

3) If both gps position and velocity measurements are rejected we wait for **gps_fusion_timeout** seconds before we do a reset.

LorenzMeier:
What do we need in terms of testing to merge this?

tumbili:
@priseborough Changes look good to me.

priseborough:
I can bench test the loss and regaining of GPS. @tumbili you mentioned you had already tested the effect of GPS position jumps in SITL.

priseborough:
One behaviour that I did notice during testing was that when re-entering GPs aiding after having been reverted due to prolonged loss of GPS, the EKF will start using the GPS again as soon as it gets a 3D fix which makes the solution move around a by several meters as the receiver  improves its initial lock.    

This could be prevented by making the GPS pass all of the checks it is required to pass before entering GPs aiding for the first time. The upside is a better solution when the EKF re-enters GPs aiding, the downside is that it will take longer to start providing a nav solution after GPs is re-gained.

At the moment I will leave it as is and see how the loss/regaining of GPS works in conjunction with optical flow.

priseborough:
@tumbili I'm getting some unexpected behaviour during SITL testing of GPS position jumps. It's getting late here so I will investigate tomorrow.

tumbili:
@priseborough I gave this a test together with the reset branch. It seemed to work fine, what was the issue you were encountering?

priseborough:
The anomaly has been resolved and was due to a scaling error in the timescale when using MAVExplorer. That means testing is complete, so I will merge it now.

PR #207, opened 2016-10-27
Add synthetic sideslip support to EKF2

priseborough:
Enables EKF2 to do wind speed estimation without airspeed sensing on planes.

Modifies and replaces https://github.com/PX4/ecl/pull/178. 

Requires https://github.com/PX4/Firmware/pull/5763 to enable activation of the sideslip function.

**Replay Testing**

Has been tested in replay on the following problem log (WingWing airframe with bad airspeed calibration and high g manoeuvres) http://logs.uaventure.com/view/K3fbyTrttBLojq3KjQ2Pk7

Testing showed that wind speed estimation using only sideslip fusion outperformed wind speed estimation using only airspeed fusion when the the plane was regularly changing course and the airspeed sensor calibration was poor .

Replay using the following parameter changes to disable airspeed fusion and enable sideslip fusion:
EKF2_ARSP_THR 0
EKF2_FUSE_BETA 1
EKF2_BETA_NOISE 0.3

Wind Velocity Estimates are stable and converge quickly
screen shot 2016-10-27 at 4 16 54 pm

Position innovations are unaffected:
screen shot 2016-10-27 at 3 53 33 pm

Replay using the following parameter changes to enable airspeed fusion and disable sideslip fusion:
EKF2_ARSP_THR 10.0
EKF2_FUSE_BETA 0
EKF2_BETA_NOISE 0.3

Wind Velocity Estimates converge quickly but are less stable due to the airspeed error
airspeed and sideslip

Position innovations are unaffected:
screen shot 2016-10-27 at 4 06 28 pm

Airspeed Innovations are large indicating large airspeed errors:
screen shot 2016-10-27 at 4 03 05 pm

Plotting GPS speed and true airspeed supports the hypothesis that poor airspeed calibration is to blame:
screen shot 2016-10-27 at 10 39 59 am

Replay using the following parameter changes to enable both airspeed fusion and sideslip fusion:
EKF2_ARSP_THR 10.0
EKF2_FUSE_BETA 1
EKF2_BETA_NOISE 0.3

Wind Velocity Estimates are not much different to  using airspeed alone:
screen shot 2016-10-27 at 4 11 19 pm

**SITL Testing**

TODO - test response to GPS loss during a fixed wing SITL flight


priseborough:
I've rebased on latest master and updated the fusion using the same pattern introduced by https://github.com/PX4/ecl/pull/210 to save RAM.

During replay testing with a bug in the fusion indexing which caused poor conditioning of the innovation variance calculation, I found that if the sideslip fusion was badly conditioned, the entire covariance matrix was being reset, rather than just the rows and columns for the states being modified, eg wind. This caused the reset to affect states other than wind. A patch to fix this has been introduced.

The default tuning parameter for sideslip 'noise' has been updated to match the value used during replay testing that gave reasonably fast wind and smooth wind speed convergence.

Here's the wind speed convergence using sideslip only for the latest rebase and updates at 1c3fd0f:

screen shot 2016-11-05 at 12 40 19 pm

LorenzMeier:
Nice! Would you say this is ready to go?

priseborough:
Yes - ready to go in. It's off by default, but can be enabled by setting the  EKF2_FUSE_BETA parameter to 1. This also reminds me that I should put together a wiki explaining the various ecl ekf features and their corresponding parameters.

PR #208, opened 2016-11-01
Save memory

bkueng:
Save memory by moving the KHP matrix to the stack and using a smaller timeout data type in DataValidator.

@priseborough @tumbili can you please review?

### DIFF
--- a/validation/data_validator_group.cpp
+++ b/validation/data_validator_group.cpp
@@ -60,6 +60,12 @@ DataValidatorGroup::DataValidatorGroup(unsigned siblings) :
 
 DataValidatorGroup::~DataValidatorGroup()
 {
+	while (_first) {
+		DataValidator* next = _first->sibling();
+		delete (_first);
+		_first = next;
+	}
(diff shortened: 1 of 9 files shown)


validation/data_validator_group.cpp
```diff
@@ -49,29 +50,47 @@ DataValidatorGroup::DataValidatorGroup(unsigned siblings) :
 	_first_failover_time(0),
 	_toggle_count(0)
 {
+	assert(siblings > 0);
```
LorenzMeier: This will crash the drone. It should rather fail gracefully.

bkueng: No, you're right it's at runtime (but only debug mode). I removed it now.

priseborough:
If you want to save some more, then the same thing can be done for KH

bkueng:
You're right, thanks! At first it looked like all elements of KH are read, but only a few columns were set before. Now I see only the columns are used which were set right before.

PR #209, opened 2016-11-02
EKF: save some stack space

priseborough:
KH doesn't need to be a class variable so this will save 24x24x4 bytes


bkueng:
That does not work, as we now have KHP & KH on the same same stack, meaning this only moves memory, but does not reduce it. But I think we can merge the two loops, and reduce KH to a 1D array of only the column elements of the current row.

priseborough:
I think the ring buffers are low hanging fruit. They are longer than they need to be for most setups and we could avoid creating a buffer instance for sensors that are not fitted (eg airspeed and EV)

PR #212, opened 2016-11-04
EKF: Fix bug causing repeated resets if GPS sensor disconnected

priseborough:
This bug could cause eventual loss of attitude reference when operating without GPS connected.


priseborough:
This is a critical bug-fix, so I am merging it now.

PR #213, opened 2016-11-06
EKF:  Reduce RAM usage and output predictor errors

priseborough:
Saves RAM and reduces output predictor errors by using the smallest data buffer length that meets time delay and update rate requirements. Buffer sizes used are sent to ground via ECL info message to enable verification and testing with default parameters has verified the correct IMU buffer size of 21 and observation buffer size of 11.

screen shot 2016-11-07 at 10 18 00 am

Replay flight log here (used https://github.com/PX4/Firmware/pull/5814): http://logs.uaventure.com/view/ULJNmyuTUe9SjAfkh6e3jG

Replay log result  here (used https://github.com/PX4/Firmware/pull/5814):  http://logs.uaventure.com/view/yN5xVF9K6CoNhFDPefK2zk

### DIFF
--- a/EKF/estimator_interface.h
+++ b/EKF/estimator_interface.h
@@ -280,14 +280,14 @@ protected:
 	 max freq (Hz) = (OBS_BUFFER_LENGTH - 1) / (IMU_BUFFER_LENGTH * FILTER_UPDATE_PERIOD_MS * 0.001)
 	 This can be adjusted to match the max sensor data rate plus some margin for jitter.
 	*/
-	uint8_t OBS_BUFFER_LENGTH;
+	uint8_t _obs_buffer_length;
 	/*
 	IMU_BUFFER_LENGTH defines how many IMU samples we buffer which sets the time delay from current time to the
 	EKF fusion time horizon and therefore the maximum sensor time offset relative to the IMU that we can compensate for.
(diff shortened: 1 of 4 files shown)


EKF/estimator_interface.h
```diff
@@ -273,8 +273,21 @@ class EstimatorInterface
...
+	 This can be adjusted to match the max sensor data rate plus some margin for jitter.
+	*/
+	uint8_t OBS_BUFFER_LENGTH;
+	/*
+	IMU_BUFFER_LENGTH defines how many IMU samples we buffer which sets the time delay from current time to the
+	EKF fusion time horizon and therefore the maximum sensor time offset relative to the IMU that we can compensate for.
+	max sensor time offet (msec) =  IMU_BUFFER_LENGTH * FILTER_UPDATE_PERIOD_MS
+	This can be adjusted to a value that is FILTER_UPDATE_PERIOD_MS longer than the maximum observation time delay.
+	*/
+	uint8_t IMU_BUFFER_LENGTH;
```
bkueng: Can you rename these to `_imu_buffer_length` & `_obs_buffer_length`, now that they're not static const anymore?

priseborough: Will fix

LorenzMeier:
@bkueng good to go?

bkueng:
yes, merged!

PR #214, opened 2016-11-14
EKF minor bug fixes

priseborough:
Fixes a bug that allows air data fusion to start when the filter is still assuming constant position.
Fixes a bug that prevents a position reset happening immediately if GPS is gained and passes checks when already using optical flow.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -761,8 +761,9 @@ void Ekf::controlAirDataFusion()
 
 	}
 
-	// Always try to fuse airspeed data if available and we are in flight
-	if (_tas_data_ready && _control_status.flags.in_air) {
+	// Always try to fuse airspeed data if available and we are in flight and the filter is operating in a normal aiding mode
+	bool is_aiding = _control_status.flags.gps || _control_status.flags.opt_flow || _control_status.flags.ev_pos;
+	if (_tas_data_ready && _control_status.flags.in_air && is_aiding) {
 		// If starting wind state estimation, reset the wind states and covariances before fusing any data
 		if (!_control_status.flags.wind) {
 			// activate the wind states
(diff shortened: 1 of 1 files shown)


EKF/control.cpp
```diff
@@ -761,10 +761,12 @@ void Ekf::controlAirDataFusion()
...
 	}
 
-	// Always try to fuse airspeed data if available and we are in flight
-	if (_tas_data_ready && _control_status.flags.in_air) {
+	// Always try to fuse airspeed data if available and we are in flight and the filter is operating in a normal aiding mode
+	bool is_aiding = _control_status.flags.gps || _control_status.flags.opt_flow || _control_status.flags.ev_pos;
+	if (_tas_data_ready && _control_status.flags.in_air && is_aiding) {
 		// If starting wind state estimation, reset the wind states and covariances before fusing any data
 		if (!_control_status.flags.wind) {
+			ECL_INFO("EKF commencing wind estimation using airspeed");
```
LorenzMeier: Its a bit verbose - is this necessary?

EKF/control.cpp
```diff
@@ -793,11 +795,14 @@ void Ekf::controlBetaFusion()
...
-	// Attempt synthetic sideslip fusion at the prescribed rate if we are in-flight and fusion has been enabled
+	// Attempt synthetic sideslip fusion at the prescribed rate if we are in-flight beta fusion has been enabled externally
+	// and the filter is operating in a normal aiding mode
 	bool beta_fusion_time_triggered = _time_last_imu - _time_last_beta_fuse > _params.beta_avg_ft_us;
-	if(beta_fusion_time_triggered && _control_status.flags.fuse_beta && _control_status.flags.in_air){
+	bool is_aiding = _control_status.flags.gps || _control_status.flags.opt_flow || _control_status.flags.ev_pos;
+	if(beta_fusion_time_triggered && _control_status.flags.fuse_beta && _control_status.flags.in_air && is_aiding) {
 		// If starting wind state estimation, reset the wind states and covariances before fusing any data
 		if (!_control_status.flags.wind) {
+			ECL_INFO("EKF commencing wind estimation using sideslip");
```
priseborough: Informative for developing, but not necessary.

LorenzMeier: I get that, but it will eventually make it into the log. Do you want / need that? Otherwise I would remove it.

priseborough: i will remove it

CarlOlsson:
I test flew this today, logfile is [here](http://review.px4.io/plot_app?log=3bdafe4f-0b42-489a-bea9-2726188872ea)

I am not sure why flight Review is telling me that LPE was used when EKF2 was used (if I am not mistaking, not so used to looking at data from the new logger yet). @bkueng

bkueng:
@CarlOlsson It looks at the `SYS_MC_EST_GROUP` parameter. But on FW this param is ignored and ekf2 is always used.
Did you see the `[sensors] Gyro #0 failover : Stale data!`? That's something that should be looked into.

priseborough:
I have checked the estimator data from this flight. The log shows that airspeed fusion did not start until the filter was operating in the normal aiding mode which is expected behaviour from patch 1af85d7.

Inflight initialisation of the velocity, position and wind states was performed successfully.

priseborough:
I noticed in the data that the synthetic sideslip fusion is currently only being performed at 1Hz due to a default parameter setting which is going to be too low to assist with dead-reckoning during GPS outage for fixed wing. I will adjust this to 7Hz and check it on the replay.

priseborough:
Running the replay using this log data http://logs.uaventure.com/view/K3fbyTrttBLojq3KjQ2Pk7 and comparing the use of sideslip at 1Hz vs 7Hz gives:

1Hz (exisiting rate):
1hz sideslip fusion

7Hz (proposed rate):
7hz sideslip fusion

Other than faster and noisier forming of initial estimates there is not much difference. If we fuse both sideslip and airspeed and compare the differences are not noticeable with airspeed fusion dominating the results:

airspeed fusion + 1Hz sideslip fusion:
1hz

airspeed fusion + 7Hz sideslip fusion:
7hz

PR #216, opened 2016-11-28
EKF: Improve use of range finder data

priseborough:
- Enables use of range finders that are misaligned in pitch relative to the Z body axis.
- Enables the range finder observation variance to be scaled with range to ground to allow for scale factor errors associated with some ranging techniques.
- Fixes a false triggering in the local position height above ground validity checks when operating close to the ground.
- Use norm() instead of length() to maintain consistency with rest of code.

### DIFF
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -99,7 +99,7 @@ void Ekf::fuseHagl()
 		_hagl_innov = pred_hagl - meas_hagl;
 
 		// calculate the observation variance adding the variance of the vehicles own height uncertainty and factoring in the effect of tilt on measurement error
-		float obs_variance = fmaxf(P[9][9], 0.0f) + sq(_params.range_noise / _R_rng_to_earth_2_2);
+		float obs_variance = fmaxf(P[9][9], 0.0f) + (sq(_params.range_noise) + sq(_params.range_noise_scaler * _range_sample_delayed.rng)) * _R_rng_to_earth_2_2;
 
 		// calculate the innovation variance - limiting it to prevent a badly conditioned fusion
 		_hagl_innov_var = fmaxf(_terrain_var + obs_variance, obs_variance);
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -70,20 +70,35 @@ bool Ekf::initHagl()
 	}
 }
 
-void Ekf::predictHagl()
+void Ekf::runTerrainEstimator()
 {
-	// predict the state variance growth
-	// the state is the vertical position of the terrain underneath the vehicle
+	// Perform a continuity check on range finder data
+	checkRangeDataContinuity();
 
-	// process noise due to errors in vehicle height estimate
-	_terrain_var += sq(_imu_sample_delayed.delta_vel_dt * _params.terrain_p_noise);
+	// Perform initialisation check
+	if (!_terrain_initialised) {
+		_terrain_initialised = initHagl();
 
-	// process noise due to terrain gradient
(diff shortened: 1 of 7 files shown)


EKF/terrain_estimator.cpp
```diff
@@ -99,7 +99,7 @@ void Ekf::fuseHagl()
 		_hagl_innov = pred_hagl - meas_hagl;
 
 		// calculate the observation variance adding the variance of the vehicles own height uncertainty and factoring in the effect of tilt on measurement error
-		float obs_variance = fmaxf(P[9][9], 0.0f) + sq(_params.range_noise / _R_rng_to_earth_2_2);
+		float obs_variance = fmaxf(P[9][9], 0.0f) + (sq(_params.range_noise) + sq(_params.range_noise_scaler * _range_sample_delayed.rng)) * _R_rng_to_earth_2_2;
```
priseborough: That's an error. thanks for spotting it.

priseborough: Actually, there is an error, but thats not it. If the vehicle is banked at 45 degrees, each metre of range error will only produce 0.71m of height error, so it should be *. However the _R_rng_to_earth_2_2_ should be squared, because it scales from range to height and we are calculating a variance that has units of m^2.

priseborough: The  height error when tilted will also be affected by deviation of the terrain away from the flat earth assumption and errors in the tilt angle estimate which will increase the error with bank angle. This means that it is not practical to scale height error with tilt. I will remove the R_rng_to_earth_2_2 scaler.

tumbili: @priseborough I actually thought that the reason for increasing the observation variance with tilt was to take into account errors in the flat earth model.

priseborough:
Closed while we restore some range finder update rate checks that were lost during the rebase.

EKF/terrain_estimator.cpp
```diff
@@ -69,37 +70,52 @@ bool Ekf::initHagl()
...
 		// predict the hagl from the vehicle position and terrain height
 		float pred_hagl = _terrain_vpos - _state.pos(2);
 
 		// calculate the innovation
 		_hagl_innov = pred_hagl - meas_hagl;
 
-		// calculate the observation variance adding the variance of the vehicles own height uncertainty and factoring in the effect of tilt on measurement error
-		float obs_variance = fmaxf(P[9][9], 0.0f) + sq(_params.range_noise / _R_to_earth(2, 2));
+		// calculate the observation variance adding the variance of the vehicles own height uncertainty
+		float obs_variance = fmaxf(P[9][9], 0.0f) + sq(_params.range_noise) + sq(_params.range_noise_scaler * _range_sample_delayed.rng);
```
tumbili: @priseborough Why aren't we computing the measurement noise in the vertical direction, e.g. multiplying range noise by sq(_R_rng_to_earth_2_2) ?

EKF/terrain_estimator.cpp
```diff
@@ -156,3 +169,27 @@ void Ekf::get_hagl_innov_var(float *hagl_innov_var)
...
+
+// check that the range finder data is continuous
+void Ekf::checkRangeDataContinuity()
+{
+	// update range data continuous flag (2Hz ie 500 ms)
+	/* Timing in micro seconds */
+	static hrt_abstime t = 0;
+	static hrt_abstime t_prev = 0;
+	static float dt = 0.0f;
+	t = hrt_absolute_time();
```
tumbili: @priseborough  I think it would be good if we can avoid using calls to the hrt timer in order to avoid dependencies.
I'll look how we can solve this.

tumbili: @devbharat Sorry, Bharat is actually the bad guy :-)

tumbili: This is addressed in #230

tumbili:
Replaced by #230

PR #229, opened 2017-01-04
EKF: fix errors found during Coverity scan

priseborough:
https://scan4.coverity.com/reports.htm#v26785/p10036/fileInstanceId=16639439&defectInstanceId=3367895&mergedDefectId=139419&fileStart=251&fileEnd=452

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -62,6 +62,7 @@ EstimatorInterface::EstimatorInterface():
 	_yaw_test_ratio(0.0f),
 	_tas_test_ratio(0.0f),
 	_terr_test_ratio(0.0f),
+	_beta_test_ratio(0.0f),
 	_time_last_imu(0),
 	_time_last_gps(0),
 	_time_last_mag(0),
@@ -69,6 +70,7 @@ EstimatorInterface::EstimatorInterface():
 	_time_last_range(0),
 	_time_last_airspeed(0),
 	_time_last_ext_vision(0),
+	_time_last_optflow(0),
 	_mag_declination_gps(0.0f),
 	_mag_declination_to_save_deg(0.0f)
 {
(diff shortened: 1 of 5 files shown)


EKF/estimator_interface.cpp
```diff
@@ -78,6 +80,13 @@ EstimatorInterface::EstimatorInterface():
 	_delta_ang_prev.setZero();
 	_delta_vel_prev.setZero();
 	memset(_vibe_metrics, 0, sizeof(_vibe_metrics));
+	memset(&_mag_sample_delayed, 0, sizeof(_mag_sample_delayed));
```
tumbili: @priseborough I'm never sure if this works because the class embodies vector class instances.

LorenzMeier:
Can you please remove all memset calls and instead use the .zero() (o equivalent) calls of the Matrix libary?

Also, when copying vectors, use a proper copy constructor instead. I don't think we should be seeing any memset or memcpy calls anywhere if this is concluded properly.

EKF/estimator_interface.cpp
```diff
@@ -106,8 +115,8 @@ void EstimatorInterface::setIMUData(uint64_t time_usec, uint64_t delta_ang_dt, u
 
 	// copy data
 	imuSample imu_sample_new = {};
-	memcpy(&imu_sample_new.delta_ang._data[0], delta_ang, sizeof(imu_sample_new.delta_ang._data));
-	memcpy(&imu_sample_new.delta_vel._data[0], delta_vel, sizeof(imu_sample_new.delta_vel._data));
+	memcpy(&imu_sample_new.delta_ang, delta_ang, sizeof(imu_sample_new.delta_ang));
```
tumbili: @priseborough Doesn't this give you the address of the delta_ang instance rather than the address of the float array?

EKF/estimator_interface.cpp
```diff
@@ -62,22 +62,31 @@ EstimatorInterface::EstimatorInterface():
...
 	_time_last_ext_vision(0),
+	_time_last_optflow(0),
 	_mag_declination_gps(0.0f),
-	_mag_declination_to_save_deg(0.0f)
+	_mag_declination_to_save_deg(0.0f),
+	_pos_ref{},
+	_mag_test_ratio{},
+	_vel_pos_test_ratio{},
+	_vibe_metrics{},
+	_mag_sample_delayed{},
```
tumbili: @bkueng Can you tell me if this kind of initialisation works for structs with non POD members?

tumbili:
@bkueng Could you do a review please?

tumbili:
@bkueng Let me fix CI first.

EKF/estimator_interface.cpp
bkueng: The address of a struct is guaranteed to be the same as its first member's. However if in future a new struct member is added in front, this will break. So the previous version would be the safer choice.

EKF/optflow_fusion.cpp
```diff
@@ -397,8 +397,6 @@ void Ekf::fuseOptFlow()
 			// run innovation consistency check
 			optflow_test_ratio[1] = sq(_flow_innov[1]) / (sq(math::max(_params.flow_innov_gate, 1.0f)) * _flow_innov_var[1]);
 
-		} else {
-			return;
```
bkueng: Looks like the loop has the following form:
```
for (uint8_t obs_index = 0; obs_index <= 1; obs_index++) {
  if (obs_index == 0) {
   X
  } else if (obs_index == 1) {
   Y
  }
}
```
w/o dependence on `obs_index` and thus can be simplified to:
```
{
  X
}
{
  Y
}
```

EKF/estimator_interface.cpp
bkueng: I'd say yes except for references.

LorenzMeier:
Rebased, patched and applied.

PR #230, opened 2017-01-05
Pr ekf rng fuse updates2

tumbili:
@devbharat @priseborough This is a replacement for #216 

@devbharat Check the last commit I did. I got rid of the hrt dependency and the static variables and fixed some other stuff.
I gave it a test in SITL and the result was as expected. Can you have a look?

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -83,8 +83,12 @@ void Ekf::controlFusionModes()
 	_gps_data_ready = _gps_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_gps_sample_delayed);
 	_mag_data_ready = _mag_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_mag_sample_delayed);
 	_baro_data_ready = _baro_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_baro_sample_delayed);
+
+	// calculate 2,2 element of rotation matrix from sensor frame to earth frame
+	_R_rng_to_earth_2_2 = _R_to_earth(2, 0) * _sin_tilt_rng + _R_to_earth(2, 2) * _cos_tilt_rng;
 	_range_data_ready = _range_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_range_sample_delayed)
-			&& (_R_to_earth(2, 2) > 0.7071f);
+			&& (_R_rng_to_earth_2_2 > 0.7071f);
+
 	_flow_data_ready = _flow_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_flow_sample_delayed)
 			&&  (_R_to_earth(2, 2) > 0.7071f);
 	_ev_data_ready = _ext_vision_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_ev_sample_delayed);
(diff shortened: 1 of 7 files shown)


tumbili:
@devbharat @priseborough I rebased this on master, can somebody please do a review so that we can bring it in?

EKF/terrain_estimator.cpp
```diff
@@ -69,37 +70,52 @@ bool Ekf::initHagl()
...
 		// predict the hagl from the vehicle position and terrain height
 		float pred_hagl = _terrain_vpos - _state.pos(2);
 
 		// calculate the innovation
 		_hagl_innov = pred_hagl - meas_hagl;
 
-		// calculate the observation variance adding the variance of the vehicles own height uncertainty and factoring in the effect of tilt on measurement error
-		float obs_variance = fmaxf(P[9][9], 0.0f) + sq(_params.range_noise / _R_to_earth(2, 2));
+		// calculate the observation variance adding the variance of the vehicles own height uncertainty
+		float obs_variance = fmaxf(P[9][9], 0.0f) + sq(_params.range_noise) + sq(_params.range_noise_scaler * _range_sample_delayed.rng);
```
tumbili: @priseborough Shouldn't we be multiplying the last term in the upper line with sq(_R_to_earth(2, 2))?

LorenzMeier:
@priseborough @ChristophTobler Can you please review?

EKF/terrain_estimator.cpp
priseborough: For a level flat earth without tilt estimation and range finder alignment errors  where the only error source is scale factor n the range finder the answer is yes. However terrain height variation underneath the vehicle and attitude errors do increase the measurement height uncertainty with range.  The derivative from vehicle tilt error and terrain tilt to height scales with range * sin(tilt angle). This tends to cancel the effect of range scale errors being 'compressed' with tilt.

tumbili:
@priseborough I pushed two more commits. I noticed that the sin and cos values of the range pitch angle were only calculated once on initialisation but not after. This means that parameter changes of the range sensor pitch angle were not having any effect.

EKF/control.cpp
```diff
@@ -84,6 +84,10 @@ void Ekf::controlFusionModes()
 	_mag_data_ready = _mag_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_mag_sample_delayed);
 	_baro_data_ready = _baro_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_baro_sample_delayed);
 
+	// update range sensor angle parameters in case they have changed
```
priseborough: I would move them into the setRangeData function inside the 

if (time_usec - _time_last_range > _min_obs_interval_us)

conditional so they are not unnecessarily calculated every EKF update.

tumbili:
@priseborough Paul, moving it in the function you proposed does not work since the variables are not members of the base class (EstimatorInterface). I could have moved the variables into the base class but I feel like they don't belong there. I moved the calculations into the terrain estimator where they are executed at the same rate as the fusion happens. What do you think?

priseborough:
Sounds good

LorenzMeier:
@tumbili Can you update your PR and rebase?

tumbili:
@LorenzMeier This was already merged some time ago. Closing.

PR #234, opened 2017-01-16
EKF: optical flow improvements

priseborough:
Partially addresses https://github.com/PX4/Firmware/issues/6347

1) Enables takeoff in position control mode when the OF sensor is out of focus and quality metric fails acceptance test by setting flow measurements to a value equivalent to zero horizontal velocity.

2) If optical flow is lost and then resumed in flight and there is no other means of determining position (eg GPS), the position will be set to the last known position when optical flow was lost.

Because I am running OSX Sierra, I have been unable to test the proposed ecl EKF changes using the gazebo sim as the gazebo installer is incompatible with my OS version.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -279,15 +279,21 @@ void Ekf::controlOpticalFlowFusion()
 						_state.pos(0) = 0.0f;
 						_state.pos(1) = 0.0f;
 
-						// reset the corresponding covariances
-						// we are by definition at the origin at commencement so variances are also zeroed
-						zeroRows(P,7,8);
-						zeroCols(P,7,8);
-
-						// align the output observer to the EKF states
-						alignOutputFilter();
+					} else {
+						// set to the last known position
+						_state.pos(0) = _last_known_posNE(0);
+						_state.pos(1) = _last_known_posNE(1);
 
 					}
+
+					// reset the corresponding covariances
+					// we are by definition at the origin at commencement so variances are also zeroed
+					zeroRows(P,7,8);
+					zeroCols(P,7,8);
+
+					// align the output observer to the EKF states
+					alignOutputFilter();
+
 				}
 			}
 
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -287,9 +287,10 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
 
 	// limit data rate to prevent data being lost
 	if (time_usec - _time_last_optflow > _min_obs_interval_us) {
-		// check if enough integration time
+		// check if enough integration time and fail if integration time is less than 50%
+		// of min arrival interval because too much data is being lost
 		float delta_time = 1e-6f * (float)flow->dt;
-		bool delta_time_good = (delta_time >= 0.05f);
+		bool delta_time_good = (delta_time >= 5e-7f * (float)_min_obs_interval_us);
 
 		// check magnitude is within sensor limits
 		float flow_rate_magnitude;
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -303,7 +303,7 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
 		// check quality metric
 		bool flow_quality_good = (flow->quality >= _params.flow_qual_min);
 
-		if (delta_time_good && flow_magnitude_good && flow_quality_good) {
+		if (delta_time_good && flow_magnitude_good && (flow_quality_good || !_control_status.flags.in_air)) {
 			flowSample optflow_sample_new;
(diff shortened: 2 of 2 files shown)


EKF/estimator_interface.cpp
```diff
@@ -303,16 +303,24 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
...
 			// copy the optical and gyro measured delta angles
-			optflow_sample_new.flowRadXY = - flow->flowdata;
 			optflow_sample_new.gyroXYZ = - flow->gyrodata;
+			if (flow_quality_good) {
+				optflow_sample_new.flowRadXY = - flow->flowdata;
+
+			} else {
+				// when on the ground with poor flow quality, assume zero ground relative velocity
+				optflow_sample_new.flowRadXY(0) = + flow->gyrodata(0);
+				optflow_sample_new.flowRadXY(1) = + flow->gyrodata(1);
```
jgoppert: You will get drift here because the flow gyro has bias. Best to just set it to 0.

EKF/control.cpp
```diff
@@ -279,15 +279,21 @@ void Ekf::controlOpticalFlowFusion()
...
-						// we are by definition at the origin at commencement so variances are also zeroed
-						zeroRows(P,7,8);
-						zeroCols(P,7,8);
-
-						// align the output observer to the EKF states
-						alignOutputFilter();
+					} else {
+						// set to the last known position
+						_state.pos(0) = _last_known_posNE(0);
+						_state.pos(1) = _last_known_posNE(1);
```
jgoppert: Resetting the state should be avoided. Better if you can stop estimating once the covariance becomes too large.

EKF/estimator_interface.cpp
priseborough: The EKF uses the angular motion compensated rate which is the optical rate minus the gyro rate. The drift will therefore be cancelled out.

EKF/control.cpp
priseborough: That's not a good idea given the federated architecture.  It needs to fuse in position measurements at the last known position to constrain angular drift.

EKF/estimator_interface.cpp
jgoppert: There are two different gyros sets though (onboard fmu and onboard px4flow). The flow gyros have a lot of bias and we can't estimate it without adding more states.

EKF/estimator_interface.cpp
```diff
@@ -303,16 +304,24 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
...
 			optflow_sample_new.quality = flow->quality;
 			// NOTE: the EKF uses the reverse sign convention to the flow sensor. EKF assumes positive LOS rate is produced by a RH rotation of the image about the sensor axis.
 			// copy the optical and gyro measured delta angles
-			optflow_sample_new.flowRadXY = - flow->flowdata;
 			optflow_sample_new.gyroXYZ = - flow->gyrodata;
+			if (flow_quality_good) {
+				optflow_sample_new.flowRadXY = - flow->flowdata;
+
+			} else {
+				// when on the ground with poor flow quality, assume zero ground relative velocity
```
ChristophTobler: I can't really see how it will be in this state, because of ```(flow_quality_good || _control_status.flags.in_air)```. If flow is bad, which is the case in this ```else```, it can't be on the ground because of the upper ```if``` statement (```_control_status.flags.in_air```)

ChristophTobler:
@priseborough With the change ```flow_quality_good || !_control_status.flags.in_air``` I get valid local positions and I can enable ```position``` mode, but ```commander takeoff``` does not work. I guess it's because the global position is never valid https://github.com/PX4/Firmware/blob/master/src/modules/commander/commander.cpp#L1205

EKF/estimator_interface.cpp
priseborough: Agree - will fix

priseborough:
@ChristophTobler The EKF cannot set global position as valid when operating with optical flow only because the WGS-84 origin of the local earth frame is not known.

jgoppert:
@priseborough in lpe I allow users to specify origin lat lon. Maybe something like that would be a good compromise.

priseborough:
@jgoppert rather than specifying it in the estimator, should we make it a system level parameter so that multiple algorithms can use it?

jgoppert:
@priseborough, totally agree, also for sensor noise specs would save a lot of space when we have time.

LorenzMeier:
@ChristophTobler @priseborough Where is this standing?

priseborough:
Discussions about adding externally set origins to estimators or modifying controllers to be able to operate in local coordinates should not be blocking this PR.

ChristophTobler:
@LorenzMeier I think this is good to go in

svpcom:
@priseborough Will it be merged to ecl and px4 in near future? Because current flow code in ekf2 is not usable without GPS fix

PR #238, opened 2017-02-05
Pr fw acro

RomanBapst:
This PR brings in acro support on the ecl side.

Changes:
1) Split rate control methods into
 - control_bodyrate() -> pure bodyrate control
 - control_euler_rate() -> control desired euler_rates (calls control_bodyrate())

2) Implemented coordinated turn logic using airspeed rather than body velocities.
See: https://books.google.ch/books?id=N4abCgAAQBAJ&pg=PA186&lpg=PA186&dq=coordinated+turn+constraint&source=bl&ots=5oupXeXFdB&sig=1MhdC7NUF5Uv9hUc6p-FB2JPBxg&hl=en&sa=X&ved=0ahUKEwj9sdzhxfnRAhXFvBoKHXUBCH4Q6AEIGjAA#v=onepage&q=coordinated%20turn%20constraint&f=false

3) Better naming of variables, e.g. body_x_rate instead of roll_rate when we mean body rates.

4) Commented sideslip control based on lateral acceleration measurements because implementation is not yet correct. Needs an integrator!

### DIFF
--- a/attitude_fw/ecl_controller.h
+++ b/attitude_fw/ecl_controller.h
@@ -56,9 +56,9 @@ struct ECL_ControlData {
 	float roll;
 	float pitch;
 	float yaw;
-	float roll_rate;
-	float pitch_rate;
-	float yaw_rate;
+	float body_x_rate;
+	float body_y_rate;
+	float body_z_rate;
 	float speed_body_u;
 	float speed_body_v;
 	float speed_body_w;
@@ -78,6 +78,7 @@ struct ECL_ControlData {
 	float groundspeed;
 	float groundspeed_scaler;
 	bool lock_integrator;
+	bool do_turn_compensation;
 };
 
 class __EXPORT ECL_Controller
@@ -87,6 +88,7 @@ public:
 	~ECL_Controller() = default;
 
 	virtual float control_attitude(const struct ECL_ControlData &ctl_data) = 0;
+	virtual float control_euler_rate(const struct ECL_ControlData &ctl_data) = 0;
 	virtual float control_bodyrate(const struct ECL_ControlData &ctl_data) = 0;
 
 	/* Setters */
@@ -96,6 +98,7 @@ public:
 	void set_k_ff(float k_ff);
 	void set_integrator_max(float max);
 	void set_max_rate(float max_rate);
+	void set_bodyrate_setpoint(float rate) {_bodyrate_setpoint = rate;};
 
 	/* Getters */
 	float get_rate_error();
--- a/attitude_fw/ecl_controller.h
+++ b/attitude_fw/ecl_controller.h
@@ -78,7 +78,6 @@ struct ECL_ControlData {
 	float groundspeed;
 	float groundspeed_scaler;
 	bool lock_integrator;
-	bool do_turn_compensation;
 };
 
 class __EXPORT ECL_Controller
--- a/attitude_fw/ecl_controller.h
+++ b/attitude_fw/ecl_controller.h
@@ -59,12 +59,6 @@ struct ECL_ControlData {
 	float body_x_rate;
 	float body_y_rate;
 	float body_z_rate;
-	float speed_body_u;
-	float speed_body_v;
-	float speed_body_w;
-	float acc_body_x;
-	float acc_body_y;
-	float acc_body_z;
 	float roll_setpoint;
 	float pitch_setpoint;
 	float yaw_setpoint;
--- a/attitude_fw/ecl_wheel_controller.cpp
+++ b/attitude_fw/ecl_wheel_controller.cpp
@@ -54,12 +54,12 @@ ECL_WheelController::ECL_WheelController() :
 
 float ECL_WheelController::control_bodyrate(const struct ECL_ControlData &ctl_data)
 {
-	/* Do not calculate control signal with bad inputs */
-	if (!(PX4_ISFINITE(ctl_data.yaw_rate) &&
-	      PX4_ISFINITE(ctl_data.groundspeed) &&
-	      PX4_ISFINITE(ctl_data.groundspeed_scaler))) {
-		return math::constrain(_last_output, -1.0f, 1.0f);
-	}
+    /* Do not calculate control signal with bad inputs */
+    if (!(PX4_ISFINITE(ctl_data.body_z_rate) &&
+          PX4_ISFINITE(ctl_data.groundspeed) &&
+          PX4_ISFINITE(ctl_data.groundspeed_scaler))) {
+        return math::constrain(_last_output, -1.0f, 1.0f);
+    }
 
 	/* get the usual dt estimate */
 	uint64_t dt_micros = ecl_elapsed_time(&_last_run);
@@ -76,8 +76,8 @@ float ECL_WheelController::control_bodyrate(const struct ECL_ControlData &ctl_da
 	/* input conditioning */
 	float min_speed = 1.0f;
 
-	/* Calculate body angular rate error */
-	_rate_error = _rate_setpoint - ctl_data.yaw_rate; //body angular rate error
+    /* Calculate body angular rate error */
+    _rate_error = _rate_setpoint - ctl_data.body_z_rate; //body angular rate error
 
 	if (!lock_integrator && _k_i > 0.0f && ctl_data.groundspeed > min_speed) {
 
--- a/attitude_fw/ecl_wheel_controller.cpp
+++ b/attitude_fw/ecl_wheel_controller.cpp
@@ -54,12 +54,12 @@ ECL_WheelController::ECL_WheelController() :
 
 float ECL_WheelController::control_bodyrate(const struct ECL_ControlData &ctl_data)
 {
-    /* Do not calculate control signal with bad inputs */
-    if (!(PX4_ISFINITE(ctl_data.body_z_rate) &&
-          PX4_ISFINITE(ctl_data.groundspeed) &&
-          PX4_ISFINITE(ctl_data.groundspeed_scaler))) {
-        return math::constrain(_last_output, -1.0f, 1.0f);
-    }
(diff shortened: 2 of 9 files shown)


tumbili:
@dagar This needs to be integrated with the changes you have prepared for the fw attitude control module in order to support acro.

attitude_fw/ecl_controller.h
```diff
@@ -56,9 +56,9 @@ struct ECL_ControlData {
 	float roll;
 	float pitch;
 	float yaw;
-	float roll_rate;
-	float pitch_rate;
-	float yaw_rate;
+	float body_x_rate;
+	float body_y_rate;
+	float body_z_rate;
```
dagar: p, q, r?

attitude_fw/ecl_yaw_controller.cpp
```diff
@@ -108,27 +89,36 @@ float ECL_YawController::control_attitude_impl_openloop(const struct ECL_Control
...
-//	static int counter = 0;
-	/* Calculate desired yaw rate from coordinated turn constraint / (no side forces) */
-	_rate_setpoint = 0.0f;
+	float constrained_roll;
+	bool inverted = false;
+
+	/* roll is used as feedforward term and inverted flight needs to be considered */
+	if (fabsf(ctl_data.roll) < math::radians(90.0f)) {
+		/* not inverted, but numerically still potentially close to infinity */
+		constrained_roll = math::constrain(ctl_data.roll, math::radians(-80.0f), math::radians(80.0f));
```
dagar: I think this needs to be limited further for the `tanf(constrained_roll)`, but what makes sense?

attitude_fw/ecl_controller.h
tumbili: @dagar fine with me, probably more consistent with the rest of the code base.

attitude_fw/ecl_yaw_controller.cpp
tumbili: @dagar On line 115 you will see that I constrained it to not be greater than the current roll setpoint. Numerically we should also be ok.

dagar:
Looking good. I'll do a simple acro implementation in the current fw_att_control for the sake of testing this and getting it in. 

I keep wondering if having this stuff in ECL is worth the effort. Calling these in the correct order while passing around a big struct that you update between calls is a bit of an awkward api.

What are your thoughts on the different coordination methods? Long term do you think we could settle on one (after some testing and comparison)?

attitude_fw/ecl_yaw_controller.cpp
dagar: Ah, good. Thanks

dagar:
Should we disable coordinated turns below a certain speed?

dagar:
This is the branch where I've been playing with a refactored fw_att_control (with and without ecl) - https://github.com/PX4/Firmware/compare/master...dagar:fw_att_control2?expand=1
It'll be much cleaner once all the manual_control_setpoints are pulled into stick_mapper.

I started with your ecl change, but then refactored and simplified to make them as simple as possible. I still think the entire thing would be much simpler pulled into fw_att_control, but this works.
https://github.com/PX4/ecl/compare/master...dagar:dagar_fw_att?expand=1

I had plain ACRO working, then I added attitude lock. RATTITUDE also works. I think the inverted case may need a closer look.

AndreasAntener:
@dagar I'd like to get this in soon together with Acro on the firmware side. Did you mean to do this with your refactoring? I'm not sure what you meant with your comment before about "simple acro implementation".

attitude_fw/ecl_wheel_controller.cpp
```diff
@@ -125,8 +121,7 @@ float ECL_WheelController::control_attitude(const struct ECL_ControlData &ctl_da
 	/* Calculate the error */
 	float yaw_error = ctl_data.yaw_setpoint - ctl_data.yaw;
 	/* shortest angle (wrap around) */
-	yaw_error = (float)fmodf((float)fmodf((yaw_error + M_PI_F), M_TWOPI_F) + M_TWOPI_F, M_TWOPI_F) - M_PI_F;
-	/*warnx("yaw_error: %.4f", (double)yaw_error);*/
+	yaw_error = (float)fmod((float)fmod((yaw_error + M_PI_F), M_TWOPI_F) + M_TWOPI_F, M_TWOPI_F) - M_PI_F;
```
dagar: Is there an existing wrap function?

dagar:
Sorry, should have been more clear. I have a big refactor of fw_att_control where I was able to use this, but I'm guessing no one wants to deal with a huge code review and significant testing at the moment.

So for right now I can implement the bare minimum acro mode (rate setpoint from sticks) in the current fw_att_control that works with this ecl update. Sound reasonable?

 Questions/Notes for this PR
  - Should we disable coordinated turns below a certain airspeed? FW_AIRSPEED_MIN perhaps?
  - is there anything specific we should validate in the logs and compare to the existing ecl? Was anyone actually using the existing coordinated turns?
  - do you think we actually need user configurable open or closed loop modes? Try we try both and keep one?
  - I think there are some problems in the control_attitude inverted case, but we can ignore it until we have something that might intentionally try control_attitude inverted. I'll do basic acro now, and acro + attitude lock later.
  - can you make the ECL_ControlData variable naming consistent and delete the unused fields?

attitude_fw/ecl_wheel_controller.cpp
AndreasAntener: There is one in geo but I guess we can't reference that

dagar: There's one in matrix. L1 is currently using _wrap_pi from geo.

AndreasAntener:
Thanks. On the implementation: we obviously have the simple implementation already since we've been flying this so we could push that. So should we add a PR for this?

I would prefer to do one big change after another. The argument for doing it together the refactoring would be combined testing. Because this here needs to be tested anyhow. How is the state of your FW attctl refactoring?

> Should we disable coordinated turns below a certain airspeed? FW_AIRSPEED_MIN perhaps?

@tumbili ?

> is there anything specific we should validate in the logs and compare to the existing ecl? Was anyone actually using the existing coordinated turns?

As far as we know no one was using the coordinated turn method, and if it would not be good as a reference. In the previous implementation, activating the coordinated turn would have resulted in having the pitch correction twice because of the so called turn compensation which actually did pretty much the same thing. Roman, please correct me here if I got it wrong.

> do you think we actually need user configurable open or closed loop modes? Try we try both and keep one?

I think it's a dev feature, and anyway not exposed to users. I remember you wanted to play with that, and possibly other methods? So this is up to you then I guess :)


> I think there are some problems in the control_attitude inverted case, but we can ignore it until we have something that might intentionally try control_attitude inverted. I'll do basic acro now, and acro + attitude lock later.

We took a shortcut and limited the turn compensation to non-inverted flights. It has numerical issues near 90degree roll. I think this is fine because it is our opinion that this is not wanted in Acro. For attitude locking I'd suggest not locking into an inverted attitude ;). Then it should be fine too. Thoughs?

> can you make the ECL_ControlData variable naming consistent and delete the unused fields?

Sure we can have another pass there. The only cleaner way of doing this passing values around would be to have separate structs for each of the methods. These methods have their interface and we won't get around to feed them the data.

dagar:
> Thanks. On the implementation: we obviously have the simple implementation already since we've been flying this so we could push that. So should we add a PR for this?

Yes that's probably fastest.

> I would prefer to do one big change after another. The argument for doing it together with the refactoring would be combined testing. Because this here needs to be tested anyhow. How is the state of your FW attctl refactoring?

By the time I replaced everything with controllib and split the header, implementation, and launcher into separate files nearly every single line has changed. The hard part is making sure I didn't break the little features I don't use. I also find the incremental refactor PRs tends to be ignored (ahem https://github.com/PX4/Firmware/pull/6530). Maybe if I expanded more CI testing with thorough fw_att coverage people would be more comfortable pushing big changes through?

> I think it's a dev feature, and anyway not exposed to users. I remember you wanted to play with that, and possibly other methods? So this is up to you then I guess :)

Only problem with a "dev feature" is that it shows up alongside the non-"dev feature" params that a PX4 FW or VTOL user needs to edit in the full param tree. Maybe one day that won't be the case in QGC.

> We took a shortcut and limited the turn compensation to non-inverted flights. It has numerical issues near 90degree roll. I think this is fine because it is our opinion that this is not wanted in Acro. For attitude locking I'd suggest not locking into an inverted attitude ;). Then it should be fine too. Thoughs?

That's a good portion of my interest here. I'm also tempted to implement MAV_CMD_DO_INVERTED_FLIGHT

Looking forward to trying this outside of the simulator. Thanks guys!

AndreasAntener:
> Yes that's probably fastest.

Will do.

>  I also find the incremental refactor PRs tends to be ignored

At the end of the day, contributors need to go out and flight test. Especially for big refactoring. Sure, CI would help a lot initially.

The roadmap should also guide this a bit. I don't know what the plan is at the moment (@LorenzMeier ?) but from what I see going on currently we'd need to get the multirotor rework finished for a new release and then take care of fixed-wing refactoring?

> Only problem with a "dev feature" is that it shows up alongside the non-"dev feature" params

Yep, UI problem ;)

> That's a good portion of my interest here. I'm also tempted to implement MAV_CMD_DO_INVERTED_FLIGHT

Cool! Inverted flight should not be a problem and the turn compensation can also be easily adjusted to work. The interesting part is where you cross 90 degrees. I'm curious with what solution you come up with. My thought is still that you disable it when you are in rate control (which for rattitude would mean to temporarily disable it when the user is on the sticks).

LorenzMeier:
Is this ready to go and tested, correct?

AndreasAntener:
Test flown once on this state, but we've used the same for a while now. Should be good to merge here but before we move to it on PX4 (#6567) a few more flights are needed.

PR #239, opened 2017-02-09
[WIP] Mag Validator

stephanbro:
This PR adds a data validator with specific checks for magnetometer data. The current error checking involves getting a GPS location, using lat and lon to get an inclination value from a lookup table, calculating the magnetometer's earth frame inclination, and generating an error with too high a difference between the two.

Additional changes:
* Adds the concept of "complex error" checking, which is a type specific error check and can be overridden by a derived class
* Moves `value[dimmensions]` to a protected member so it can be used by derived classes
* Check the state of complex errors when getting the validator's confidence
* Create validators based on a type enum, which currently only includes Magnetometer and Standard

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ px4_add_module(
 		l1/ecl_l1_pos_controller.cpp
 		validation/data_validator.cpp
 		validation/data_validator_group.cpp
+		validation/mag_validator.cpp
 		EKF/estimator_interface.cpp
 		EKF/ekf.cpp
 		EKF/ekf_helper.cpp
--- a/validation/data_validator.cpp
+++ b/validation/data_validator.cpp
@@ -43,6 +43,7 @@
 #include <ecl/ecl.h>
 
 DataValidator::DataValidator(DataValidator *prev_sibling) :
+	_value{0.0f},
 	_error_mask(ERROR_FLAG_NO_ERROR),
 	_timeout_interval(20000),
 	_time_last(0),
@@ -54,7 +55,6 @@ DataValidator::DataValidator(DataValidator *prev_sibling) :
 	_lp{0.0f},
 	_M2{0.0f},
 	_rms{0.0f},
-	_value{0.0f},
 	_vibe{0.0f},
 	_value_equal_count(0),
 	_value_equal_count_threshold(VALUE_EQUAL_COUNT_DEFAULT),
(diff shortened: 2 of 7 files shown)


LorenzMeier:
Nice! Is this still WIP or good to go?

stephanbro:
@LorenzMeier The concept in general is working, but I was hoping to get someone check out how I'm getting the global frame mag here: https://github.com/PX4/ecl/pull/239/files#diff-15e2f40ae61f93b78f65ab464dbfb0fdR93

I'm currently seeing large errors when I'm tilting, so I was hoping to have some one verify I'm doing things correctly.

LorenzMeier:
@priseborough Can you pitch in here?

priseborough:
The is nothing obviously wrong with the way you are rotating from body to earth frame given my understanding of the definition for the vehicle_attitude quaternion. Neither the vehicle_attitude topic quaternion or math library conjugate function  are adequately documented, so you could try  using the conjugate_inversed function instead. 

Comparing inclination to database value is a method that I have attempted to use before, but the presence of local ground based magnetic anomalies gave too many false alarms for the method to be useful, so the check will need to be optional with an adjustable sensitivity.

It also had the limitation that it could easily miss large errors in XY components that gave a correct horizontal field strength, but caused large heading errors at takeoff.

An alternative pre-flight magnetometer integrity check would require the user to pick up the vehicle and rotate it in heading through a minimum angle change (as detected by the IMU). It would then  check the magnetic heading innovations from the estimator. This is unlikely to be acceptable for a consumer product, but for a professional user used to following a pre-flight procedure, it would be  a more reliable method and less sensitive subject to ground based field anomalies.

stephanbro:
Thanks for the info @priseborough. I will try to using the conjugate_inversed and see if that helps the issue I'm seeing.

I agree about the need for the check to be optional or overrideable at least, as well as the ability to tweak the window of the expected inclination. As for the sensitivity of the measurement, I've had good success in the past using an inclination check with a similar setup so I believe it at least has the potential to detect errors without being a hindrance with false alarms.

priseborough:
I would prefer that this is not implemented in the ecl library and is moved to the PX4/Firmware/modules/commander/PreflightCheck.cpp instead where it can be used regardless of which type of estimator the codebase is running. When you do so, please provide a single parameter enabling the sensitivity to be adjusted and the check to be turned off (eg by setting the parameter to a negative number).

PR #244, opened 2017-02-28
Pr range aid

tumbili:
@priseborough This is a first shot at implementing range finder aid when the primary height source is not range finder.
I moved all the control logic for height source selection into one function. Also I have a function which decides if conditions are met to switch to range finder. I still need to implement better checks, e.g. figure out if terrain height is stable, ...
I did some flight tests on Friday but did some changes after. So I'll get new logs and post them here.

This PR also included two bug fixes:
1) baro offset filtering was not working
2) resetting height based on range finder was not correct

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -256,6 +256,10 @@ struct parameters {
 	float rng_gnd_clearance{0.1f};		// minimum valid value for range when on ground (m)
 	float rng_sens_pitch{0.0f};		// Pitch offset of the range sensor (rad). Sensor points out along Z axis when offset is zero. Positive rotation is RH about Y axis.
 	float range_noise_scaler{0.0f};		// scaling from range measurement to noise (m/m)
+	float vehicle_variance_scaler{0.0f};	// gain applied to vehicle height variance used in calculation of height above ground observation variance
+	float max_hagl_for_range_aid{5.0f};	// maximum height above ground for which we allow to use the range finder as height source (if range_aid == 1)
+	float max_vel_for_range_aid{1.0f};	// maximum ground velocity for which we allow to use the range finder as height source (if range_aid == 1)
+	int range_aid{0};			// allow switching primary height source to range finder if certian conditions are met
 
 	// vision position fusion
 	float ev_innov_gate{5.0f};		// vision estimator fusion innovation consistency gate size (STD)
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -259,7 +259,7 @@ struct parameters {
 	float vehicle_variance_scaler{0.0f};	// gain applied to vehicle height variance used in calculation of height above ground observation variance
 	float max_hagl_for_range_aid{5.0f};	// maximum height above ground for which we allow to use the range finder as height source (if range_aid == 1)
 	float max_vel_for_range_aid{1.0f};	// maximum ground velocity for which we allow to use the range finder as height source (if range_aid == 1)
-	int range_aid{0};			// allow switching primary height source to range finder if certian conditions are met
+	int32_t range_aid{0};			// allow switching primary height source to range finder if certian conditions are met
 	float range_aid_innov_gate{1.0f}; 		// gate size used for innovation consistency checks for range aid fusion
 
 	// vision position fusion
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -260,6 +260,7 @@ struct parameters {
 	float max_hagl_for_range_aid{5.0f};	// maximum height above ground for which we allow to use the range finder as height source (if range_aid == 1)
 	float max_vel_for_range_aid{1.0f};	// maximum ground velocity for which we allow to use the range finder as height source (if range_aid == 1)
 	int range_aid{0};			// allow switching primary height source to range finder if certian conditions are met
+	float range_aid_innov_gate{1.0f}; 		// gate size used for innovation consistency checks for range aid fusion
 
 	// vision position fusion
 	float ev_innov_gate{5.0f};		// vision estimator fusion innovation consistency gate size (STD)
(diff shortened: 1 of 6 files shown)


LorenzMeier:
@priseborough Can you review / rebase?

priseborough:
Will look at it today

EKF/control.cpp
```diff
@@ -760,7 +757,40 @@ void Ekf::controlRangeFinderFusion()
 		}
 
 		_fuse_height = true;
+	}
 
+
+}
+
+bool Ekf::rangeAidConditionsMet()
```
priseborough: This function needs to apply some hysteresis to the range and velocity check.

tumbili:
@priseborough Thanks for the review!  I agree, a hysteresis is needed and I will add it.

ChristophTobler:
@tumbili nice! I was thinking of something similar. I tested it in SITL (`gazebo_iris_opt_flow`) and it worked once I commented https://github.com/PX4/ecl/blob/d92da8f46da684c344f9031d68b0b5d359ce5993/EKF/control.cpp#L777 because `_fault_status.value` was never `0`. I couldn't quite see what it is...

tumbili:
range_aid_mod

tumbili:
@priseborough I added a hysteresis to the horizontal velocity and the height as discussed above.
The upper plot shows the results form a test flight in SITL. The first arrow shows the switch from range fusion to baro fusion as soon as the drone starts to move too fast. The  second arrow shows the switch when the maximum height is exceeded.
Finally, I placed a box under the drone in gazebo to test the innovation check. The behavior is nice, since there is not a significant height chance observable, however, it might still be necessary to add a hysteresis to the innovation check. You can see that the switch from using range to baro and back happens quite fast (3. and 4. arrow).

ChristophTobler:
@tumbili I have tested the PR with https://github.com/PX4/Firmware/pull/6989 and it seemed to work nicely. https://github.com/PX4/Firmware/pull/6989

LorenzMeier:
Good to merge?

priseborough:
I'm happy with the hysteresis changes

ChristophTobler:
I wanted to test it yesterday but the range finder didn't work... Have to check why. I can fit in a test today

EKF/common.h
```diff
@@ -349,6 +353,10 @@ struct parameters {
 		rng_gnd_clearance = 0.1f;
 		rng_sens_pitch = 0.0f;
 		range_noise_scaler = 0.0f;
+		max_hagl_for_range_aid = 5.0f;
+		max_vel_for_range_aid = 1.0f;
```
ChristophTobler: maybe use parameters instead?

tumbili:
@ndepal I pushed some more changes which were likely causing your problem. The baro offset calculation which is done when baro is not fused was not correct.
I'm still not happy with how I determine if the terrain is stable to use for range fusion. At the moment I'm just checking the terrain estimator innovations but those can be very noisy and can cause rapid switching between baro and range sensor. I need a better way to determine if the terrain is stable, e.g. by computing a filtered derivative of the terrain state. I'll think about this and report back.

ChristophTobler:
I added a check for spikes and abrupt distance changes which updates the offset. I think, this is a much smoother experience when using a rangefinder.
With this and https://github.com/PX4/ecl/pull/274, the range aid mode works well (only tested in gazebo with a ramp).

tumbili:
@ChristophTobler Thanks! Let's give this a flight test on a real drone ;-) I also fixed another bug after my last test flight and the results from @ndepal

priseborough:
Any updates on testing of this?

LorenzMeier:
@priseborough Would you mind commenting on the current state here?

priseborough:
It's been reviewed, but needs testing repeated following the changes made on the 12th May. We are looking to verify that it seamlessly switches between the two height sources and uses range finder for an accurate drift free height hold when below 3.5m in a hover. The effect of slow walking pace flight over steps and ramps also needs to be checked.

The intended behaviour is that when hovering, or in a vertical ascent or descent close to the ground, that the distance to ground be used as the primary heigh reference to assist with accurate ascent/descent control.

If no-one else is available to test it I can fit a lidar-lite to a quad.

ChristophTobler:
More on this can be found [here](https://github.com/PX4/Firmware/pull/6989). 
@priseborough  We've just built a new quad with a lidar-lite v3 and can do another test. But I guess it would be good if you could test it as well

priseborough:
Something I just thought of reading the comments on height deltas at change over is that if the estimator has a height innovation offset caused by vertical accel error, then adjusting the offset for the sensor being switched to so there is a zero innovation at switchover  will result in a change in a height adjustment after switchover. 

This can also happen when the sensor is noisy if the switchover offset is based on a single sample. The noise effect can be mitigated by continuously calculating the offset using a LPF when the sensor is not being used.

I will be interested to see how much these effects show during testing.

tumbili:
@priseborough I had exactly the problem you described during my SITL testing with gazebo. For some reason(maybe timing of the simulator) I always had large non-zero innovations when using gps altitude as primary height source even though gps altitude was equal to the ground truth altitude.  As a result the vehicle would continuously loose or gain altitude when switching to range finder height fusion.

priseborough:
It could be avoided somewhat by adjusting the height sensor offset for the innovation when switching to range finder fusion

priseborough:
However any algorithm using single event point samples is still vulnerable to measurement spikes on switchover. If we run the terrain estimator at all times then we could use the terrain vertical position to set the range finder height offset at switchover.

priseborough:
I've checked the code and the terrain estimator is now running continuously when range finder measurements are being received.

priseborough:
Testing over a range of objects ranging from 0.5 to over 3m in height at varying speeds was successful.

[06_59_25.ulg.pdf](https://github.com/PX4/ecl/files/1080059/06_59_25.ulg.pdf)

PR #245, opened 2017-03-02
Reduce EKF prediction delta time jitter

priseborough:
The EKF currently takes the IMU data from the sensor_combined topic at a nominal time delta of 4 msec (250Hz). The output predictor updates the output states every time it receives new IMU data from the sensor _combined message.

To avoid running the EKF covariance prediction at a small time step that is computationally expensive and can cause problems with rounding errors, the IMU data is accumulated and down sampled to an average time step of 10 msec.

The algorithm used to do this and maintain the correct average rate means that the number of 4msec IMU frames accumulated could vary between 1 and 4, which created timing jitter inside the EKF.  This can make tuning process noise levels harder.

The changes in this PR significantly reduce the variability in the EKF time step by:

1) Making the target EKF rate an integer multiple of the IMU rate. This slightly increases the average prediction tie step for the EKF from just over 10msec to 12msec, but the variation reduces significantly which makes filter tuning more deterministic.
2) Changing the algorithm used to adjust the collection time criteria provide a smoother correction.

This has been tested on replay with negligible effect on the innovation levels:

Replay summary before:

INFO  [ekf2_replay] GPS vel innov RMS =  0.078
INFO  [ekf2_replay] GPS pos innov RMS =  0.083
INFO  [ekf2_replay] Hgt innov RMS =  0.551
INFO  [ekf2_replay] Mag innov RMS = 0.0073

Replay summary after:

INFO  [ekf2_replay] GPS vel innov RMS =  0.076
INFO  [ekf2_replay] GPS pos innov RMS =  0.084
INFO  [ekf2_replay] Hgt innov RMS =  0.552
INFO  [ekf2_replay] Mag innov RMS = 0.0073

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -84,7 +84,7 @@ Ekf::Ekf():
 	_time_last_arsp_fuse(0),
 	_time_last_beta_fuse(0),
 	_last_disarmed_posD(0.0f),
-	_last_dt_overrun(0.0f),
+	_imu_collection_time_adj(0.0f),
 	_airspeed_innov(0.0f),
 	_airspeed_innov_var(0.0f),
 	_beta_innov(0.0f),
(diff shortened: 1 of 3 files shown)


LorenzMeier:
@priseborough Can you rebase or is this superseded?

priseborough:
This has been superseded, because the commit was included in PR #248

PR #247, opened 2017-03-09
EKF: don't accept GPS data without a 3D lock

priseborough:
This fixes a bug that allows the EKF to try using GPS data when there is no longer a 3D fix. This prevents the timeout reversion back to non-GPS operation that is supposed to occur after 10 seconds of no GPS lock


LorenzMeier:
Checks fail

PR #254, opened 2017-03-17
EKF: improve recovery from bad IMU accel data

priseborough:
This PR uses vertical velocity and position innovation consistency check failure to detect bad accelerometer data caused by clipping or aliasing which can cause large vertical acceleration errors and loss of height estimation. When bad accel data is detected it

1) Reports bad accel status using bit position 11 in the estimator solution status message.
2) Forces fusion of vertical velocity and height data by bypassing innovation consistency checks to reduce the vertical velocity and position error.
3) Inhibits accelerometer bias learning to prevent accel bias estimates being corrupted by forced fusion.
4) Increases accelerometer process noise to reduce effect of bad accel data on state estimates.

**Testing**

Testing on replay logs from vehicle with significant vibration induced clipping has been completed.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -506,6 +506,7 @@ union ekf_solution_status {
 	uint16_t pred_pos_horiz_rel : 1; // 8 - True if the EKF has sufficient data to enter a mode that will provide a (relative) position estimate
 	uint16_t pred_pos_horiz_abs : 1; // 9 - True if the EKF has sufficient data to enter a mode that will provide a (absolute) position estimate
 	uint16_t gps_glitch         : 1; // 10 - True if the EKF has detected a GPS glitch
+	uint16_t accel_error        : 1; // 11 - True if the EKF has detected bad accelerometer data
     } flags;
     uint16_t value;
 };
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -169,6 +169,11 @@ struct extVisionSample {
 #define RNG_MAX_INTERVAL	2e5
 #define EV_MAX_INTERVAL		2e5
 
+// bad accelerometer detection and mitigation
+#define BADACC_PROBATION	10E6	// Number of usec that accel data declared bad must continuously pass checks to be declared good
+#define BADACC_HGT_RESET	1E6	// Number of usec that accel data must continuously fail checks to trigger a height reset
+#define BADACC_BIAS_PNOISE_MULT 2.0f	// The delta velocity process noise is multiplied by this value when accel data is declared bad
+
 struct parameters {
 	// measurement source control
 	int fusion_mode;		// bitmasked integer that selects which of the GPS and optical flow aiding sources will be used
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -172,7 +172,7 @@ struct extVisionSample {
 // bad accelerometer detection and mitigation
 #define BADACC_PROBATION	10E6	// Number of usec that accel data declared bad must continuously pass checks to be declared good
 #define BADACC_HGT_RESET	1E6	// Number of usec that accel data must continuously fail checks to trigger a height reset
-#define BADACC_BIAS_PNOISE_MULT 2.0f	// The delta velocity process noise is multiplied by this value when accel data is declared bad
+#define BADACC_BIAS_PNOISE	4.9f	// The delta velocity process noise is set to this when accel data is declared bad (m/s**2)
 
 struct parameters {
 	// measurement source control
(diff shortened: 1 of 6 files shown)


ChristophTobler:
I have a log which should be useful http://logs.px4.io/plot_app?log=ef7ccff4-05a3-4bba-bc3d-34ad4a741f14 for replay. At around 250s there is the problem of bad acc_z data. The quad thinks it is going down even though it is going up.

I have seen this phenomenon on a Pixracer and Snapdragon now.

ChristophTobler:
@priseborough The log is from a Snapdragon Flight and I used this https://github.com/PX4/Firmware/pull/6969 filtering

priseborough:
Replay testing has been completed.

Here is a replay analysis of a log from a copter where the hard-mounted sensor was rattling and causing clipping of the Z accelerometer.

**Z acceleration** - note Z accel rises to -5 m/s/s average at 1110 seconds:
screen shot 2017-04-07 at 11 27 58 am

**Before Plots** - These plots are before the patches in the PR have been applied.

**Vertical Innovations** - note rapid divergence in height before the 5 second timeout resets states to Baro and GPS:
vertical innovations - before

**IMU Delta Velocity Bias State Variance** - note the number of times of this being set to zero which indicates the high angular vibration is causing the accel bias to be inhibited to reduce estimation errors.
dvel bias variance - before2

**IMU Delta Velocity Bias Estimates** - estimation is stable.
dvel bias - before2

**After Plots** This is the same replay, but with the PR applied.

**Vertical Innovations** - note significant reduction in height divergence:
vertical innovations - after2

**IMU Delta Velocity Bias State Variance** - note the longer period around 1110 seconds where the bad accel check has been triggered and the variances are bing held at zero to protect the bias estimate form the effects of the bad accel data.
dvel bias variance - after2

**IMU Delta Velocity Bias Estimates** - estimation is stable an unaffected by the changes.
dvel bias - after2

priseborough:
@ChristophTobler - thanks for getting us that log. I have  tested it with a similar improvement:

**Vertical Innovations - Before**



**Vertical Innovations - After**

priseborough:
@ChristophTobler - thanks for getting us that log. I have  tested it with a similar improvement. The maximum height error reduced from 63 to 19m. Unfortunately Github will not let me attach additional plots.

priseborough:
I will now see if earlier intervention in the accel process noise increase and a larger increase can reduce the height gain further.

priseborough:
Here are the vertical innovations before applying commit 7b996c5

vert innov - after

Here they are after:
vert innov - after - retuned

The larger value of delta velocity process noise used when the bad accel is detected results in a further reduction in innovation from the reset until the clipping stops at 360 seconds.

ChristophTobler:
@priseborough http://logs.px4.io/plot_app?log=b0618fe1-d97f-4aea-8f4f-dc013251ffee here's another one from a Pixracer.

I have two frames where this happens regularly. A Snapdragon Flight and a Pixracer, both on racer frames. Both are dampened. The Pixracer uses the dampening material from a Pixhawk mini (and before on a different dampening material) and the Snapdragon uses a custom dampening system (see pics).

Would you recommend harder dampening instead?

pixracerdampening
snapdragon-setup

priseborough:
1) Foam material with significant hysteresis (damping) can transmit high levels of vibration at high frequency 
2) That frame design will have very lightly damped structural modes - the motors may be exciting them - try a three bladed prop to move the blade passage frequency away from any modes.
3) When it comes to isolating  high frequency motor/prop vibration you want to keep the natural frequency of the isolation system well below the motor and blade passage frequency => softer mounts. 

- For the pixracer, try using four 10x10mm squares of this foam https://hobbyking.com/en_us/anti-vibration-foam-orange-latex-190mm-x-140mm-x-6mm.html.
- For the snapdragon I would mount it on a subframe and  mount the subframe to the vehicle chassis using the latex foam. The open cell urethane foam is a poor isolator, and compressed as you have it there, it will be ineffective.
- all interconnect cables need to have enough loop in them to avoid transmitting vibration.

priseborough:
@LorenzMeier for info

ChristophTobler:
@priseborough I'll try that, thanks!

ChristophTobler:
@priseborough different dampening seemed to have fixed the issue on the Snapdragon, thanks. I guess we could need some dampening DOs and DON'Ts on the devguide.

LorenzMeier:
Still open?

ChristophTobler:
My issues were fixed by replacing the damping

PR #256, opened 2017-03-27
EKF: Add ability to use stored mag bias estimates

priseborough:
This PR adds parameters that will enable learned mag bias estimates to be stored for use at the next startup. The logic to check and save the parameter values based on filter status reporting will be implemented externally to the ecl library.


priseborough:
I think this can be moved outside of the ecl library into the ekf2 module

PR #257, opened 2017-03-31
EKF: Do not move EKF origin when performing GPS checks

priseborough:
This fixes a bug that was introduced when GPS checks were enabled to run whenever the vehicle was on the ground by this PR https://github.com/PX4/ecl/pull/255 . This resulted in the EKF origin being continually being updated to match the GPS position when on ground.

Doing this can cause large jumps in GPS position and generate innovation check errors after landing and also reduces the usefulness of pre-flight GPs position innovation consistency checks.

Moving the origin is unnecessary because the origin is already performed here https://github.com/PX4/ecl/blob/master/EKF/gps_checks.cpp#L68  after all enabled GPS checks have passed.

### DIFF
--- a/EKF/gps_checks.cpp
+++ b/EKF/gps_checks.cpp
@@ -167,10 +167,6 @@ bool Ekf::gps_is_good(struct gps_message *gps)
 		_gps_check_fail_status.flags.hdrift = false;
 	}
 
-	// Save current position as the reference for next time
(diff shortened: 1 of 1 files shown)


priseborough:
Analysis from two test flights with the same power cycle where the copter was landed at a different point to takeoff to test the effect of the change.

[flight_04_05.ulg.pdf](https://github.com/PX4/ecl/files/887032/flight_04_05.ulg.pdf)

PR #258, opened 2017-04-06
EKF: Critical Bug Fix

priseborough:
The previous practice of relying on the off-diagonals being zero and maintaining the diagonals (variances) caused problems with conditioning of the covariance matrix on one flight. This resulted in the Y-axis magnetometer fusion failing the numerical health test and the covariance matrix being reset in mid flight. 

This change stores the variances when the learning inhibit becomes active and ensures that the rows and columns in the covariance matrix for the inhibited states are explicitly maintained at zero. When the delta velocity bias learning inhibit is released the variances are re-instated.

During replay of the problem flight log, this eliminated the in-air covariance reset event.


priseborough:
Flight test data pending

priseborough:
This was closed because of a covariance reset event occurred during testing of the PR in conjunction with latest PX4/Firmware master. 

The log generated by this flight generated multiple unknown log message error in the ekf2_replay program (see example below) which needs to be fixed to complete testing.

ERROR [replay] unknown log message type 73, size 118 (offset 16031406)

priseborough:
replaced by https://github.com/PX4/ecl/pull/259

bkueng:
@priseborough you can ignore the `ERROR [replay] unknown log message type 73, size 118`, I'll push a fix for that.

PR #259, opened 2017-04-06
EKF: Protect against covariance prediction and update errors

priseborough:
The previous practice of relying on the off-diagonals being zero and maintaining the diagonals (variances) caused problems with conditioning of the covariance matrix on one flight. This can result in the magnetometer fusion failing the numerical health test and the covariances being reset in mid flight.

This change stores the variances when the learning inhibit becomes active and ensures that the rows and columns in the covariance matrix for the inhibited states are explicitly maintained at zero. When the delta velocity bias learning inhibit is released the variances are re-instated.

It also protects the covariance prediction from excessively small or large delta time values that could poor numerical conditioning.

During replay of a problem flight log, this eliminated the in-air covariance reset event seen here when the state variances for the magnetic field states reset:

reset

Here is the replay without the fix:
reset replayed

Here is the replay with the fix:
reset fixed

**Flight Test Results**

This was tested on a pixracer with rapid attitude and throttle inputs made in stabilise mode to activate the delta velocity bias learning inhibit.

IMU delta velocity bias states - normal convergence
dvelbiasxyz

IMU delta velocity bias state variances - note the variances drop to zero and are re-instated when the inhibit activates and clears.
dvelbiasvarxyz

Magnetomer bias states - normal convergence
magbiasxyz

Magnetometer bias state variances - normal convergence
magbiasvarxyz

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -143,7 +143,7 @@ void Ekf::predictCovariance()
 	float dvy_b = _state.accel_bias(1);
 	float dvz_b = _state.accel_bias(2);
 
-	float dt = _imu_sample_delayed.delta_ang_dt;
+	float dt = math::constrain(_imu_sample_delayed.delta_ang_dt, 0.0005f * FILTER_UPDATE_PERIOD_MS, 0.002f * FILTER_UPDATE_PERIOD_MS);
 
 	// compute noise variance for stationary processes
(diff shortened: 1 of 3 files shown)


priseborough:
@tumbili , @devbharat , @yyassa , @bkueng  for your information.

Can one of you please review.

Thanks

PR #261, opened 2017-04-12
Add wind estimation for multi-rotors

priseborough:
This PR implements the fusion method derived here: https://github.com/PX4/ecl/blob/master/matlab/scripts/Inertial%20Nav%20EKF/GenerateNavFilterEquations.m#L425

It enables the drag of a multi-rotor measured perpendicular to the thrust axis to be used to form aa rough estimate of wind velocity. It has been tested on replay and flight test.

Here are the XY acceleration innovations and wind velocity estimates from a flight test with around 5m/s wind swinging between SW SSW.

XY drag accel innovations
screen shot 2017-04-12 at 1 42 37 pm

Vehicle and Wind Velocity
At 535 and 565 seconds, the vehicle was allowed to drift downwind in attitude control mode to provide a check on wind drift. The vehicle speed asymptotically approaches the wind speed estimate and the direction of drift is within 25 deg of the wind direction.
screen shot 2017-04-12 at 1 36 48 pm

### DIFF
--- a/EKF/drag_fusion.cpp
+++ b/EKF/drag_fusion.cpp
@@ -50,8 +50,7 @@ void Ekf::fuseDrag()
 	float Kfusion[24] = {}; // Kalman gain vector
 	float R_ACC = _params.drag_noise; // observation noise in specific force drag (m/sec**2)
 
-	float eas2tas = 1.0f; // conversion factor from EAS to TAS - TODO calculate from barometric pressure
-	float rho = 1.225f / sqrtf(eas2tas); // air density (kg/m**3)
+	float rho = fmaxf(_air_density, 0.1f); // air density (kg/m**3)
 
 	// calculate inverse of ballistic coefficient
 	if (_params.bcoef_x < 1.0f || _params.bcoef_y < 1.0f) {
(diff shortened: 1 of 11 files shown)


EKF/drag_fusion.cpp
```diff
@@ -0,0 +1,304 @@
...
+	float vn = _state.vel(0);
+	float ve = _state.vel(1);
+	float vd = _state.vel(2);
+
+	// get latest wind velocity in earth frame
+	float vwn = _state.wind_vel(0);
+	float vwe = _state.wind_vel(1);
+
+	// predicted specific forces
+	// calculate relative wind velocity in earth frame and rotte into body frame
```
tumbili: typo

EKF/drag_fusion.cpp
```diff
@@ -0,0 +1,304 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2015 Estimation and Control Library (ECL). All rights reserved.
```
tumbili: @priseborough Maybe 2017?

tumbili:
@priseborough Reviewed, looks good! Just found two minor things.

PR #265, opened 2017-04-24
EKF: add option to output velocity as time derivative of position

priseborough:
This PR adds a parameter selectable option to output a more kinematically position and velocity. This will have degraded tracking relative to the EKF states, but the velocity will be closer to the first derivative of the position and reduce the effect of inertial prediction errors on control loops that are operating in a pure velocity feedback mode.

Results show that this significantly reduces steady state velocity estimation errors due to inertial sensor bias, but at the expense of an increased effect of baro disturbances on vertical velocity error and therefore height tracking.

The option is disabled by default.

These changes have been tested on replay and a single flight test.

**Vertical Velocity Consistency**
The following plots show the local vertical velocity (black), time derivative of the local vertical position (green) and GPS vertical velocity (red) generated using  _params.link_vel_pos = 0 (the default) and  _params.link_vel_pos = 1 which activates the option introduced by this PR. It can be seen that with the option enabled, the local output vertical velocity is closer to the derivative of the local vertical position, but does not match the GPs velocity as well.

_params.link_vel_pos = 0
vert vel - default

_params.link_vel_pos = 1
vert vel - option

** EKF state tracking
The following plots show the norm of the output predictor velocity (cyan) and position (magenta) tracking errors relative the EKF at the EKF fusion time horizon. It can be seen that the option degrades the tracking.

_params.link_vel_pos = 0
tracking - default

_params.link_vel_pos = 1
tracking - option

**UPDATE**

After discussion with the position control developers, the PR has been modified so that the revised algorithm is performed in parallel to the legacy algorithm, but only for the vertical states and the vertical position derivative is made available via the ecl EKF interface. This removes the need for a parameter and enables the controllers to decide which vertical velocity measure estimate is the best to use. The impact on resources of estimating the extra two states in the output predictor was negligible with a 1% rise in CPU utilisation from 54% to 55%. See plots here https://github.com/PX4/ecl/pull/265#issuecomment-297364600

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -113,6 +113,11 @@ public:
 		return _buffer[_tail];
 	}
 
+	unsigned get_oldest_index()
+	{
+		return _tail;
+	}
+
 	inline data_type get_newest()
 	{
 		return _buffer[_head];
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -79,9 +79,16 @@ struct ext_vision_message {
 
 struct outputSample {
 	Quaternion  quat_nominal;	// nominal quaternion describing vehicle attitude
-	Vector3f    vel;	// NED velocity estimate in earth frame in m/s
-	Vector3f    pos;	// NED position estimate in earth frame in m/s
-	uint64_t 	time_us;	// timestamp in microseconds
+	Vector3f    vel;		// NED velocity estimate in earth frame in m/s
+	Vector3f    pos;		// NED position estimate in earth frame in m/s
+	uint64_t    time_us;		// timestamp in microseconds
+};
+
+struct outputVert {
+	float	    vel_d;		// D velocity calculated using alternative algorithm
+	float	    vel_d_integ;	// Integral of vel_d
+	float	    dt;			// delta time in seconds
+	uint64_t    time_us;		// timestamp in microseconds
 };
 
 struct imuSample {
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -478,10 +478,10 @@ bool Ekf::collect_imu(imuSample &imu)
 */
 void Ekf::calculateOutputStates()
 {
-	// use latest IMU data
+	// Use full rate IMU data at the current time horizon
 	imuSample imu_new = _imu_sample_new;
 
-	// correct delta angles for bias offsets and scale factors
+	// correct delta angles for bias offsets
 	Vector3f delta_angle;
 	float dt_scale_correction = _dt_imu_avg / _dt_ekf_avg;
 	delta_angle(0) = _imu_sample_new.delta_ang(0) - _state.gyro_bias(0) * dt_scale_correction;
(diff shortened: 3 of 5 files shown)


priseborough:
Here are the same plots from a flight test using a pixracer board on a 250 size quad with the control parameter set to 1.

vert vel - flight test

tracking - flight

MaEtUgR:
@priseborough thanks for your efforts! It looks like it's actually flyable but has some unwanted side effects/is just a different compromise...
- Using the velocity of option 0 is good for closed loop velocity control because it has better tracking but not so good for absolute conditions because it has more bias.
- Using the velocity of option 1 is good for absolute conditions (e.g. check for < 0 for backwards and what not) because there's mch less bias but the performance for closed loop control is worse.

Right?

Would it be possible to use both outputs for their dedicated cases? The actual EKF state does not differ right? Is this a bad idea in terms of complexity? Do you have a better idea?

priseborough:
Switching between output observer mode in flight is easy to do and can be achieved by changing the parameter value (but not saving it of course).

I expect that option 1 will be more sensitive to baro disturbances (eg fast flight height variation), but would be a better option when landing where small velocity errors are a problem.

As short term fix we could switch to option 1 when the landing descent rate limit is active. My preference longer term is that PX4 position controller moves to tracking a reference point at all times.

priseborough:
It is also possible to run both calculations in parallel and have two outputs published, a velocity and a position derivative. The controllers could then select what they wanted. This will increase CPU load by a few percent and also increase RAM by up to 500bytes due to the need to add 6 additional states to the output buffer.

MaEtUgR:
ok, so we have all possibilities open.
If I have to decide what to use then I would need to do comparison tests... If you already know what would work well feel free to adopt.

priseborough:
rebased

mzahana:
Barometer doesn't work very well low altitude or close to ground. So wouldn't using positions derivatives based on height reported by barometer make it worse than using accelerometer? 
If a range sensor, e.g. Lidar, is used, then using the second option would improve in most cases because of the improved height accuracy, no?
Also as a side note, is the velocity calculated based on finite differences of positions only? I remember that there are more accurate methods than just simple finite differences. I will try to find some references and share later.
Thanks for the efforts.

priseborough:
It is not as simple as using accelerometer vs using baro or lidar. The ecl  EKF combines all sources of data using assumptions about the error characteristics of each sensor. This fusion runs on a delayed time horizon using buffered sensor data so that all measurement are fused at the correct time. The resulting state estimates then need to be propagated forward to the current time horizon. This is done by the output predictor.

The output predictor applies a strapdown INS algorithm to IMU data at the current time horizon to calculate quaternion, velocity and position output states. The output states are buffered and compared to the EKF states at the fusion time horizon and a correction is then applied to the output state buffer so that the output states track the EKF states.

This also filters the observation noise which are the small steps in EKF states produced whenever a GPS, magnetometer or baro observation is fused. 

The exisiting algorithm minimises tracking error relative to EKF states and independently corrects the output velocity ad position history. The alternate algorithm in this PR corrects the output velocity history first, then integrates the corrected velocity to recalculate the position history. This ensures that the velocity output is the second derivative of the position, but does not track the EKF states as closely.

priseborough:
@MaEtUgR Can you think of a way that the ekf2_main module could tell if the position controller was controlling vertical velocity directly and in a mode that was sensitive to offsets in estimated vertical velocity?

priseborough:
@mzahana Using range finder close to the ground is a good option if that sensor is available, however the effect of rotor wash on static pressure in ground effect can be compensated for in the estimator for a limited time sufficient for takeoff and landing if it is told by the control loops that landing is or takeoff is expected. This techniques is used successfully by ardupilot.

mzahana:
@priseborough Thanks for the explanation. Is that Implemented for height only or, xy too?

priseborough:
x, y and z

MaEtUgR:
@priseborough Technically that's definitely possible but wouldn't that end up in a terrible hack if the estimator adjusts its behaviour depending on the controllers current properties (controller ouput to feed the model is something different). I think that could very well end up in unjustifiable complexity and unpredictable system reactions. How do you see that?

priseborough:
I agree - Interactions with the height control loop are a concern. The  switch would need to have a minimum  interval and be based on data that did not change with vehicle response.

The second approach is to run both calculations in parallel and publish a z_dot in the vehicle_local_position and vehicle_global_position messages. The controllers can then make the decision as to what velocity they wish to use. If I modify the predictor to only do the alternative calculation using the vertical states, then the impact on processing time and RAM will be reduced.

I will prototype the second approach and quantify the impact on resources.

priseborough:
With the output observer reworked to only apply the new calculation to the vertical states, the impact on reported RAM and CPU usage is negligible.

**before**
before

**after**
after

priseborough:
Here is a plot of the two vehicle _local_position vertical velocity estimates (cyan is the original vz and magenta is the newly added vertical derivative value) together with the first order difference calculated time derivative of the z position. The z_deriv value closely tracks the z position derivative.

screen shot 2017-04-27 at 11 49 09 am

LorenzMeier:
We should target a call next Tuesday to discuss this and the proper implementation on the control side.

priseborough:
OK

priseborough:
Rebased and squashed. Note that this PR no longer changes default behaviour - it adds an additional vertical position derivative output that can be used by the flight controllers if required. It increases CPU utilisation by 1%.

priseborough:
We need to make a call on whether this functionality should be included. It is a zero impact change in terms of autopilot behaviour because it produces an output that is currently un-used, however it does cause a 1% increase in CPU utilisation so there is no point in calculating this output unless it is going to be used.

I can have the calculation enabled by parameter so that by default, so that there is effectively no impact on CPU  utilisation, however this will add to code complexity.

LorenzMeier:
I would enable by default and start to sequence it into the control loops. The kinematically correct output is important from a controller perspective.

priseborough:
I've rebased and logged the derivative output using https://github.com/priseborough/Firmware/tree/pr-ekfOutputObserver. Here is a comparison of the various height rate measures including a post processing generated first order derivative of the lpos.z output

screen shot 2017-06-12 at 6 54 53 pm

PR #267, opened 2017-04-25
EKF: Increase sensitivity and add tuning of bad accel checks

priseborough:
These changes implement an improved method of detecting the loss of vertical estimation accuracy due to vibration induced accelerometer clipping. The statistical confidence check is not a combined check using the product of the vertical velocity and position innovations and tuneable via a single confidence parameter. The amount of failure time required to trigger a state reset is also adjustable.

This has resulted in reductions in the time required to recover from clipping. The following plots show the vertical velocity and position innovation results from replay testing of a flight log with a clipping event that starts at 622.5 seconds. 

**Before**
before

**After**
after


priseborough:
Squashed.

priseborough:
rebased

PR #270, opened 2017-04-29
EKF: fix reporting of optical flow fusion errors

priseborough:
Optical flow fusion numerical errors were being written to the wrong index in _innov_check_fail_status.


priseborough:
Here is the EKF output from a test flight

[log_15_2017-4-29-11-54-32.ulg.pdf](https://github.com/PX4/ecl/files/965833/log_15_2017-4-29-11-54-32.ulg.pdf)

PR #272, opened 2017-05-10
EKF: Improve control of IMU delta velocity bias learning

priseborough:
This enables IMU delta velocity bias learning to be enabled for individual axes for applications where the biases may be unobservable for extended periods of time. An example may be a multi-rotor or ground vehicle operating in a very still position hold for extended periods of time. An algorithm to dynamically set the components will be added later. 

Other minor amendments include:

- Minor cleanup of the covariance prediction comments.
- Removal of unnecessary variable copy operations.

Here is an example from testing with the mask parameter set to 4 which enables only the Z component of bias learning:

screen shot 2017-05-11 at 8 01 28 am


priseborough:
I have run into unexplained height variations during flight testing, so have closed it until these are resolved.

PR #275, opened 2017-05-15
EKF: Prevent long periods of flight without movement causing yaw drift

priseborough:
This PR modifies the logic that us used to automatically select whether the magnetometer heading or 3-axis measurement fusion method is used. The previous logic always used the 3-axis method once the vehicle had started flying, however this can cause problems for vehicle that enter into stable position hold for extended periods immediately after takeoff.

The revised logic uses horizontal acceleration to check if yaw is observable independent of the magnetometer and uses  about the vertical to check if mag bias states are observable.
If neither yaw of mag biases are observable, it saves the magnetic field variances, the learned declination and switch to magnetic yaw fusion. When yaw or biases become observable, the saved variances are reinstated and it and switches back to 3D mag fusion.

Here is a replay of a flight where the vehicle hovered for an extended periodi n 3-axis mag fusion mode which resulted in the mag bias states drifting, which led to eventual loss of heading and a flyaway with large velocity innovations and repeated resets back to GPS before the EKF eventually recovered.

XY mag bias estimates
screen shot 2017-05-16 at 9 33 39 am

NE velocity innovations
screen shot 2017-05-16 at 9 39 59 am

Here are the same plots after the changes:

XY mag bias estimates - estimates only update wen the vehicle has movement.
screen shot 2017-05-16 at 9 41 15 am

NE velocity innovations
screen shot 2017-05-16 at 9 42 51 am

Here are the mag field variances after the changes - variances are reinstated when movement checks indicate observability.

screen shot 2017-05-16 at 9 44 38 am

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -888,9 +888,10 @@ void Ekf::controlMagFusion()
 	}
 
 	// If we are on ground, store the local position and time to use as a reference
+	// Also reset the flight alignment falg so that the mag fields will be re-initialised next time we achieve flight altitude
 	if (!_control_status.flags.in_air) {
 		_last_on_ground_posD = _state.pos(2);
-
+		_flt_mag_align_complete = false;
 	}
 
 	// checs for new magnetometer data tath has fallen beind the fusion time horizon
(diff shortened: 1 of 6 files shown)


priseborough:
I have completed a flight test of a multi-rotor in still air conditions with 1000sec of static hover from takeoff, followed by 200 seconds of manoeuvring flight, followed by another 250 sec of static hover. 

Mag bias estimation doesn't start until hover finishes at 1000 sec:
screen shot 2017-05-16 at 12 11 55 pm

The mag heading and mag field innovations  show where it is switching between the heading and 3-axis fusion method:
screen shot 2017-05-16 at 12 14 30 pm

Velocity innovations remained low throughout:
screen shot 2017-05-16 at 12 13 11 pm

The XY IMU delta velocity bias estimates are unobservable during static hover and drift to incorrect values, however this does cause a problem because they recover immediately when the vehicle motion make them observable and have no adverse effect on the velocity innovations.
screen shot 2017-05-16 at 12 19 09 pm

priseborough:
EKF data from another flight with alternating periods of hover and manoeuvring.

[log001.ulg.pdf](https://github.com/PX4/ecl/files/1003425/log001.ulg.pdf)

priseborough:
I am happy for this to go in when someone has reviewed the code.

LorenzMeier:
@tumbili Can you review?

tumbili:
@LorenzMeier Yes!

EKF/control.cpp
```diff
@@ -888,9 +888,10 @@ void Ekf::controlMagFusion()
 	}
 
 	// If we are on ground, store the local position and time to use as a reference
+	// Also reset the flight alignment falg so that the mag fields will be re-initialised next time we achieve flight altitude
```
tumbili: typo

LorenzMeier:
Awesome, thanks! I'd like to get this into the 1.6 release as we've seen this causing real-world issues.

EKF/control.cpp
```diff
@@ -899,26 +900,81 @@ void Ekf::controlMagFusion()
...
+			// record the last time that movement was suitable for use of 3-axis magnetometer fusion
+			if (_mag_bias_observable || _yaw_angle_observable) {
+				_time_last_movement = _imu_sample_delayed.time_us;
+			}
+
+			// decide whether 3-axis magnetomer fusion can be used
+			bool use_3D_fusion = _control_status.flags.tilt_align && // Use of 3D fusion requires valid tilt estimates
+					_control_status.flags.in_air && // don't use when on the ground becasue of magnetic anomalies
+					(_flt_mag_align_complete || height_achieved) && // once in-flight field alignment has been performed, ignore relative height
+					((_imu_sample_delayed.time_us - _time_last_movement) < 2000000); // Using 3-axis fusion for a minimum period after to allow for false negatives
```
tumbili: @priseborough What do you think about using 2e6, I find it easier to read and prevent mistakes?

LorenzMeier:
So the review would be rather urgently needed 8).

EKF/ekf.cpp
```diff
@@ -479,6 +488,15 @@ void Ekf::calculateOutputStates()
...
 	delta_angle(2) = _imu_sample_new.delta_ang(2) - _state.gyro_bias(2) * dt_scale_correction;
 
+	// calculate a yaw change about the earth frame vertical
+	float spin_del_ang_D = _R_to_earth_now(2,0) * delta_angle(0) +
+			_R_to_earth_now(2,1) * delta_angle(1) +
+			_R_to_earth_now(2,2) * delta_angle(2);
+	_yaw_delta_ef += spin_del_ang_D;
+
+	// calculate filtered yaw rate to be used by the magnetomer fusion type selection logic
+	_yaw_rate_lpf_ef = 0.95f * _yaw_rate_lpf_ef + 0.05f * spin_del_ang_D / _imu_sample_new.delta_ang_dt;
```
tumbili: @priseborough Any reason for not using dynamic filter coefficients as you did above for the filtered acceleration?

EKF/ekf.h
```diff
@@ -266,6 +266,13 @@ class Ekf : public EstimatorInterface
 
 	matrix::Dcm<float> _R_to_earth;	// transformation matrix from body frame to earth frame from last EKF predition
 
+	// used by magnetomer fusion mode selection
+	Vector2f _accel_lpf_NE;			// Low pass filtered horizontal earth frame acceleration (m/s**2)
+	float _yaw_delta_ef{0.0f};		// Recent change in yaw angle measured about the earth frame D axis (rad)
+	float _yaw_rate_lpf_ef{0.0f};		// Filtered angular rate abut earth frame D axis (rad/sec)
```
tumbili: typo

EKF/ekf.h
```diff
@@ -266,6 +266,13 @@ class Ekf : public EstimatorInterface
...
+	bool _mag_bias_observable{false};	// true when there is enough rotation to make magnetomer bias errors observable
```
tumbili: typo

EKF/ekf.h
```diff
@@ -328,7 +335,11 @@ class Ekf : public EstimatorInterface
 	float _baro_hgt_offset{0.0f};		// baro height reading at the local NED origin (m)
 
 	// Variables used to control activation of post takeoff functionality
-	float _last_on_ground_posD{0.0f}; // last vertical position when the in_air status was false (m)
+	float _last_on_ground_posD{0.0f};	// last vertical position when the in_air status was false (m)
+	bool _flt_mag_align_complete{true};	// true when the in-flight mag field alignment has been completed
+	uint64_t _time_last_movement{0};	// last system time in usec that sufficient movement to use 3-axis magnetomer fusion was detected
```
tumbili: typo :-)

tumbili:
@LorenzMeier I had to review carefully since the changes are not trivial. I didn't find anything worrying.

EKF/control.cpp
priseborough: As long as it is interpreted as an integer by the compiler and we don't get a type conversion.

LorenzMeier: It would end up being a double, so no, that's not a good idea.

tumbili: @LorenzMeier So should we change this in all other locations as well? E.g. in control.cpp we used this a lot.

EKF/ekf.cpp
priseborough: I was trying to save operations because that step is operating at 250Hz and having an accurate time constant was not important.

tumbili: @priseborough Ok

EKF/control.cpp
priseborough: Is 2 * 1000 * 1000 acceptable?

tumbili: Yeah,  much better!

PR #276, opened 2017-05-26
EKF: Add feature to use the mag to only update the mag states

CarlOlsson:
FYI @priseborough 

@CarlOlsson I have added a brief explanation of why this change is necessary:

Flight of fixed wing vehicles with a misaligned magnetometer or significant soft iron effects can result in the roll and pitch estimates being biased. This causes incorrect learning of X and Y delta velocity bias states.

If the vehicle is a type that changes to a different orientation for landing (eg tailsitter), then the incorrectly learned bias states will create errors in velocity and position tracking during landing.

The fusion mode added by PR prevents this occurring, but still enables the earth and mag bias field states to be learned during up ad away flight to be used during landing to constrain heading drift if required.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -977,6 +977,18 @@ void Ekf::controlMagFusion()
 				_control_status.flags.mag_hdg = true;
 			}
 
+			// perform switch-over from only updating the mag states to updating all states
+			if (!_control_status.flags.update_mag_states_only && _control_status_prev.flags.update_mag_states_only) {
+				// reset the mag field covariances
+				zeroRows(P, 16, 21);
+				zeroCols(P, 16, 21);
+
+				// re-instate the last used variances
+				for (uint8_t index = 0; index <= 5; index ++) {
+					P[index+16][index+16] = _saved_mag_variance[index];
+				}
+			}
+
 		} else if (_params.mag_fusion_type == MAG_FUSE_TYPE_HEADING) {
 			// always use heading fusion
 			_control_status.flags.mag_hdg = true;
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -979,13 +979,13 @@ void Ekf::controlMagFusion()
 
 			// perform switch-over from only updating the mag states to updating all states
 			if (!_control_status.flags.update_mag_states_only && _control_status_prev.flags.update_mag_states_only) {
-				// reset the mag field covariances
+				// When re-commencing use of magnetometer to correct vehicle states
+				// set the field state variance to the observation variance and zero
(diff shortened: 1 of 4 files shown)


EKF/control.cpp
```diff
@@ -977,6 +977,18 @@ void Ekf::controlMagFusion()
...
 
+			// perform switch-over from only updating the mag states to updating all states
+			if (!_control_status.flags.update_mag_states_only && _control_status_prev.flags.update_mag_states_only) {
+				// reset the mag field covariances
+				zeroRows(P, 16, 21);
+				zeroCols(P, 16, 21);
+
+				// re-instate the last used variances
+				for (uint8_t index = 0; index <= 5; index ++) {
+					P[index+16][index+16] = _saved_mag_variance[index];
```
priseborough: We do need to reset the off-diagonals (co-variances), but further thought is needed on how the diagonals (variances) should be set. The _saved_mag_variance value was intended to save the variances when switching in and out of 3-axis fusion, however in this case where we have remained in3-axis fusion the learned mag field states are valid, so there needs to be a good reason for why we would not continue to use them.

priseborough:
There could be unintended interactions between this PR and the recently added functionality to switch in and out of 3-axis fusion. Replay testing on flight logs is required.

EKF/control.cpp
priseborough: After further thought I think that when we come out of the inhibit condition, if we are still in 3-axis fusion, then the variances should be set to alignment values.

priseborough:
That looks better. I will do reversion testing on this changes, but am not expecting any issues as the functionality is off by default.

priseborough:
We also need to think about how to handle reporting of the control status which is currently via the get_control_mode get_control_mode(uint16_t *val) accessor. At the moment the accessor is only copying the first 16 bits.

The accessor in estimator_interface.h will need to be updated and this will also require a corresponding change to ekf2_main.cpp when the upstream library reference is updated.

CarlOlsson:
God point. I updated the function in ecl

priseborough:
I've flight tested this using the following test bench which hardwired the feature on: https://github.com/priseborough/Firmware/tree/test_ecl_pr276-wip

Flight characteristics were unchanged and the log showed that filter operation was as expected.

[log001.ulg.pdf](https://github.com/PX4/ecl/files/1073238/log001.ulg.pdf)

priseborough:
This PR will require the PX4/Firmware estimator status topic to be modified to accomodate the increased size of the control mode flag, however we were already out of bits and the increased size is going to be necessary longer term as we add more measurement types.

PR #278, opened 2017-06-01
Fix compilation errors with GCC 7

jlecoeur:
See associated changes in https://github.com/PX4/Firmware/pull/7332

### DIFF
--- a/EKF/mag_fusion.cpp
+++ b/EKF/mag_fusion.cpp
@@ -149,9 +149,9 @@ void Ekf::fuseMag()
 		_mag_test_ratio[index] = sq(_mag_innov[index]) / (sq(math::max(_params.mag_innov_gate, 1.0f)) * _mag_innov_var[index]);
 		if (_mag_test_ratio[index] > 1.0f) {
 			healthy = false;
-			_innov_check_fail_status.value |= ((1 << (index + 3)) != 0);
+			_innov_check_fail_status.value |= (1 << (index + 3));
 		} else {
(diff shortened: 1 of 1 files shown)


EKF/mag_fusion.cpp
```diff
@@ -149,9 +149,9 @@ void Ekf::fuseMag()
 		_mag_test_ratio[index] = sq(_mag_innov[index]) / (sq(math::max(_params.mag_innov_gate, 1.0f)) * _mag_innov_var[index]);
 		if (_mag_test_ratio[index] > 1.0f) {
 			healthy = false;
-			_innov_check_fail_status.value |= (1 << (index + 3));
+			_innov_check_fail_status.value |= ((1 << (index + 3)) != 0);
 		} else {
-			_innov_check_fail_status.value &= !(1 << (index + 3));
+			_innov_check_fail_status.value &= !((1 << (index + 3)) != 0);
```
bkueng: That does not look correct. I think it should be `&= ~(1 << (index + 3));`

jlecoeur: You are right @bkueng ! 
I thought `_innov_check_fail_status` contained only booleans (hence the !=0 test), but `value` is an uint16_t.
I fixed it.

LorenzMeier:
@priseborough This looks like it could be a real issue and should be propagated.

CarlOlsson:
@LorenzMeier 
Yes this was a real issue.
Before this PR
image

After this PR
image

PR #281, opened 2017-06-09
EKF: Enable operation close to magnetic poles

priseborough:
**DESCRIPTION**

This PR enables operation close to the earths magnetic poles where the normal method of determining yaw orientation using the on-board compass is unreliable. It uses the following strategy:

1) A parameter mag_field_vertical is used to specify if the mode is active.
2) If the parameter mag_field_vertical is set to either 1 or 2, the vehicles yaw will be aligned to the value in degrees set by parameter mag_yaw_ground
3) In flight 3-axis fusion will be enabled if mag_field_vertical = 2, and no mag fusion of any type will be used if mag_field_vertical = 1. 
4) If the yaw variance is excessive, the current yaw estimate is fused as an observation to prevent a situation where a badly conditioned covariance matrix causes filter instability.
5) The last in-flight yaw angle estimate is saved and used to constrain the heading drift when on the ground so subsequent takeoffs can be performed after landing at arbitrary yaw angles

**TESTING**

Completed. See following comment.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -245,6 +245,10 @@ struct parameters {
 	int32_t mag_fusion_type{0};		///< integer used to specify the type of magnetometer fusion used
 	float mag_acc_gate{0.5f};		///< when in auto select mode, heading fusion will be used when manoeuvre accel is lower than this (m/sec**2)
 	float mag_yaw_rate_gate{0.25f};		///< yaw rate threshold used by mode select logic (rad/sec)
+	int32_t mag_field_vertical{0};		///< set to 1 or 2 when the earth field is too close to vertical to give reliable heading.
+						///< A value of 1 will assume yaw = mag_yaw_ground when on the ground and do no yaw or mag fusion in air.
+						///< A value of 2 will assume yaw = mag_yaw_ground when on the ground and do 3-axis mag fusion in the air.
+	float mag_yaw_ground{0.0f};		///< This yaw angle will be used for alignment and fusion when on the ground if mag_field_vertical = 1 or 2 (deg)
 
 	// airspeed fusion
 	float tas_innov_gate{5.0f};		///< True Airspeed innovation consistency gate size (STD)
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1230,6 +1230,14 @@ void Ekf::controlMagFusion()
 			_control_status.flags.mag_3D = false;
 		}
 
+		// If we are not fusing in observations to constrain yaw variance growth, and yaw variance has become large,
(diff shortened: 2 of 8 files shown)


priseborough:
Flight testing has been completed with mag_yaw_ground = 90.0 and mag_field_vertical = 1 and 2. The PX4/Firmware patches required to modify the parameters are pushed to here: https://github.com/priseborough/Firmware/tree/ekfHighLatMag-wip

EKF2_EF_VERT = 1
[EKF2_EF_VERT=1.ulg.pdf](https://github.com/PX4/ecl/files/1063042/EKF2_EF_VERT.1.ulg.pdf)

EKF2_EKF_VERT = 2
[EKF2_EF_VERT=2.ulg.pdf](https://github.com/PX4/ecl/files/1063050/EKF2_EF_VERT.2.ulg.pdf)

priseborough:
@CarlOlsson are you able to have a look at this? It has been tested, but because of the amount of change I would like someone else to review the code.

priseborough:
I've rebased.

priseborough:
It looks like we are out of parameter space in the parameter block, so I will need to do some house keeping before this can be pulled in.

kechaofan:
@priseborough What if we don't have any magnetometer and the _mag_data_ready will always be false? 
Should we add some parameter to distinguish between this kind of situation and disable the _mag_data_ready  check?

kechaofan:
I have met another problem. I set mag_field_vertical = 1 to disable the magnetometer and re-calibrate the gyro before flying. But the ekf2 was likely to produce a wrong gyro_bias_z and make yaw angle shifting faster than it should be. What can i do about that?

priseborough:
@kechaofan  Did you move it whilst on ground? The vehicle needs to be static from startup until flight if you want to fly without a magnetometer. As a general rule if you are doing sensor calibrations, then you must restart the autopilot before flying.

kechaofan:
@priseborough I didn't move it whilst on ground. And i did follow the calibration rules with FC restart.
Since there is no magnetometer, I was wondering why EKF2 can still produce a non-zero gyro_bias_z. All the sensors I have are imu, barometer, range finder and optical flow.Which one will contribute to gyro_bias_z or Is this caused by a badly conditioned covariance matrix? If so, what can i do about that?

priseborough:
It assumes a constant yaw angle while on ground.

priseborough:
The PR adds the ability to not fuse magnetometer data, however the logic is still tied to arrival of data. It has not been tested with no sensor fitted. Supporting that case will require the patch to be modified.

priseborough:
Closed so that it can be rebased after PR #322 has been merged.

priseborough:
Replaced by https://github.com/PX4/ecl/pull/335

PR #283, opened 2017-06-19
EKF matrix typedef cleanup

dagar:
The EKF common.h typedefs were conflicting with using the typedefs in Matrix. It became an issue when trying to use the typedefs in the WIP PX4 wind_estimator.


priseborough:
Please provide a link to a corresponding PR on the PX4 firmware side with the change required to use this. 

ekf2_main.cpp does use the Quaternion type

dagar:
I believe Lorenz already made the change in PX4. For ekf2_main you can use the matrix types directly, either with the nice typedefs (matrix::Quatf) or the full template specialization matrix::Quaternion<float>.

PR #286, opened 2017-06-23
Pr range aid fixes

ChristophTobler:
Here are some fixes that should solve the remaining issues when using range aid.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -479,6 +479,17 @@ void Ekf::controlHeightSensorTimeouts()
 	// check if height has been inertial deadreckoning for too long
 	bool hgt_fusion_timeout = ((_time_last_imu - _time_last_hgt_fuse) > 5e6);
 
+	// if in range aid mode, check faultiness that otherwise would never change back
+	if (_params.range_aid) {
+		// check if range finder data is available
+		rangeSample rng_init = _range_buffer.get_newest();
+		_rng_hgt_faulty = ((_time_last_imu - rng_init.time_us) > 2 * RNG_MAX_INTERVAL);
+
+		// check if GPS height is available
+		gpsSample gps_init = _gps_buffer.get_newest();
+		_gps_hgt_faulty = ((_time_last_imu - gps_init.time_us) > 2 * GPS_MAX_INTERVAL);
+	}
+
 	// reset the vertical position and velocity states
 	if ((P[9][9] > sq(_params.hgt_reset_lim)) && (hgt_fusion_timeout || continuous_bad_accel_hgt)) {
 		// boolean that indicates we will do a height reset
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -508,7 +508,6 @@ void Ekf::controlHeightSensorTimeouts()
 			if (reset_to_gps) {
 				// set height sensor health
 				_baro_hgt_faulty = true;
-				_gps_hgt_faulty = false;
 
 				// declare the GPS height healthy
 				_gps_hgt_faulty = false;
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -730,9 +730,11 @@ void Ekf::controlHeightFusion()
 				}
 			}
 
-		} else if (_baro_data_ready && !_baro_hgt_faulty && !_in_range_aid_mode) {
+		} else if (_baro_data_ready && !_baro_hgt_faulty &&
+			         !(_in_range_aid_mode && !_range_data_ready && !_rng_hgt_faulty)) {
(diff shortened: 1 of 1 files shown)


ndepal:
Trying to follow this, I have a couple of questions.

In one commit, you handle the case of "in range aid mode and hgt mode is range", I don't really think that makes sense. Shouldn't it just be that, when using rng for hgt, the range aid flag is ignored?

EKF/control.cpp
```diff
@@ -479,6 +479,17 @@ void Ekf::controlHeightSensorTimeouts()
...
 
+	// if in range aid mode, check faultiness that otherwise would never change back
+	if (_params.range_aid) {
+		// check if range finder data is available
+		rangeSample rng_init = _range_buffer.get_newest();
+		_rng_hgt_faulty = ((_time_last_imu - rng_init.time_us) > 2 * RNG_MAX_INTERVAL);
+
+		// check if GPS height is available
+		gpsSample gps_init = _gps_buffer.get_newest();
+		_gps_hgt_faulty = ((_time_last_imu - gps_init.time_us) > 2 * GPS_MAX_INTERVAL);
```
ndepal: Why is only the check for GPS and not for baro needed here?

EKF/control.cpp
```diff
@@ -720,9 +730,10 @@ void Ekf::controlHeightFusion()
 				}
 			}
 
-		} else if (_baro_data_ready && !_baro_hgt_faulty && !_in_range_aid_mode) {
+		} else if (_baro_data_ready && !_baro_hgt_faulty && _rng_hgt_faulty) {
```
ndepal: I don't understand why a check for the range height is needed here.

E.g. what happens if not in range aid mode but the ranger is good (`_rng_hgt_faulty == false`)? It seems like then it would end up not fusing anything?

EKF/control.cpp
```diff
@@ -756,9 +787,10 @@ void Ekf::controlHeightFusion()
 				}
 			}
 
-		} else if (_gps_data_ready && !_gps_hgt_faulty && !_in_range_aid_mode) {
+		} else if (_gps_data_ready && !_gps_hgt_faulty && _rng_hgt_faulty) {
```
ndepal: Same as in the baro check. The decision on whether we can switch to GPS should not depend on whether the ranger is bad.

ChristophTobler:
@ndepal 

> In one commit, you handle the case of "in range aid mode and hgt mode is range", I don't really think that makes sense. Shouldn't it just be that, when using rng for hgt, the range aid flag is ignored?

That's wrong in the commit message. If you have a look at https://github.com/PX4/ecl/pull/286/commits/c382af7a7d2214b111b3dae2c7dd78f1ce0a8997#diff-4cac31c333867233bb8a08de1e812a9aR762 you can see that it just switches to baro independent on range aid. The highlighted line could/should go away.

EKF/control.cpp
ChristophTobler: Because baro checks are already included in every mode

EKF/control.cpp
ChristophTobler: That's true! Needs to be changed

ChristophTobler: should be
```
else if (_baro_data_ready && !_baro_hgt_faulty &&
!(_in_range_aid_mode && !_range_data_ready && !_rng_hgt_faulty))
```
to prevent switching to baro if in range aid, rng is not faulty but not ready

EKF/control.cpp
ChristophTobler: same as for GPS

EKF/control.cpp
```diff
@@ -730,9 +730,11 @@ void Ekf::controlHeightFusion()
 				}
 			}
 
-		} else if (_baro_data_ready && !_baro_hgt_faulty && !_in_range_aid_mode) {
+		} else if (_baro_data_ready && !_baro_hgt_faulty &&
```
ndepal: I still don't understand why any check regarding the range sensor is needed in this if statement.
Whether the ranger can be used is checked in the if statement above. Now we're just trying to figure out if baro can be used. This has nothing to do with the range sensor or the range aid feature.
What situation are you taking care of with this?

ChristophTobler: Range data is not available every iteration (_range_data_ready = false). Because of that it would switch back and forth between range finder and baro.

ndepal: Ah! Great, thank you

ChristophTobler:
@tumbili can you verify?

tumbili:
@ChristophTobler The changes make sense to me.

PR #288, opened 2017-06-29
EKF: Improve protection against covariance instability in delta velocity bias states

priseborough:
1) Variances for each axis  are increased to prevent the ratio of the largest to smallest value exceeding 100. Large ratios can result in poor stability of the delta velocity bias state covariance prediction and update operations.

2) The covariance prediction step explicitly constrains the variance from falling below the target minimum of 1E-8 which is equivalent to about 0.01 m/sec^2 of accel bias uncertainty. 

These changes improve the protection against badly conditioned IMU delta velocity bias state variances. This is necessary due to the limitations associated with single precision arithmetic that limits the range of values that can be used for the EKF2_ACC_B_NOISE parameter. The order and rate that different sensors are fused changes the minimum safe value for this parameter meaning that the safe minimum parameter value varies between board types. This requires active protection measures.

The symptons of badly conditioned delta velocity bias variances are:

1) The accel bias variance for the sensor axis aligned with the vertical collapses to the minimum value allowed by the ConstrainVariances() function.
2) Delta velocity bias states  fail to converge on correct values and in extreme cases can even go to the limits imposed by the ConstrainStates()  function.

This is most likely to happen during static operation.

**TESTING**

http://logs.px4.io/plot_app?log=2c193c1a-beda-4913-915c-8599ce3ccf9b

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -212,7 +212,7 @@ struct parameters {
 
 	// process noise
 	float gyro_bias_p_noise{1.0e-3f};	// process noise for IMU rate gyro bias prediction (rad/sec**2)
-	float accel_bias_p_noise{3.0e-3f};	// process noise for IMU accelerometer bias prediction (m/sec**3)
+	float accel_bias_p_noise{6.0e-3f};	// process noise for IMU accelerometer bias prediction (m/sec**3)
 	float mage_p_noise{1.0e-3f};		// process noise for earth magnetic field prediction (Guass/sec)
 	float magb_p_noise{1.0e-4};		// process noise for body magnetic field prediction (Guass/sec)
 	float wind_vel_p_noise{1.0e-1f};		// process noise for wind velocity prediction (m/sec/sec)
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -743,24 +743,25 @@ void Ekf::fixCovarianceErrors()
 
 		// To ensure stability of the covariance matrix operations, the ratio of a max and min variance must
 		// not exceed 100 and the minimum variance must not fall below the target minimum
+		// Also limit variance to a maximum equivalent to a 1g uncertainty
 		const float minStateVarTarget = 1E-8f;
 		float minAllowedStateVar = fmaxf(0.01f * maxStateVar, minStateVarTarget);
 		for (uint8_t stateIndex=13; stateIndex<=15; stateIndex++) {
-		    P[stateIndex][stateIndex] = math::constrain(P[stateIndex][stateIndex], minAllowedStateVar, sq(10.0f * _dt_ekf_avg));
+			P[stateIndex][stateIndex] = math::constrain(P[stateIndex][stateIndex], minAllowedStateVar, sq(_gravity_mss * _dt_ekf_avg));
 		}
 
 		// If any one axis has fallen below the safe minimum, all delta velocity covariance terms must be reset to zero
 		if (resetRequired) {
 		    float delVelBiasVar[3];
 		    // store all delta velocity bias variances
-		    for (uint8_t i=0; i<=2; i++) {
-			delVelBiasVar[i] = P[i+13][i+13];
+		    for (uint8_t stateIndex=13; stateIndex<=15; stateIndex++) {
+			delVelBiasVar[stateIndex-13] = P[stateIndex][stateIndex];
 		    }
 		    // reset all delta velocity bias covariances
 		    zeroCols(P,13,15);
 		    // restore all delta velocity bias variances
-		    for (uint8_t i=0; i<=2; i++) {
-			P[i+13][i+13] = delVelBiasVar[i];
+		    for (uint8_t stateIndex=13; stateIndex<=15; stateIndex++) {
+			P[stateIndex][stateIndex] = delVelBiasVar[stateIndex-13];
 		    }
 		}
 
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -729,11 +729,42 @@ void Ekf::fixCovarianceErrors()
 		zeroRows(P,13,15);
 		zeroCols(P,13,15);
 	} else {
-		// constrain variances
-		for (int i = 13; i <= 15; i++) {
-			P[i][i] = math::constrain(P[i][i], 0.0f, P_lim[4]);
+		// Find the maximum delta velocity bias state variance and request a covariance reset if any variance is below the safe minimum
+		const float minSafeStateVar = 1e-9f;
+		float maxStateVar = minSafeStateVar;
+		bool resetRequired = false;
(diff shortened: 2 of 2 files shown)


priseborough:
@CarlOlsson Can you have a look at this please. TXS.

CarlOlsson:
Yes, will review and test next week!

CarlOlsson:
Hi Paul,
You get constant messages that ekf2 is low on stack in the logfile from the testing of this pr, is this expected behavior?

EKF/covariance.cpp
```diff
@@ -729,11 +729,42 @@ void Ekf::fixCovarianceErrors()
...
-		for (int i = 13; i <= 15; i++) {
-			P[i][i] = math::constrain(P[i][i], 0.0f, P_lim[4]);
+		// Find the maximum delta velocity bias state variance and request a covariance reset if any variance is below the safe minimum
+		const float minSafeStateVar = 1e-9f;
+		float maxStateVar = minSafeStateVar;
+		bool resetRequired = false;
+		for (uint8_t stateIndex=13; stateIndex<=15; stateIndex++) {
+		    if (P[stateIndex][stateIndex] > maxStateVar) {
+			maxStateVar = P[stateIndex][stateIndex];
+		    } else if (P[stateIndex][stateIndex] < minSafeStateVar) {
```
CarlOlsson: Is the if statement here necessary? If P[stateIndex][stateIndex] > maxStateVar is false then P[stateIndex][stateIndex] < minSafeStateVar must be true. (Except for the case when P[stateIndex][stateIndex] == maxStateVar )

EKF/common.h
```diff
@@ -208,7 +208,7 @@ struct parameters {
 
 	// process noise
 	float gyro_bias_p_noise{1.0e-3f};	// process noise for IMU rate gyro bias prediction (rad/sec**2)
-	float accel_bias_p_noise{3.0e-3f};	// process noise for IMU accelerometer bias prediction (m/sec**3)
```
CarlOlsson: Why is this still needed if we add the protection in covariance.cpp?

EKF/common.h
```diff
@@ -208,7 +208,7 @@ struct parameters {
...
+	float accel_bias_p_noise{6.0e-3f};	// process noise for IMU accelerometer bias prediction (m/sec**3)
```
CarlOlsson: Why is this still needed if we add the protection in covariance.cpp?

EKF/covariance.cpp
```diff
@@ -729,11 +729,42 @@ void Ekf::fixCovarianceErrors()
...
+		float minAllowedStateVar = fmaxf(0.01f * maxStateVar, minStateVarTarget);
+		for (uint8_t stateIndex=13; stateIndex<=15; stateIndex++) {
+		    P[stateIndex][stateIndex] = math::constrain(P[stateIndex][stateIndex], minAllowedStateVar, sq(10.0f * _dt_ekf_avg));
+		}
+
+		// If any one axis has fallen below the safe minimum, all delta velocity covariance terms must be reset to zero
+		if (resetRequired) {
+		    float delVelBiasVar[3];
+		    // store all delta velocity bias variances
+		    for (uint8_t i=0; i<=2; i++) {
```
CarlOlsson: please use "stateIndex" as above for consistency, or change to i above

EKF/covariance.cpp
```diff
@@ -729,11 +729,42 @@ void Ekf::fixCovarianceErrors()
...
+			resetRequired = true;
+		    }
+		}
+
+		// To ensure stability of the covariance matrix operations, the ratio of a max and min variance must
+		// not exceed 100 and the minimum variance must not fall below the target minimum
+		const float minStateVarTarget = 1E-8f;
+		float minAllowedStateVar = fmaxf(0.01f * maxStateVar, minStateVarTarget);
+		for (uint8_t stateIndex=13; stateIndex<=15; stateIndex++) {
+		    P[stateIndex][stateIndex] = math::constrain(P[stateIndex][stateIndex], minAllowedStateVar, sq(10.0f * _dt_ekf_avg));
```
CarlOlsson: Could you please explain the "sq(10.0f * _dt_ekf_avg)" term?

priseborough:
No, the low stack space is a recent development that occurred after I rebased. I'll increase the allocation by 100 bytes.

EKF/covariance.cpp
priseborough: Suppose that the first axis  (stateIndex 13) was 1e-5. This would set maxStateVar to 1e-5.  Suppose the next axis checked was 1e-8 (safe). We do not want to trigger a reset yet. Suppose the last axis checked was 1e-10 (unsafe). Now we want to trigger a reset.

EKF/common.h
priseborough: Because based on a number of logs there is not enough process noise to enable the bias learning to follow faster temperature induced bias changes during flight. The ideal tuning value is where without protection, it would fall to a value around the target minimum and the protection is there as a backup.

EKF/covariance.cpp
priseborough: will fix

EKF/covariance.cpp
priseborough: Will fix

priseborough: It was limiting the variance to equivalent of ~1g,  I will use the gravity constant instead to make that clearer and add a comment.

EKF/covariance.cpp
priseborough: stateIndex = i+13

priseborough: I will change it to use state index and use stateIndex-13 to access the delVelBiasVar array

priseborough:
I have rebased and tested the updates on a miniature quad using optical flow with range finder as the primary height source. This has been the configuration that has previously caused covariance collapse due to higher measurement rates and lower observation noise parameter used by the range finder height sensor. The RAM allocated in the ekf2_main.cpp was increased from 5800 to 5900 for this test. 

http://logs.px4.io/plot_app?log=283a8d69-9438-43d9-b0d7-33885026c7a5

delta velocity bias estimates converge and are stable with this tuning

screen shot 2017-07-04 at 8 49 23 am

All covariance values were constrained by the limit at times during the flight. When the vehicle was in a static hover, the X and Y variances did slowly start to increase from the minimum, so there was sufficient process noise to achieve that, however for this sensor configuration, a large process noise variance could be used.

screen shot 2017-07-04 at 8 50 02 am

EKF/common.h
priseborough: This is going to be hard to achieve for all setups because the fusion mode, observation rates and observation noise parameters used all affect how rapidly the variances reduce. The value of 0.06 is still too small for some setups, but we cannot make it large without having noisy delta velocity bias estimates for others.

EKF/covariance.cpp
CarlOlsson: true true

PR #289, opened 2017-06-29
Pr height fixes

ChristophTobler:
reset from baro to gps and from gps to baro did not exist in `ControlHeightFusion()` and gps was always faulty because `_gps_buffer` didn't have any measurements because they got remove with `pop_first_older_than()`

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -740,6 +740,16 @@ void Ekf::controlHeightFusion()
 			if (_control_status_prev.flags.baro_hgt != _control_status.flags.baro_hgt) {
 				_hgt_sensor_offset = 0.0f;
 			}
+		} else if (_control_status.flags.gps_hgt && _gps_data_ready && !_gps_hgt_faulty) {
+			// switch to gps if there was a reset to gps
+			_fuse_height = true;
+			_in_range_aid_mode = false;
+
+			// we have just switched to using gps height, calculate height sensor offset such that current
+			// measurment matches our current height estimate
+			if (_control_status_prev.flags.gps_hgt != _control_status.flags.gps_hgt) {
+				_hgt_sensor_offset = _gps_sample_delayed.hgt - _gps_alt_ref + _state.pos(2);
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


EKF/control.cpp
```diff
@@ -740,6 +737,16 @@ void Ekf::controlHeightFusion()
...
 				_hgt_sensor_offset = 0.0f;
 			}
+		} else if (_control_status_prev.flags.gps_hgt && _gps_data_ready && !_gps_hgt_faulty) {
+			setControlGPSHeight();
+			_fuse_height = true;
+			_in_range_aid_mode = false;
+
+			// we have just switched to using gps height, calculate height sensor offset such that current
+			// measurment matches our current height estimate
+			if (_control_status_prev.flags.gps_hgt != _control_status.flags.gps_hgt) {
```
ndepal: You only get to here if `_control_status_prev.flags.gps_hgt == true` and by calling `setControlGPSHeight();` you ensure that also `_control_status.flags.gps_hgt == true`, so you would never enter this.

ChristophTobler: should be `_control_status.flags.gps_hgt` in line 740... Thanks!

ndepal: In that case, the call to `setControlGPSHeight()` isn't strictly necessary, right?

ChristophTobler: That's true!

priseborough:
Some ratcheting is unavoidable when using a simple method of setting n offset at each handover. The way to eliminate it is to run a  continuous offset estimator for each sensor type when that sensor is not the primary height reference.

priseborough:
The existing algorithm will meet the requirement for vertical takeoff and landing with transition to horizontal flight at mission altitude.

ndepal:
@priseborough I understand why it is happening, but I think this needs to be addressed.

In my use case (flying missions that are planned to go over pre-defined obstacles with multiple stops in between), it is crucial that I'm still flying with absolute altitude.
This should be possible (the information is there).
E.g. by setting `_hgt_sensor_offset = 0` [here](https://github.com/PX4/ecl/blob/92ddd886c9919bbf5d2613cc9d2d2a8d3128a80a/EKF/control.cpp#L805) this could be achieved. Of course then we'd have the problem that there are potentially jumps in the height innovations. But with a solution that smoothes out these jumps (slew rate limiting?), I think this should be the behavior.

Running a continuous offset estimator for GPS height while in range aid mode would not solve this problem, right?

Issue #290, opened 2017-06-29
Use baro for ekf alignment


ChristophTobler:
I think it would make sense if we stick to baro only for alignment (as we already do for GPS) -> https://github.com/PX4/ecl/blob/7f1cca5b28087055f8343ec069852bedd3b86cd7/EKF/ekf.cpp#L207-L259. Because on many setups the range finder outputs rubbish/measurements outside of max/min when close to ground. Therefore the ekf never aligns.

@priseborough what do you think? Is there something against that?

mhkabir:
I ran into this, haha. Took me an hour to figure out why the EKF wouldn't initialise :P

priseborough:
@ChristophTobler  Its a  valid option, the baro is the most reliable  sensor on startup. The only other way to handle it I can think of would be to change the range finder driver to return the minimum value if not flying and the reading is outside the valid range.

LorenzMeier:
The driver change would mask the missing reading, which would potentially hide real HW issues.

ChristophTobler:
@LorenzMeier I agree. And the driver should work independent of the flight status.
So I will change it to use baro if there aren't any objections.

PR #292, opened 2017-06-30
use baro for the ekf alignment if range finder is primary height source

ChristophTobler:
As discussed in https://github.com/PX4/ecl/issues/290, I changed it to use baro if range finder is primary height.

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -204,34 +204,9 @@ bool Ekf::initialiseFilter()
 	}
 
 	// accumulate enough height measurements to be confident in the qulaity of the data
-	if (_primary_hgt_source == VDIST_SENSOR_RANGE) {
-		if (_range_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_range_sample_delayed)) {
-			if ((_hgt_counter == 0) && (_range_sample_delayed.time_us != 0)) {
-				// initialise the counter height fusion method when we start getting data from the buffer
-				_control_status.flags.baro_hgt = false;
-				_control_status.flags.gps_hgt = false;
-				_control_status.flags.rng_hgt = true;
-				_control_status.flags.ev_hgt = false;
-				_hgt_counter = 1;
-
-			} else if ((_hgt_counter != 0) && (_range_sample_delayed.time_us != 0)) {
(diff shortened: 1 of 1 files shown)


priseborough:
Do you have a test log?

ChristophTobler:
@priseborough
Here's a log where alignment failed (old):
http://logs.px4.io/plot_app?log=56189966-f135-4be3-b187-4c2a4dcb1464
And here's one with this PR:
http://logs.px4.io/plot_app?log=7271d90a-642f-4e95-aceb-244bfb69733c

priseborough:
I have verified that it is able to switch to range finder when it is available.

jlecoeur:
@ChristophTobler @priseborough @mhkabir I think this introduced another bug in height estimation.
Was this tested for flight at fixed altitude above ground with baro+sonar and no GPS? 
In my scenario I give an altitude command of -1m in offboard mode. 
After this PR, I have a large offset in the actual drone altitude (it either descends to the ground, or flies too high at approx 2m)
By checking the commit before this PR, the 1m height is correctly tracked (but of course #290 is back!)

ChristophTobler:
@jlecoeur Thanks for reporting. Can you share the two logs?

jlecoeur:
Not at the moment. I can make new tests tomorrow.

ChristophTobler:
I think I know what the problem is. Before, https://github.com/PX4/ecl/blob/ec2b54fde7a590178483ab9623459109003a4a44/EKF/control.cpp#L759-L765 did not run because `_control_status_prev.flags.rng_hgt` was already true because of https://github.com/PX4/ecl/pull/292/files#diff-0871d0235139c63531138e2793b76de7L213. Hence, the `_hgt_sensor_offset` was zero.
However, with the new changes, https://github.com/PX4/ecl/blob/ec2b54fde7a590178483ab9623459109003a4a44/EKF/control.cpp#L759-L765 runs after the initialization with baro. Therefore, `_hgt_sensor_offset` is non-zero. 

This is not a problem with a good rangefinder as it will output the minimum distance when close to ground (or at least negligible). I suspect that your sonar will output a lot of noise when on ground and therefore `_hgt_sensor_offset` is initialized with a wrong value.

PR #294, opened 2017-07-07
EKF: Enable planes to recover from bad mag data at start of flight

priseborough:
Adjusts yaw by the amount of the error between GPS and EKF course if innovations are large.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -412,6 +412,7 @@ union filter_control_status_u {
 		uint32_t ev_hgt      : 1; // 14 - true when height data from external vision measurements is being fused
 		uint32_t fuse_beta   : 1; // 15 - true when synthetic sideslip measurements are being fused
 		uint32_t update_mag_states_only   : 1; // 16 - true when only the magnetometer states are updated by the magnetometer
+		uint32_t fixed_wing  : 1; // 17 - true when the vehicle is operating as a fixed wing vehicle
 	} flags;
 	uint32_t value;
 };
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1050,14 +1050,12 @@ void Ekf::controlMagFusion()
 				if (!_control_status.flags.mag_3D) {
 					if (!_flt_mag_align_complete) {
 						// If we are flying a vehicle that flies forward, eg plane, then we can use the GPS course to check and correct the heading
-						// if we are doing wind estimation and using the zero sideslip assumotion, then it is reasonable to assume the plane is flying forward
-						bool zeroSideslipValid = _control_status.flags.fuse_beta;
-						if (zeroSideslipValid) {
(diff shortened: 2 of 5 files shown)


priseborough:
Replay testing performed using https://github.com/priseborough/Firmware/tree/pr-ekfFixedWingYawReset

priseborough:
Can someone please provide a fixed wing log where logging starts at boot (SDLOG_MODE=1)

priseborough:
There are some problems showing up on replay due to interaction with other reset logic that will require some further work.

priseborough:
Issues have been resolved and testing using replay of a fixed wing log has shown the changes are able to detect and recover from a 90 degree compass yaw error.

Testing was performed by replaying a log with the compass data rotated in yaw by +90 degrees in ekf2_main.cpp by sign and axis swaps, eg:

_mag_data_sum[0] += sensors.magnetometer_ga[1];
_mag_data_sum[1] += -sensors.magnetometer_ga[0];

Here are the velocity and magnetometer innovations with good compass data and no yaw reset using GPS course required:

**GPS Velocity Innovations - Good Data**
vel innov - good data

**Magnetometer Innovations - Good Data**
mag innov - good data

Here they are with the magnetometer data rotated in yaw by +90 deg which resulted in a single yaw correction event at the start of flight 

screen shot 2017-07-07 at 9 11 27 pm

**GPS Velocity Innovations - 90 deg compass yaw**
vel innov - bad compass

**Magnetometer Innovations - 90 deg compass yaw**
mag innov - bad compass

Replaying the log using yawed compass data without the PR changes resulted in repeated GPS resets that lasted for 4 minutes before the EKF was able to recover the yaw angle:

screen shot 2017-07-07 at 9 19 10 pm

priseborough:
I have completed further replay testing using a FW log with long straight line mission legs http://review.px4.io/plot_app?log=ed369adf-4e46-486a-bc18-14a2ed188bd0 which gives a bad magnetometer time to pull the heading round between turns. During turns the heading is corrected using the observed GPs velocity change, however during straight line segments it is only observable using the magnetometer.

This testing has shown that although this PR successfully detected and corrected magnetometer errors on launch, if flight is continues with a bad magnetometer, there is a possibility that the heading will be degraded during straight line segments sufficiently to cause a GPS rejection and reset event at the next turn. 

The following plot shows the GPS velocity innovations when the log was replayed with a 30 degree compass misalignment. The initial yaw error was corrected, however the compass error was not so great that the EKF completely ignored. Angular misalignment is the worst type of magnetometer error  because the EKF does not have the ability to learn it. 

screen shot 2017-07-08 at 10 11 51 am

Ideally the system should be alerted that a magnetometer fault has been detected and the EKF should stop using the magnetometer. As a minimum the operator should be alerted so the mission can be aborted and the plane recovered.

LorenzMeier:
I think this is good to go. Let's get this in and the reporting on top.

PR #295, opened 2017-07-07
rng aid: use hysteresis with lower hagl check

ChristophTobler:
This lowers the offset error when switching to range finder in range aid mode.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -902,7 +902,15 @@ bool Ekf::rangeAidConditionsMet(bool in_range_aid_mode)
 			use_range_finder = false;
 		}
 
-		use_range_finder &= ((_hagl_innov * _hagl_innov / (sq(_params.range_aid_innov_gate) * _hagl_innov_var)) < 1.0f);
+		// use hysteresis to check for hagl
+		if (in_range_aid_mode) {
+			use_range_finder &= ((_hagl_innov * _hagl_innov / (sq(_params.range_aid_innov_gate) * _hagl_innov_var)) < 1.0f);
+
+		} else {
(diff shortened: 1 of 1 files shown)


ChristophTobler:
> How did you come up with 0.1 and 0.01? Could that be a bit too restrictive? Should range_aid_innov_gate be a parameter exposed to the user and the hysteresis just uses e.g. 0.1*range_aid_innov_gate?

I printed the values (see the attached output of [this comment](https://github.com/PX4/Firmware/pull/7506#issuecomment-313635026)). I guess a parameter could make sense, yes. But maybe not  `range_aid_innov_gate` because if you increase the value it lowers the hagl check term (because of the division). Or we would have to use some kind of `1/range_aid_innov_gate` term for  a comparison.

ChristophTobler:
@ndepal FYI

ndepal:
> But maybe not `range_aid_innov_gate` because if you increase the value it lowers the hagl check term (because of the division).

If you increase the gate, the condition is more easily satisfied, right? Its the same as if it were on the right hand side:
```
use_range_finder &= ((_hagl_innov * _hagl_innov / (_hagl_innov_var)) < sq(_params.range_aid_innov_gate))
```
That's what I'd expect from a "gate" parameter.

ChristophTobler:
Yes , you are right! So for the Hysteresis maybe x1 and x0.01? Or how should we decide?

ChristophTobler:
rebased and tested with
baro: http://logs.px4.io/plot_app?log=2944266a-9395-4d59-940b-b66762232079
gps: http://logs.px4.io/plot_app?log=fb1d8bdd-b884-4548-9cf2-4dd405ca2196

PR #293, opened 2017-07-04
EKF: Allow fallback to non-aiding mode if external vision fusion is lost

priseborough:
Addresses https://github.com/PX4/Firmware/issues/7491#issuecomment-312613868

PR #297, opened 2017-07-12
Fixed vision position fusion bug

lachyd:
Without this fix, the system constantly re-initialises the vision position and state estimator when new vision data becomes available.


TobiMiller:
is this related to this PR?
https://github.com/PX4/ecl/pull/293

lachyd:
I don't think so. This fix is not in PR293.

PR #298, opened 2017-07-12
init rng offset with ground clearance param when on ground to avoid n…

ChristophTobler:
…oisy init

Should solve https://github.com/PX4/ecl/pull/292#issuecomment-314708514

@jlecoeur can you please retest with this one?

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -773,10 +773,16 @@ void Ekf::controlHeightFusion()
 		// we have just switched to using range finder, calculate height sensor offset such that current
 		// measurment matches our current height estimate
 		if (_control_status_prev.flags.rng_hgt != _control_status.flags.rng_hgt) {
-			if (_terrain_initialised) {
+			// use the parameter rng_gnd_clearance if on ground to avoid a noisy offset initialization (e.g. sonar)
+			if (_control_status.flags.in_air && _terrain_initialised) {
+
 				_hgt_sensor_offset = _terrain_vpos;
-			} else {
+			} else if (_control_status.flags.in_air) {
+
 				_hgt_sensor_offset = _R_rng_to_earth_2_2 * _range_sample_delayed.rng + _state.pos(2);
+			} else {
+
+				_hgt_sensor_offset = _params.rng_gnd_clearance;
 			}
 		}
 	} else if ((_params.vdist_sensor_type == VDIST_SENSOR_RANGE) && _baro_data_ready && !_baro_hgt_faulty) {


EKF/control.cpp
```diff
@@ -757,10 +757,19 @@ void Ekf::controlHeightFusion()
...
-				_hgt_sensor_offset = _terrain_vpos;
+			// use the parameter rng_gnd_clearance if on ground to avoid a noisy offset initialization (e.g. sonar)
+			if (_control_status.flags.in_air) {
+
+				if (_terrain_initialised) {
+
+					_hgt_sensor_offset = _terrain_vpos;
+				} else {
+
+					_hgt_sensor_offset = _R_rng_to_earth_2_2 * _range_sample_delayed.rng + _state.pos(2);
```
jlecoeur: A per https://dev.px4.io/en/advanced/parameter_reference.html#EKF2_HGT_MODE, I do not think we should update the hgt offset when range sensor is the primary sensor (though it makes perfect sense when baro or gps is primary).
Shouldn't it read like this?
```cpp
		// we have just switched to using range finder, calculate height sensor offset such that current
		// measurment matches our current height estimate
		if (_control_status_prev.flags.rng_hgt != _control_status.flags.rng_hgt) {
			// use the parameter rng_gnd_clearance if on ground to avoid a noisy offset initialization (e.g. sonar)
			if (_control_status.flags.in_air && _terrain_initialised) {
				hgt_sensor_offset = _terrain_vpos;

			} else {

				_hgt_sensor_offset = _params.rng_gnd_clearance;
			}
		}
```

ChristophTobler: There is one case where it would not behave correctly, I think. If you start with baro at a certain altitude and then go e.g. uphill and then switch to range sensor and the terrain is not valid (`_terrain_initialized` should/will be replaced by `get_terrain_vert_pos()`, see https://github.com/PX4/ecl/pull/301), it will use `_hgt_sensor_offset = _params.rng_gnd_clearance;` and this has the effect of a sudden height change of the drone which is undesired.

jlecoeur: To me this is not a problem as it is said in the doc that this setting should be used only over flat terrain and that NED origin may move up and down with terrain. So the Z position command becomes a height-above-ground command.

In your example, I agree that a sudden change of altitude is not desired, but 
1. changing the primary height source in flight sounds like a bad idea anyway
2. if not with sudden altitude change, the drone should converge to the commanded height. In your example the drone will keep an altitude error. Let's say the drone is flying at 2m altitude using baro. Then it moves at constant altitude above a 1.5 meter hill, now baro altitude is still 2m, but the sonar returns a height of 0.5m. By switching to range sensor and updating `_hgt_sensor_offset`, the drone will stay 0.5m above ground, which is not the commanded height of 2m.

When sonar becomes primary height sensor, I would think the *baro* offset needs to be updated, and not the sonar offset.

ChristophTobler: >1. changing the primary height source in flight sounds like a bad idea anyway

Yes, but we have to account for that.

>2. if not with sudden altitude change, the drone should converge to the commanded height. In your example the drone will keep an altitude error. Let's say the drone is flying at 2m altitude using baro. Then it moves at constant altitude above a 1.5 meter hill, now baro altitude is still 2m, but the range sensor returns a height of 0.5m. By switching to range sensor and updating _hgt_sensor_offset, the drone will stay 0.5m above ground, which is not the commanded height of 2m.

Yes, it stays at 0.5m. But that's what I would expect. I think it should use the altitude at the time the user switches.
@priseborough what do you think?

jlecoeur: That is also true. I guess it depends whether the filter is supposed to estimate height or altitude when range sensor is primary.

ChristophTobler: >Yes, it stays at 0.5m. But that's what I would expect. I think it should keep the altitude from the time the user switches.

What I meant was it should keep the height. With range finder as primary height source it cannot estimate absolute altitude

ChristophTobler:
I rebased and changed it to an `else if` instead of to `if`s . Logic is still the same.

EKF/control.cpp
jlecoeur: Ok, so just to summarize the discussion:
- we agree that the drone should not jump after switching primary height source
- I find it incorrect that with the current code, the estimate is not the height (there may be an offset added to the range measurement), nor the altitude (baro is not primary), but something in between like height-relative-to-when-param-changed.

ChristophTobler: If you switch to range finder, the estimate will be the current height. It just needs the offset to be able to fuse it into the estimator (the difference between estimated z and sensor distance). So no matter what happened before, at the point you switch to the range finder, it will stay at the same height above the ground.

ChristophTobler: In the end
```
hgt_sensor_offset = _terrain_vpos;
```
is just the filtered value of
```
_hgt_sensor_offset = _R_rng_to_earth_2_2 * _range_sample_delayed.rng + _state.pos(2);
```
And we have to use the second term it at that particular point in time the terrain estimate is not valid.

jlecoeur: > If you switch to range finder, the estimate will be the current height.

I beg to differ, but please correct me if I am wrong. This line https://github.com/PX4/ecl/blob/8ed61ac0520686041e3ee193e693639d97b663db/EKF/vel_pos_fusion.cpp#L162 says `z_innovation = z_estimated - (-range + terrain_vpos)`
In other words z_estimated is corrected towards `(-range + terrain_vpos)`, which is not the height, but the height + estimated terrain altitude when range sensor was activated.

ChristophTobler: Yes, that is correct. My formulation was probably not very good. I just meant it behaves the same way as if you would have started from that location where you switched. 
And in most cases your suggestion would behave the exact same way. The only difference is that your suggestion would react with a sudden increase in height if the terrain is at that point not valid and in air. And before you even agreed that this is an unwanted behavior. So i'm wondering what you want to change. Do you think there should be an z estimate reset (e.g. to 0.5m in the above example)?

jlecoeur: Yes exactly, I would like the estimate to converge to 0.5m in the above example. Range sensor directly measures height, so I think the offset should always be 0 (or a fixed value defined by where the sensor is mounted on the drone)  when range sensor is primary.
I understand that offsets are required to fuse everything without trouble. So yes maybe a z estimate reset + update of baro offset would work.

ChristophTobler:
@jlecoeur So you would prefer something like
```
if (_control_status_prev.flags.rng_hgt != _control_status.flags.rng_hgt) {
    resetHeight();
    _hgt_sensor_offset = _params.rng_gnd_clearance;
}
```
right?
@priseborough what do you think?

jlecoeur:
@ChristophTobler Yes, something like that.
@priseborough to summarize, I am worried that `_state.pos(2)` does not estimate height when range is primary, as suggested in the doc.

ChristophTobler:
I'm going to merge this one as it is a fix that better gets in sooner than later. I'm going to open an issue/enhancement for your suggestion @jlecoeur

priseborough:
@ChristophTobler Due to the critical nature of the estimator can we please ensure that test evidence is supplied and a code review completed before merging functional changes. I apologise that I was unable to get to this last week due to overseas travel, however there should have been either link to a log from the test of this branch or a log enabling the reviewer to verify the correct behaviour.

ChristophTobler:
@priseborough You are absolutely right! I tested the two PRs but did not link the logs to this one... https://github.com/PX4/ecl/pull/295#issuecomment-315765659

priseborough:
I would like to get @CarlOlsson added to the list of reviewers, do you know how we can do this?

ChristophTobler:
I guess he has to have merge/admin rights to this repo

LorenzMeier:
He has now.

EKF/control.cpp
```diff
@@ -773,10 +773,16 @@ void Ekf::controlHeightFusion()
 		// we have just switched to using range finder, calculate height sensor offset such that current
 		// measurment matches our current height estimate
 		if (_control_status_prev.flags.rng_hgt != _control_status.flags.rng_hgt) {
-			if (_terrain_initialised) {
+			// use the parameter rng_gnd_clearance if on ground to avoid a noisy offset initialization (e.g. sonar)
+			if (_control_status.flags.in_air && _terrain_initialised) {
+
 				_hgt_sensor_offset = _terrain_vpos;
```
priseborough: I think we should also constrain _terrain_vpos so that it is always a minimum of _params.rng_gnd_clearance larger than _state.pos(2).

ChristophTobler: That would make sense, yes. I guess you mean in the terrain estimator?

priseborough: yes, in the terrain estimator

PR #300, opened 2017-07-13
sensor_combined stores deltas in usec instead of seconds

nicolaerosia:
needed by https://github.com/PX4/Firmware/pull/7602 

Signed-off-by: Nicolae Rosia <nicolae.rosia@gmail.com>


priseborough:
This looks OK in isolation, but will need to wait until we have a solution to backwards compatibility for replay before merging this and the corresponding Firmware change.

nicolaerosia:
I have rebased this, unfortunately the changes already went into PX4/Firmware, so after merging this we'll need to update ecl

PR #303, opened 2017-07-17
Split get_terrain_vert_pos() into valid and get_vpos

ChristophTobler:
So we don't have to use dummy variables to check if terrain is valid

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -734,7 +734,7 @@ void Ekf::controlHeightFusion()
 			// we have just switched to using range finder, calculate height sensor offset such that current
 			// measurment matches our current height estimate
 			if (_control_status_prev.flags.rng_hgt != _control_status.flags.rng_hgt) {
-				if (_terrain_initialised) {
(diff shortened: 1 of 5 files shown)


priseborough:
This PR is generating the following compile error when building with PX4/Firmware master. It should not have been merged without the corresponding compatibility changes on the PX4/Firmware side of the interface.


/Users/paul/src/Firmware/src/modules/ekf2/ekf2_main.cpp: In member function 'virtual void Ekf2::run()':
/Users/paul/src/Firmware/src/modules/ekf2/ekf2_main.cpp:936:27: error: void value not ignored as it ought to be
    lpos.dist_bottom_valid = _ekf.get_terrain_vert_pos(&terrain_vpos);

I am reverting it until the required PX4/Firmware changes have been submitted.

ChristophTobler:
https://github.com/PX4/Firmware/pull/7585 they are already here

priseborough:
Then they should be linked.

PR #304, opened 2017-07-17
Improved and fixed fusion of external vision heading

lericson:
We were unable to fuse external vision heading even though the signal was good and constant. Tests showed a couple of things were askew in the implementation, so we fixed it.

:wave:

### DIFF
--- a/EKF/mag_fusion.cpp
+++ b/EKF/mag_fusion.cpp
@@ -424,7 +424,12 @@ void Ekf::fuseHeading()
 	Vector3f mag_earth_pred;
 	float measured_hdg;
 
-	// determine if a 321 or 312 Euler sequence is best
+	if (!_control_status.flags.mag_hdg && !_control_status.flags.ev_yaw) {
+		// there is no yaw observation
+		return;
+	}
+
+	// determine if a 321 or 312 Euler sequence is numerically better
 	if (fabsf(_R_to_earth(2, 0)) < fabsf(_R_to_earth(2, 1))) {
 		// calculate observation jacobian when we are observing the first rotation in a 321 sequence
 		float t9 = q0*q3;
@@ -464,20 +469,16 @@ void Ekf::fuseHeading()
 		euler321(2) = 0.0f;
 		Dcmf R_to_earth(euler321);
 
-		// calculate the observed yaw angle
-		if (_control_status.flags.mag_hdg) {
+		if (_control_status.flags.ev_yaw) {
+			const Quatf q = _ev_sample_delayed.quat;
+			float y = 2.0f*(q(0)*(3) + q(1)*q(2));
+			float x = 1.0f - 2.0f*(q(2)*q(2) + q(3)*q(3));
+			measured_hdg = atan2f(y, x);
+		} else {
 			// rotate the magnetometer measurements into earth frame using a zero yaw angle
 			mag_earth_pred = R_to_earth * _mag_sample_delayed.mag;
 			// the angle of the projection onto the horizontal gives the yaw angle
 			measured_hdg = -atan2f(mag_earth_pred(1), mag_earth_pred(0)) + _mag_declination;
-		} else if (_control_status.flags.ev_yaw) {
-			// convert the observed quaternion to a rotation matrix
-			Dcmf R_to_earth_ev(_ev_sample_delayed.quat);	// transformation matrix from body to world frame
-			// calculate the yaw angle for a 312 sequence
-			measured_hdg = atan2f(R_to_earth_ev(1, 0) , R_to_earth_ev(0, 0));
-		} else {
-			// there is no yaw observation
-			return;
 		}
 
 	} else {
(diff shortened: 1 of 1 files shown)


lericson:
More detail is available in each commit.

mhkabir:
@lericson Nice! Thanks!

priseborough:
Can you please provide a log with logging from boot and EKF2_REC_RPL=1 so this can be verified on replay or provide a log from a test of these changes.

EKF/mag_fusion.cpp
```diff
@@ -623,14 +623,14 @@ void Ekf::fuseHeading()
 		}
 	}
 
-	// wrap the heading to the interval between +-pi
-	measured_hdg = wrap_pi(measured_hdg);
```
priseborough: Does this mean there is a bug in the wrap_pi function? If so then the function should be fixed.

lericson: Iirc the previous implementation did ((y mod pi) - x) mod pi, which isn't the same as what I did here. I don't think it's broken, it's just not what is needed here.

priseborough:
@lericson Do you have a test log?

lericson:
Before or after the change?

priseborough:
After the change will do. Because the estimator is flight critical we do require testing that exercises the area of code that has been modified when there have been functional changes.

lericson:
I'll get right on it

- Ludvig

priseborough:
What definition of yaw have you assumed for the EV interface and how does the new calculation handle either pitch or roll at +-90 degrees?

lericson:
It is here defined as the body's rotation about the world Z axis, meaning roll and pitch don't affect the heading if applied separately.

lericson:
http://review.px4.io/plot_app?log=de4d80a1-73f6-49bc-ae65-bbc974232b3d

Side note, we're having difficulties getting the filter to accept the vision position updates now. Not sure if this was always the case, but it seems like it rejects anything that is not close to the origin as being too unlikely.

I would have liked to show you the filter performance in flight instead, but because of the aforementioned and the fact that our onboard computer dies on take-off (hexacopter drawing too much current on take-off, it seems.)

lericson:
Only the first ten or so seconds are interesting in the log I uploaded. Cheers

priseborough:
Your answer about the definition of yaw used for the conversion from quaternions to yaw angle is incomplete. What does the yaw angle do at the roll and pitch extremities (eg +-90 deg roll or +- 90 deg pitch) ? This needs to be consistent with the respective calculation of the observation Jacobians (H matrix).

Yaw is usually defined as the first rotation in either a 321 or 312 Euler angle sequence.

lericson:
I don’t know how else to answer than that it is the rotation about the Z axis of the navigation frame. Rotating ±90° about the body’s X or Y axis would not affect this heading given that the body is parallel to the navigation frame in the XY plane. Only if the two were applied in succession (i.e. roll 90° then pitch 90°) would the heading change (in this case by 90°.)

It is the heading as commonly referred to in Tait-Bryan angles. The code is more or less the same as in https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Quaternion_to_Euler_Angles_Conversion <https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Quaternion_to_Euler_Angles_Conversion>

priseborough:
OK, that reference uses a 321 sequence which is singular at +- 90 degrees pitch. It is only compatible with the Jacobians in the 321 method section. 

A different equation needs to be used in the 312 section, or external vision yaw forced to always use 321.

I will look at the 312 derivation again.

priseborough:
I derived the conversions between 312 euler angles and  rotation matrix  and  found an error in the 312 section in the conversion from _R_to_earth to euler angles that explains what you were seeing. I will submit a PR today to fix it.

priseborough:
I've submitted a PR that cleans up the calculation of EV yaw angle measurement using a traceable derivation. https://github.com/PX4/ecl/pull/309

EKF/mag_fusion.cpp
priseborough: I have modelled both the legacy wrap_pi  https://github.com/PX4/Matrix/blob/cf924956d7d62ce18bfc4f8441e9177ddb69c0dc/matrix/helper_functions.hpp#L21-L38) and the algorithm from this PR  in Matlab here: 

[test.txt](https://github.com/PX4/ecl/files/1164500/test.txt)

Both are producing the same output so I do not see the point of this change.

lericson: Interesting, I thought it would be modulo pi given the function name.

lericson: I've looked at it as well and it does indeed seem to be the same. Strange. That would mean the only thing preventing convergence of the Kalman filter was the heading computation.

priseborough:
I am closing this because the change to wrap_pi does not change behaviour and the change to the 312 rotation sequence section is not correct.

PR #306, opened 2017-07-19
Revert "Revert "Split get_terrain_vert_pos() into valid and get_vpos""

priseborough:
Reverts PX4/ecl#305


priseborough:
#303 Is now linked to the required compatibility changes https://github.com/PX4/Firmware/pull/7585

PR #307, opened 2017-07-19
constrain _terrain_vpos to be a minimum of _params.rng_gnd_clearance …

ChristophTobler:
…larger than _state.pos(2)

@priseborough Or should it be rejected?

### DIFF
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -102,6 +102,11 @@ void Ekf::runTerrainEstimator()
 			_cos_tilt_rng = cosf(_params.rng_sens_pitch);
 
 		}
+
+		//constrain _terrain_vpos to be a minimum of _params.rng_gnd_clearance larger than _state.pos(2)
+		if (_terrain_vpos - _state.pos(2) < _params.rng_gnd_clearance) {
+			_terrain_vpos = _params.rng_gnd_clearance + _state.pos(2);
+		}
 	}
 }
 


priseborough:
I'm not sure what you mean by rejected. This is a constraint on the state prediction model, not fusion test.

priseborough:
Can we get this tested?

ChristophTobler:
I can go out and test this PR.

I'm not sure I quite understand. Are these two scenarios? One is having wrong readings (LidarLite example) and the other one having no readings whilst being in flight?

priseborough:
The problem can occur after flying out of range for  period and coming back in range. Very occasionally this causes the range finder to start returning 'stuck' readings that vary only slightly around a small value (eg 3m) for several seconds which fools the height above ground estimator because it has a large variance at that point in time and immediately accepts the data.

ChristophTobler:
pr_vpos_constraint
log: http://logs.px4.io/plot_app?log=ba458ed8-ec9e-4724-8e47-5bf8b3264379
I think `dist_bottom` should never go below `_params.rng_gnd_clearance` (see end of log). And it would probably make sense to output `_params.rng_gnd_clearance`, if on ground.

ChristophTobler:
>The problem can occur after flying out of range for period and coming back in range. Very occasionally this causes the range finder to start returning 'stuck' readings that vary only slightly around a small value (eg 3m) for several seconds which fools the height above ground estimator because it has a large variance at that point in time and immediately accepts the data.

So something like a hysteresis where the range data has to "prove" that it is not stuck if the estimator was not running before?

EKF/terrain_estimator.cpp
```diff
@@ -139,6 +139,11 @@ void Ekf::fuseHagl()
 			_time_last_hagl_fuse = _time_last_imu;
 			_innov_check_fail_status.flags.reject_hagl = false;
 
+			//constrain _terrain_vpos to be a minimum of _params.rng_gnd_clearance larger than _state.pos(2)
```
priseborough: Looking at your test data, this check should be moved to runTerrainEstimator  at the end of the else section after line 104.

priseborough: That way the constraint condition will be applied regardless of whether range data passes checks.

ChristophTobler: You are right! Tested in SITL. Will do a quick flight test later today.

ChristophTobler:
vpos_constraint
log: http://logs.px4.io/plot_app?log=3a879911-c9b6-4bda-9841-6030bbca5a64

Still drops below 0 (at around 30.7s) because `pos[2]` in ekf2_main.cpp and `_state.pos(2)` in ecl is not exactly the same. Is that an issue?

priseborough:
Not an issue for the EKF. The EKF state prediction runs on a delayed time horizon and winds the state forward to current tim using an output predictor, so when you plot states from the estimator_status message they can be up to 200msec old compared to what is in the local_position message.

If it is  a problem for the control loops, then an output constraint could be added in ekf2_main.cpp to the distance to ground .

PR #308, opened 2017-07-19
EKF: Update documentation and make compatible with Doxygen

priseborough:


priseborough:
@dagar Can you confirm (or otherwise) that I have used the correct syntax for Doxygen to be able to use the comments. TXS.

dagar:
Yes that looks fine, but I didn't know we were bothering with doxygen anymore. I'm sure we've let it rot on the PX4 side, and the generation probably needs all the generated headers (uorb, etc) in place first.

Standalone doxygen for ecl would also make sense and could be generated and uploaded by travis-ci.

PR #309, opened 2017-07-21
EKF: Improve efficiency of yaw fusion for External Vision 

priseborough:
The yaw angle fusion can either use the measured magnetic field projected onto the horizontal or quaternions from an external vision system. When using external vision quaternions, the calculations for projection of the magnetic field were still being performed.

This PR moves the mag field projection and yaw calculation into the if (_control_status.flags.mag_hdg) branch. It also replaces the conversion from the External Vision quaternions with a more efficient direct calculation with a referenced derivation.

A log from testing with a wide range of ground rotations including those performed during a mag calibration sequence can be found here: 

http://logs.px4.io/plot_app?log=a174c2b2-af05-404e-a242-f32843c3b06b

### DIFF
--- a/EKF/mag_fusion.cpp
+++ b/EKF/mag_fusion.cpp
@@ -460,21 +460,25 @@ void Ekf::fuseHeading()
 		Eulerf euler321(_state.quat_nominal);
 		predicted_hdg = euler321(2); // we will need the predicted heading to calculate the innovation
 
-		// Set the yaw angle to zero and rotate the measurements into earth frame using the zero yaw angle
-		euler321(2) = 0.0f;
-		Dcmf R_to_earth(euler321);
-
 		// calculate the observed yaw angle
 		if (_control_status.flags.mag_hdg) {
+			// Set the yaw angle to zero and rotate the measurements into earth frame using the zero yaw angle
+			euler321(2) = 0.0f;
+			Dcmf R_to_earth(euler321);
+
 			// rotate the magnetometer measurements into earth frame using a zero yaw angle
 			mag_earth_pred = R_to_earth * _mag_sample_delayed.mag;
+
 			// the angle of the projection onto the horizontal gives the yaw angle
 			measured_hdg = -atan2f(mag_earth_pred(1), mag_earth_pred(0)) + _mag_declination;
+
 		} else if (_control_status.flags.ev_yaw) {
-			// convert the observed quaternion to a rotation matrix
-			Dcmf R_to_earth_ev(_ev_sample_delayed.quat);	// transformation matrix from body to world frame
-			// calculate the yaw angle for a 312 sequence
-			measured_hdg = atan2f(R_to_earth_ev(1, 0) , R_to_earth_ev(0, 0));
+			// calculate the yaw angle for a 321 sequence
+			// Expressions obtained from yaw_input_321.c produced by https://github.com/PX4/ecl/blob/master/matlab/scripts/Inertial%20Nav%20EKF/quat2yaw321.m
+			float Tbn_1_0 = 2.0f*(_ev_sample_delayed.quat(0)*_ev_sample_delayed.quat(3)+_ev_sample_delayed.quat(1)*_ev_sample_delayed.quat(2));
+			float Tbn_0_0 = sq(_ev_sample_delayed.quat(0))+sq(_ev_sample_delayed.quat(1))-sq(_ev_sample_delayed.quat(2))-sq(_ev_sample_delayed.quat(3));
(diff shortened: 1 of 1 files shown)


mhkabir:
@lericson Can you please test this?

lericson:
We're trying to get our newly-arrived PixRacer to run so I can test this, strangely they fail to appear as USB serial devices.

lericson:
These changes don't work for us. Specifically the 312-based computation doesn't agree with the 321-based computation for zero roll and zero pitch situations. I can't yet explain why that would be as this doesn't happen in a Python prototype I made of the code.

priseborough:
That doesn't make sense. You are sending across quaternion which describes the orientation of the vehicle and is agnostic to the choice of Euler sequence the EKF makes internally.

priseborough:
I will run some test code to check that the result is consistent if the EKF's own quaternion is used.

lericson:
There are two ways of computing the heading given a quaternion, yes? Selection of which one of these two to use is, somewhat curiously, contingent on a property of the current estimated orientation.  I'm sure this is not an issue in practice as the estimated orientation is typically close to the external vision's signal.

So, these two computations should typically agree given a non-degenerate quaternion. However, when a degenerate quaternion is presented, specifically one that is upside-down (180° roll), the estimated yaw is flipped.

I have such an example of this at work, but it's 23:45 over here and sadly I didn't get around to pressing submit on the comment I wrote.

Also I would be curious to know when this singularity is supposed to appear, as I've as of yet been unable to see it in action.

priseborough:
The singularity for a 321 sequence appears at +- 90 degrees pitch and results in rapid switching of 180 degrees in the yaw angle as the vehicle pitches through the singularity. For that reason we use a 312 sequence at that orientation for the heading fusion which works well with vehicles such as tail-sitters.

With respect to your comment about a degenerate quaternion - are you referring to when the signs on the quaternion are swapped? In such a case there is no effect on the yaw angle because the calculations of the rotation matrix terms Tbn_1_0, Tbn_0_0, Tbn_0_1_neg and Tbn_1_1 use the product of quaternions.

lericson:
Right, I understand the point now. Sorry about that. I also constructed this program to prove your point,

```python
import numpy as np

def qdot(q0, q1):
    w0, x0, y0, z0 = q0
    w1, x1, y1, z1 = q1
    return np.array([- x1*x0 - y1*y0 - z1*z0 + w1*w0,
                       x1*w0 + y1*z0 - z1*y0 + w1*x0,
                     - x1*z0 + y1*w0 + z1*x0 + w1*y0,
                       x1*y0 - y1*x0 + z1*w0 + w1*z0], dtype=np.float64)

quat = lambda alpha, axis: np.r_[np.cos(alpha/2.0), np.sin(alpha/2.0)*axis]
one_degree_pitch = quat(np.pi/180.0, np.r_[0.0, 1.0, 0.0])
q = quat(0.0, np.r_[1.0, 0.0, 0.0])

for i in range(360):
    qw, qx, qy, qz = q
    y = 2.0*(qw*qz + qx*qy)
    x = 1.0 - 2.0*(qy*qy + qz*qz)
    yaw = np.arctan2(y, x)
    print(i, yaw)
    q = qdot(q, one_degree_pitch)
```

In actual fact I think the `one_degree_pitch` above would be one degree roll.

lericson:
So this I guess is inevitable then, that the heading sometimes will flip when switching between the two modes of computation?

priseborough:
If by heading you mean the yaw used to calculate the innovation, then yes that will jump, however what is important is that the EKF is internally consistent so the innovations do not jump. 

The EKF attitude as defined by the quaternion states will not jump.

lericson:
I see, so basically the predicted heading would also flip. Did I get that right?

- Ludvig

priseborough:
There should not be be a sign flip due to the way the optimum sequence is selected. The way to visualise it is that if the angle between the X-axis and the horizontal is larger than that for the Y axis, it will use 312. Visualise the 312 yaw as being the yaw angle of the Y axis projection onto the horizontal minus  90 deg. Visualise the 321 yaw as being the yaw angle of the projection of the X axis intot he horizontal.

There will be step change in yaw angle in both the prediction and measurement when it switches between sequences.

priseborough:
I've tested this replaying the following log: http://logs.px4.io/plot_app?log=e5c8c8c0-d060-44c5-8d01-60447ec4e597 which has external vision position and quaternion data generated using the updated SITL log. Despite multiple switching between  312 and 321, heading innovations are continuous.

screen shot 2017-07-29 at 2 44 43 pm

priseborough:
I will add a reboot required attribute to the aid mask. It really does need to be set prior to alignment for other reasons as well.

PR #311, opened 2017-07-26
EKF: Add ability to use EV and GPS data together

priseborough:
Fuse external vision data using a relative position odometry method when GPS data is also being used and enable both GPOS and EV data to be fused on the same time step.

The previous filter logic turned of EV data fusion when GPS data was present to avoid conflict between the two data sources. This PR switches to using a delta position observation model for EV position data when GPs data is also being fused.

Initial response to https://github.com/PX4/Firmware/issues/7408

A replay log containing EV and GPs data is required to test this. This should be obtained with SDLOG_MODE = 2 and EKF2_REC_RPL = 1

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -115,7 +115,6 @@ void Ekf::controlFusionModes()
 
 	// control use of observations for aiding
 	controlMagFusion();
-	controlExternalVisionFusion();
 	controlOpticalFlowFusion();
 	controlGpsFusion();
 	controlAirDataFusion();
@@ -123,9 +122,12 @@ void Ekf::controlFusionModes()
 	controlDragFusion();
 	controlHeightFusion();
 
-	// for efficiency, fusion of direct state observations for position and velocity is performed sequentially
-	// in a single function using sensor data from multiple sources (GPS, external vision, baro, range finder, etc)
+	// For efficiency, fusion of direct state observations for position and velocity is performed sequentially
+	// in a single function using sensor data from multiple sources (GPS, baro, range finder, etc)
 	controlVelPosFusion();
+
+	// Additional data from an external vision sensor can also be fused.
+	controlExternalVisionFusion();
 
 	// report dead reckoning if we are no longer fusing measurements that constrain velocity drift
 	_is_dead_reckoning = (_time_last_imu - _time_last_pos_fuse > _params.no_aid_timeout_max)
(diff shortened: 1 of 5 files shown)


priseborough:
@mhkabir I am thinking of extending this PR to include the changes required to handle the  misalignment of the vision system navigation frame relative to the EKF's NED frame. The algorithm  will:

1) Use the product of the EKF quaternion and EV quaternion inverse to calculate the quaternion describing the rotation from EKF to EV Navigation frames.  This will be done after data is extracred from the observation buffer, hopefully with EV and EKF states that are time aligned.
2) Convert the quaternion to a rotation vector.
3) Filter the rotation vector components using a first order LPF with specified time constant. This will also include spike protection. The filter will look after jitter noise caused by timing errors and the fact that EV observations will not coincide with the EKF prediction cycle.
4) Convert the filtered rotation vector to a rotation matrix which is then used to rotate the EV NED position observations from the EV nav to the EKF nav frame.

This is the most direct approach I can think of.

Additionally It would also help the EKF to receive two timestamps from the EV system which indicate the last time the EV system reset the position or quaternion states. This will help the EKF to handle loop closure events, etc that cause large steps in state estimates.

Let me know what you think.

mhkabir:
@priseborough Approach seems perfect.

> Additionally It would also help the EKF to receive two timestamps from the EV system which indicate the last time the EV system reset the position or quaternion states. This will help the EKF to handle loop closure events, etc that cause large steps in state estimates.

Regarding this, I've been meaning to add this to the API for some time. I will send a PR with this to Mavlink.

priseborough:
I've added the alignment algorithm and pushed it to show the method. I now need to add the parameters required to control it. We can either have the vision system specify if it uses an NED or arbitrary reference frame or have a parameter in the EKF to specify.

Some more code also required with to adjust the learned angular offset when there is a state reset from either the EKF or EV.

ChristophTobler:
@priseborough You can now test with vision position estimates in gazebo if that's of any help
https://github.com/PX4/Firmware/pull/7686

ChristophTobler:
Didn't seem to work in simulation (not sure if I got all parameters right)
http://logs.px4.io/plot_app?log=60ea870b-b71f-47f6-b65d-8ebd152c1631

ChristophTobler:
Was an error in my tranformation... https://github.com/PX4/sitl_gazebo/pull/115 
Here's a new log: http://logs.px4.io/plot_app?log=e5c8c8c0-d060-44c5-8d01-60447ec4e597

priseborough:
@ChristophTobler thanks for that log. i have been able to use it to test the changes on replay. For the test I added an NED offset of (5,10,15) metres to the EV position data. The replay log is here: http://logs.px4.io/plot_app?log=46af31be-959a-4964-b28d-efe0ae10dedd

The EKF starts tracking the EV system in absolute position mode until GPS checks pass. It then resets the position to the GPS and continues using both EV and GPS position, but with the EV data fused as an incremental position measurement.

screen shot 2017-07-29 at 8 37 08 am

priseborough:
I will submit the handling of reference frame angular offsets in a separate PR.

priseborough:
@ChristophTobler or @mhkabir If you are both happy with this to go in as is or have any required amendments, let me know.

mhkabir:
@priseborough Some comments : 

> The EKF starts tracking the EV system in absolute position mode until GPS checks pass. It then resets the position to the GPS and continues using both EV and GPS position, but with the EV data fused as an incremental position measurement.

I'm concerned about the reset. Will the position reset to GPS cause a jump in estimation? I'm just thinking of the behaviour in 3 broad scenarios we'll need to handle : 

1. Start flying outdoors (GPS aiding, no vision), then transition to indoors.
2. Start flying indoors (no GPS, vision aiding), then transition to outdoors.
3. Mixed environment where both are active and aiding.

1 and 3 should be fine, looking at the code, but I'm not sure about 2.

> I will submit the handling of reference frame angular offsets in a separate PR.

Sounds fine to me. We'll need support for the angular offsets (in particular yaw) anyway before this can be tested fully. Do you see a compelling reason to put it into a new PR? :)

priseborough:
The compelling reason for a new PR is that the angular alignment changes are more complex, may take some time to work through and this PR already solves https://github.com/PX4/Firmware/issues/7408.

Yes the states will jump, however the position reset information is already published on the local position message to be handled by the control loops.

priseborough:
The local position resets are already handled in mc_pos_control_main.cpp.

The global position message is not be published until the EKF has set an origin and started using GPS, so loops using that message are unaffected.

mhkabir:
> The compelling reason for a new PR is that the angular alignment changes are more complex, may take some time to work through and this PR already solves PX4/Firmware#7408.

Ok, makes complete sense. I will await your alignment PR.

> Yes the states will jump, however the position reset information is already published on the local position message to be handled by the control loops.

This is fine from the controls side, but if something is using the state estimate offboard (e.g a 3D mapping system), this is not a good thing to happen.
I don't see why we need to reset to GPS position when it becomes available. What I did to solve this GPS-denied to GPS-enabled transition scenario was to initialise the local-global origin at the point where GPS aiding comes online. Then you don't need to reset states and can just switch to incremental visual odometry fusion in the next iteration.

priseborough:
OK, I see the point - it will make integration with off-board consumers much easier if local position is continuous where possible. I've added a patch that achieves this. The following plot shows the local position being maintained through the reset.

screen shot 2017-07-29 at 3 31 53 pm

mhkabir:
Thank you, that's awesome!

PR #312, opened 2017-07-29
EKF: Enable operation with arbitrary External Vision reference frame

priseborough:
This PR adds a parameter activated function that uses the navigation to body frame quaternion from the external vision system to estimate the rotation from the EKF to the EV reference frame. This is then used to rotate EV position measurements into the EKF NED navigation frame before they are used.

It has been created in response to requirements identified in https://github.com/PX4/Firmware/issues/7408.

This function is enabled by setting bit position 6 in EKF2_AID_MASK to true. If bit position 4 is also set to true (EKF is using the EV system's yaw angle) then the compensation is disabled because the EV system's navigation frame becomes the reference.

**TESTING**

Testing has been performed using SITL with the following branch to enable the EV data to be rotated into a different frame of reference. 

https://github.com/priseborough/Firmware/tree/pr-ekfExtVisQuat

See the following log produced running make posix_sitl_default gazebo_iris.

https://review.px4.io/plot_app?log=b17ec2dd-f150-4b47-9534-e9ff96106561

Position innovations were small and the EV system alignment was correctly estimated as 1 radian of yaw. 
screen shot 2017-09-26 at 4 27 42 pm

The test was successfully repeated with EKF2_AID_MASK = 73 to enable GPS fusion:

https://review.px4.io/plot_app?log=ef6fac87-b038-49fe-bc9c-fe07a783473c

Position innovations were small and the EV system alignment was correctly estimated as 1 radian of yaw. 
screen shot 2017-09-26 at 5 19 37 pm

The position hold with GPS fusion (EKF2_AID_MASK = 73) was more as precise than with using EV data (EKF2_AID_MASK = 72). This may be due to the loss of velocity observations and/or the use of differential position to fuse EV data when it is not in a NED frame.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -167,12 +167,6 @@ void Ekf::controlExternalVisionFusion()
 				} else {
 					resetPosition();
 					resetVelocity();
-					// we cannot use an absolue position from a rotating reference frame
-					if (_params.fusion_mode & MASK_ROTATE_EV) {
-						_fuse_hpos_as_odom = true;
-					} else {
-						_fuse_hpos_as_odom = false;
-					}
 
 				}
 			}
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -133,6 +133,7 @@ void Ekf::controlFusionModes()
 
 	// report dead reckoning if we are no longer fusing measurements that directly constrain velocity drift
 	_is_dead_reckoning = (_time_last_imu - _time_last_pos_fuse > _params.no_aid_timeout_max)
+			&& (_time_last_imu - _time_last_delpos_fuse > _params.no_aid_timeout_max)
 			&& (_time_last_imu - _time_last_vel_fuse > _params.no_aid_timeout_max)
 			&& (_time_last_imu - _time_last_of_fuse > _params.no_aid_timeout_max);
 
@@ -142,6 +143,13 @@ void Ekf::controlExternalVisionFusion()
 {
 	// Check for new exernal vision data
 	if (_ev_data_ready) {
+
+		// if the ev data is not in a NED reference frame, then the transformation between EV and EKF navigation frames
+		// needs to be calculated and the observations rotated into the EKF frame of reference
+		if ((_params.fusion_mode & MASK_ROTATE_EV) && (_params.fusion_mode & MASK_USE_EVPOS) && !(_params.fusion_mode & MASK_USE_EVYAW)) {
+			// rotate EV measurements into the EKF Navigation frame
+			calcExtVisRotMat();
+		}
 
 		// external vision position aiding selection logic
 		if ((_params.fusion_mode & MASK_USE_EVPOS) && !_control_status.flags.ev_pos && _control_status.flags.tilt_align && _control_status.flags.yaw_align) {
@@ -154,12 +162,17 @@ void Ekf::controlExternalVisionFusion()
 				// reset the position if we are not already aiding using GPS, else use a relative position
 				// method for fusing the position data
 				if (_control_status.flags.gps) {
-					_hpos_odometry = true;
+					_fuse_hpos_as_odom = true;
 
 				} else {
 					resetPosition();
 					resetVelocity();
-					_hpos_odometry = false;
+					// we cannot use an absolue position from a rotating reference frame
+					if (_params.fusion_mode & MASK_ROTATE_EV) {
+						_fuse_hpos_as_odom = true;
+					} else {
+						_fuse_hpos_as_odom = false;
+					}
 
 				}
 			}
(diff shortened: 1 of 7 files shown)


priseborough:
Forcing the EKF to use incremental position fusion solved the problem with the innovation divergence, but at the expense of additional drift.

screen shot 2017-07-29 at 5 53 05 pm

mhkabir:
Looks good to me, code-wise.

> Investigate options enabling use of absolute position fusion when the EV frame is not NED.

Do you have any new ideas on this? I have been thinking about it as well, and I don't see a good way to do it without introducing drift of delta fusion (as opposed to absolute).

ChristophTobler:
Sorry, I was on vacation... Not sure I quite get this:

>Investigate options enabling use of absolute position fusion when the EV frame is not NED.

Isn't this what this PR is about? ->
>This is a placeholder for development of a capability to operate the EKF with external navigation systems that use a different frame of reference to NED.

@priseborough Not sure how you tested this because the referenced log uses NED.

priseborough:
I hacked a rotation of the EV data in ekf2_main

priseborough:
I've rebased and updated the quaternion order. Testing to follow.

priseborough:
@ChristophTobler Is there a way I can reproduce you simulation test? The old log will no longer replay due to the number of uORB message changes..

ChristophTobler:
@priseborough Yes. Just use the normal simulation `make posix_sitl_default gazebo_iris` and change the parameter `EKF2_AID_MASK` to use external vision.

priseborough:
I'm not happy with the drift performance when using the delta position observation model. Further tuning is required.

priseborough:
It looks like the overshoot and drift are caused by the observation variance being too high for what is essentially a relative measurement and the default value of EKF2_EV_DELAY of 175 msec being much larger than SITL is achieving. Reducing  EKF2_EV_DELAY to 10 and using a relative position observation error of 0.01m helped significantly.

Changliu52:
Awesome upgrade! @priseborough 
How do I test this (GPS+OF+EV fusion using delta position) in real flight?
I am thinking to upload https://github.com/priseborough/Firmware/tree/pr-ekfExtVisQuat to my pixracer and stream the vision pose through vision_pose/pose from mavros. Is this sensible?

Kind Regards,
Chang

priseborough:
@Changliu52 Enable use of OF and EV data via the EKF2_AID_MASK parameter. In theory using mavros should work, but there is work for you to do on the firmware side. Data received needs to be time stamped when received by the firmware and any delay relative to the time stamp compensated using the EKF2_EV_DELAY parameter.  Data needs to be published to the vehicle_vision_position and vehicle_vision_attitude topics.

priseborough:
I've rebased on ecl. This required resolving a number of conflicts due to changes to prior changes to the fusion control and reset logic, so I will need to repeat the SITL testing using https://github.com/priseborough/Firmware/tree/pr-ekfExtVisQuat

priseborough:
SITL test log here: https://logs.px4.io/plot_app?log=4e5ec8ed-8691-488e-a15a-12d8c49c7115

mhkabir:
What is the status here?

priseborough:
@mhkabir  Your review status is still pending.

EKF/control.cpp
```diff
@@ -243,16 +256,57 @@ void Ekf::controlExternalVisionFusion()
 			_ev_sample_delayed.posNED(1) -= pos_offset_earth(1);
 			_ev_sample_delayed.posNED(2) -= pos_offset_earth(2);
 
-			// if GPS data is being used, then use an incremental position fusion method for EV data
-			if (_control_status.flags.gps) {
-				_hpos_odometry = true;
+			// Use an incremental position fusion method for EV data if using GPS or if the observations are not in NED
+			if (_control_status.flags.gps || (_params.fusion_mode & MASK_ROTATE_EV)) {
+				_fuse_hpos_as_odom = true;
```
mhkabir: I believe that this makes https://github.com/PX4/ecl/pull/312/files#diff-4cac31c333867233bb8a08de1e812a9aR171 redundant.

EKF/control.cpp
```diff
@@ -154,12 +162,17 @@ void Ekf::controlExternalVisionFusion()
...
 				if (_control_status.flags.gps) {
-					_hpos_odometry = true;
+					_fuse_hpos_as_odom = true;
 
 				} else {
 					resetPosition();
 					resetVelocity();
-					_hpos_odometry = false;
+					// we cannot use an absolue position from a rotating reference frame
+					if (_params.fusion_mode & MASK_ROTATE_EV) {
```
mhkabir: If we power up without GPS lock and have (non-NED) visual odometry, it should still be possible to perform a yaw alignment between the NED and vision frame and fuse the odometry as absolute position (after rotation) to avoid the drift of delta-fusion.

mhkabir:
Looks good in general. I think we can merge this now, and then iterate further on the comment about absolute vision fusion I raised. 

My idea would be to switch to absolute vision position fusion whenever we are not aiding with GPS even when the vision frame is not NED. This would also require tracking the position difference between the NED origin and the vision origin (in addition to the rotation this PR tracks.)

EKF/control.cpp
mhkabir: In addition, if we are tracking the absolute position delta between the vision and NED frame origins, then if we lose GPS while flying, and have visual odometry, we should be able to switch back to absolute VO fusion.

EKF/control.cpp
ChristophTobler: I think you are right

ChristophTobler:
@priseborough I agree with @mhkabir. If you can verify that the redundant part can be removed (and actually remove it), this can go in.:+1:

ChristophTobler:
@priseborough @mhkabir Can we not wait for the EV to NED frame (`_ev_rot_mat`) to converge and then fuse the absolute position?

priseborough:
@ChristophTobler Yes for operation without GPS, we could do that with a small modification.

mhkabir:
Let's do that in the next iteration over the vision interface. I'm putting together a 250 racer with VIO and GPS, so it should be easy to get datasets and fly around to test new code.

PR #313, opened 2017-07-31
EKF: Improve in-flight mag error detection, recovery and isolation for fixed wing

priseborough:
This PR improves the handling of in-flight yaw reset and magnetometer error handling   and further develops the functionality introduced by https://github.com/PX4/ecl/pull/294 which was able to recover from yaw errors present at launch/takeoff.

The following functions have been added:

If the EKF rejects GPS data long enough to result in an in-flight reset (which also causes loss of global position validity), then the yaw will be reset to the wind relative GPS course over ground vector.

If two in-flight mag yaw failure and reset events occur, then the mag is declared as unhealthy and magnetometer fusion is halted resulting in a warning message to the console:

"EKF stopping magnetometer use"

and bit position 18 in the filter status message being set to true.

**Testing**

Testing is ongoing with an initial assessment  performed using replay of the following flight log: http://review.px4.io/plot_app?log=ed369adf-4e46-486a-bc18-14a2ed188bd0 with the magnetometer data rotated in ekf2_main.cpp to simulate a  sensor yaw misalignment.

Testing with a crash log http://review.px4.io/plot_app?log=1c1b7c69-d55c-4f63-a287-85d5be996453 from https://github.com/PX4/Firmware/issues/7609 showed that the changes prevented the GPS rejection being repeated after first in-flight reset and would have enabled the flight to continue given an appropriate failsafe strategy.

velocity innovations just prior to crash - current master
screen shot 2017-07-31 at 4 50 25 pm

velocity innovations just prior to crash - this PR
screen shot 2017-07-31 at 4 52 05 pm

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -413,6 +413,7 @@ union filter_control_status_u {
 		uint32_t fuse_beta   : 1; ///< 15 - true when synthetic sideslip measurements are being fused
 		uint32_t update_mag_states_only   : 1; ///< 16 - true when only the magnetometer states are updated by the magnetometer
 		uint32_t fixed_wing  : 1; ///< 17 - true when the vehicle is operating as a fixed wing vehicle
+		uint32_t mag_fault   : 1; ///< 18 - true when the magnetomer has been declared faulty and is no longer being used
 	} flags;
 	uint32_t value;
 };
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -443,8 +443,13 @@ void Ekf::controlGpsFusion()
 
 			if (do_reset) {
 				// Reset states to the last GPS measurement
-				resetPosition();
-				resetVelocity();
+				if (_control_status.flags.fixed_wing) {
+					// if flying a fixed wing aircraft, do a complete reset that includes yaw, velocity and position
+					realignYawGPS();
+				} else {
+					resetVelocity();
+					resetPosition();
+				}
 				ECL_WARN("EKF GPS fusion timeout - reset to GPS");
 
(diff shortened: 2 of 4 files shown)


Antiheavy:
Here are a couple more flight logs from today where we intentionally screwed up the magnetometer.

Flight with normal mag orientation, bad mag cal using magnet-https://logs.px4.io/plot_app?log=3f6b5d68-5b03-4825-a0d8-e587dce77269
 
Flight with wrong mag orientation- https://logs.px4.io/plot_app?log=ca1f903c-8354-4b86-9ee8-910fe206fbc7

@priseborough feel free to use these as test datasets against this PR - they may or maynot be helpful.

priseborough:
@Antiheavy Thanks for that. The vehicle_land_detected message has changed since the version you are using, so I had to remove the maybe_landed filed, but after doing that, bit logs were able to be replayed and triggered the start of flight reset. There is not enough straight line flying in these logs for the bad mag to have enough time to pull the yaw away and cause a second reset.

I was able to exercise the new functionality with replay of http://review.px4.io/plot_app?log=ed369adf-4e46-486a-bc18-14a2ed188bd0 log using this test branch to misalign the mag https://github.com/priseborough/Firmware/tree/ekfReplayTest-wip.

Wiht a 0.5 radian compass misalignment, current ecl master does a yaw reset  on launch, but the misaligned compass pulls the yaw away during the straight legs, resulting in multiple GPS reset events.

before

With these changes, the  compass is rejected after the second yaw reset event and the flight continues with no further GPS rejection.

after

after - mag innovations

PR #314, opened 2017-07-31
Pr check stuck rng

ChristophTobler:
@priseborough Is this somewhat what you were thinking of (https://github.com/PX4/ecl/pull/307#issuecomment-316368891)?

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -105,6 +105,8 @@ void Ekf::controlFusionModes()
 	_range_data_ready = _range_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_range_sample_delayed)
 			&& (_R_rng_to_earth_2_2 > 0.7071f);
 
+	checkForStuckRange();
+
 	_flow_data_ready = _flow_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_flow_sample_delayed)
 			&&  (_R_to_earth(2, 2) > 0.7071f);
 	_ev_data_ready = _ext_vision_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_ev_sample_delayed);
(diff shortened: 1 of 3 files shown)


ChristophTobler:
[Here](http://logs.px4.io/plot_app?log=3465176e-fc75-4ef3-b70a-fd1fdc8b033a)'s a SITL log with your described range finder behavior.

priseborough:
@ChristophTobler  Yes, that's representative of the type of fault I have seen before. Now in theory we could get a false positive If the range finder went out of range and then the vehicle flew over a flat roof and hovered there, however that is an edge case and eventually vertical movement of the vehicle would cause the variance check to pass.

@CarlOlsson Does this look sensible to you - I think it would have prevented the terrain estimation error we were looking at the other day.

ChristophTobler:
>Now in theory we could get a false positive If the range finder went out of range and then the vehicle flew over a flat roof and hovered there, however that is an edge case and eventually vertical movement of the vehicle would cause the variance check to pass.

I thought of that as well. But not using good range data for a short while is less dangerous as using false data, I think.

priseborough:
Agree.

CarlOlsson:
I will review as soon as I get some time in the coming days!

CarlOlsson:
I have tested this is SITL now. However I could not reproduce the intended behavior. 
image
15s: The vehicle takes off
20s: The vehicle reaches 10 m altitude
25s: The range finder suddenly starts to output a fixed value of 2m. The hagl innovations goes to 8 m. The hagl_test_ratio spikes. The hagl_innov_variance starts to grow since the ekf is rejecting the range measurements.
48s: The EKF2_RNG_GATE param is increased to speed up the test until hagl_test_ratio is below one.
50s: The hagl_test_ratio goes below one and the range finder measurements are not rejected anymore. The dist_bottom_valid flag becomes 1 and dist_bottom becomes 2.

-> This should not have happened, the ekf should reject the range sensor since it is stuck on 2 m. Is this a correct interpretation?

priseborough:
@CarlOlsson There needs to have been no range finder data for more than 10 seconds before the check will operate - see line 954 in control.cpp. The intended scenario is a vehicle coming back within range from up and away flight.

ChristophTobler:
@CarlOlsson Sorry for the late response but I was on vacation.
@priseborough is right. There needs to have been no range finder data (invalid data) for more than 10 seconds. Your test scenario would be like flying over a house with a flat roof where it should not be rated as stuck.

CarlOlsson:
Ok, yes you are right that the failure case we have seen is that the sensor gets stuck if it is out of range not suddenly when in range. So makes sense, thanks @ChristophTobler for the implementation!

PR #317, opened 2017-08-11
Python based testing

PeterDuerr:
In order to facilitate adding tests for ecl (unfortunately I am not that fond of writing unit tests in C++ :-)), I wrote a (very simple) Python wrapper based on SWIG. Exploiting Python libraries (pytest, hypothesis, and optionally matplotlib), one can now easily add assert-based tests, property-based tests, as well as plot data during (selected) tests.

To test my work, please checkout my branch and run `cmake -DPythonTests=1 ../EKF && make pytest-quick`. You can also run `make pytest-plots` to see graphs like the attached...

If you agree with this general approach, one could refactor the main classes in the C++ code slightly to make the wrapper more elegant and powerful.

Before I go any further though, please let me know what you think. I would love to hear your feedback. 

Peter

test_accel_z_bias_converges

### DIFF
--- a/EKF/tests/pytest/lint
+++ b/EKF/tests/pytest/lint
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+###############################################################################
+#
+# Copyright (c) 2017 Estimation and Control Library (ECL). All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in
+#    the documentation and/or other materials provided with the
+#    distribution.
+# 3. Neither the name ECL nor the names of its contributors may be
+#    used to endorse or promote products derived from this software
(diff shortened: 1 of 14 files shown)


dagar:
Awesome! We should talk about getting this into the CI system.

EKF/tests/pytest/lint
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python3
+###############################################################################
```
dagar: No .py file extension?

dagar:
Can you list the package dependencies?

EKF/tests/pytest/lint
PeterDuerr: AFAIK, no .py file extension is quite standard for executable Python scripts (see, e.g., the `pip` command).

PeterDuerr:
@dagar In the latest commit I added a `requirements.txt` file and instead of having the hard-coded list in `CMakeLists.txt` as before, I read it at configuration time. Hope that helps...

LorenzMeier:
We have Travis enabled on this repo - would you be interested to extend the .travis file to run the tests you added?

PeterDuerr:
@LorenzMeier Hi Lorenz, sure, it should be fairly straight-forward to do so... I'll have a look on Monday.

LorenzMeier:
Awesome!

PR #320, opened 2017-08-24
The EKF can start normally when we don't have magnetometer. 

kechaofan:
We can set the parameter "EKF2_MAG_TYPE" to NONE in QGC when we don't have magnetometer or do not want to use it. And the EKF2 can still work.


priseborough:
This is not a complete solution, requires the vehicle to be aligned to north on startup and does not enable yaw bias to be learned before takeoff. Please look at https://github.com/PX4/ecl/pull/281.

PR #321, opened 2017-08-25
EKF trivial code style cleanup

dagar:
Mostly minor white space, except sideslip_fusion.cpp which was a horrible mix of tabs and spaces.


dagar:
@priseborough have you considered including the generated code directly? In cases where the generated code isn't included contiguously you could add macros to reference different blocks, or break it up into separate generated pieces.

priseborough:
Yes, but the autocode generator needs to be modified first to make the formatting and variable naming compatible. I still need to do some editor find and replace changes on it before importing.

PR #322, opened 2017-09-11
EKF Add mag fusion mode for VTOL

priseborough:
Adds a mode where if flying as a fixed wing vehicle, 3D mag fusion is only used to update the field estimates. This prevents misaligned mag or soft iron effects biasing the vehicle attitude solution and causing incorrect accel bias values that can cause undesirable velocity estimation error transients when transitioning back into RW flight if the transition involves a large attitude change.

### DIFF
--- a/EKF/mag_fusion.cpp
+++ b/EKF/mag_fusion.cpp
@@ -165,12 +165,11 @@ void Ekf::fuseMag()
 
 	/*
 	When flying as a fixed wing aircraft, a misaligned magnetometer can cause an error in pitch/roll and accel bias estimates.
-	When MAG_FUSE_TYPE_AUTOFW is selected and the vehicle is flying as a fixed wing as detected by the use of wind estimation
-	and current airspeed estimates, then magnetometer fusion is only allowed to access the magnetic field states.
+	When MAG_FUSE_TYPE_AUTOFW is selected and the vehicle is flying as a fixed wing, then magnetometer fusion is only allowed
+	to access the magnetic field states.
 	*/
 	_control_status.flags.update_mag_states_only = (_params.mag_fusion_type == MAG_FUSE_TYPE_AUTOFW)
-			&& _control_status.flags.wind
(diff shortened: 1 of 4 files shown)


EKF/mag_fusion.cpp
```diff
@@ -163,6 +163,15 @@ void Ekf::fuseMag()
 		return;
 	}
 
+	/*
+	When flying as a fixed wing aircraft, a misaligned magnetometer can cause an error in pitch/roll and accel bias estimates.
+	When MAG_FUSE_TYPE_AUTOFW is selected and the vehicle is flying as a fixed wing as detected by the use of wind estimation
+	and current airspeed estimates, then magnetometer fusion is only allowed to access the magnetic field states.
+	*/
+	_control_status.flags.update_mag_states_only = (_params.mag_fusion_type == MAG_FUSE_TYPE_AUTOFW)
+			&& _control_status.flags.wind
```
CarlOlsson: why not use the _control_status.flags.fixed_wing instead?

priseborough: Good idea, I'll push a patch.

CarlOlsson:
@priseborough what is left for this to go in?

priseborough:
Testing - could be done in SITL.

priseborough:
After rebasing and testing in SITL using make posix_sitl_default gazebo_standard_vtol and EKF2_MAG_TYPE = 3 is causing the EKF to never pass the yaw alignment check. To be investigated further.

priseborough:
The bug has been found and fixed. Changes tested using 'make posix_sitl_default gazebo_standard_vtol' with EKF2_MAG_TYPE set to 3

Log file here: https://logs.px4.io/plot_app?log=35ab8f93-7514-4f4c-b002-e66ee6c1b6ea

priseborough:
@CarlOlsson  If you can review the changes, we can get this in.

PR #323, opened 2017-09-11
EKF: use hamiltonian convention for quaternion multiplication order

priseborough:
See https://github.com/PX4/Firmware/pull/7908


MaEtUgR:
@priseborough Please merge this such that I can put the correct submodule update for ecl in https://github.com/PX4/Firmware/pull/7908
We want to have a commit hash in there that is in the ecl master history.

Issue #328, opened 2017-09-29
Missing wind_is_valid fcn


CarlOlsson:
1) Say that you launch a fixed wing aircraft with a blocked pitot. You are still able to fly based on GPS velocity, but some time into the flight there is a spike from the airspeed sensor, could be a single measurement. The EKF will initialize the wind states based on that single data point and then report the same wind states throught the entire flight. The consumer of this data will then use the stuck wind estimate since there is no wind_is_valid() function.

Resulting wind estimate from a single airspeed measurement at ~48s
![30952129-d7927490-a426-11e7-9c0e-5e422d2984c6](https://user-images.githubusercontent.com/7795133/31019653-ae2b51ca-a530-11e7-8339-3d61b9e54b58.png)

2) A fixed wing aircraft is flying and fusing data from an airspeed sensor. Suddenly the sensor stops outputting data. You would expect that the wind velocity state variance would grow during this time, to the contrary they are set to zero since _control_status.flags.wind becomes false after 10s.

Simulated airspeed sensor failure between ~60 to ~95 sec
![image](https://user-images.githubusercontent.com/7795133/31019605-89f9a298-a530-11e7-89e4-98ae8959cb55.png)

CarlOlsson:
There is a way for the consumer to check if the wind states are valid

wind_estimate message
```
float32 variance_east # Wind estimate error variance in east / Y direction (m/sec)**2 - set to zero (no uncertainty) if not estimated
```

Not so clean but it removes the need for a wind_is_valid field in the wind_estimate topic.

PR #329, opened 2017-10-02
EKF: set wind states to 0 if we are not estimating wind

CarlOlsson:
As described [here](https://github.com/PX4/ecl/issues/328) the consumer of estimated wind speed needs to check the covariance of the wind velocity states in order to know if they are nonzero and hence valid. However, this is not done e.g. [here](https://github.com/PX4/Firmware/blob/master/src/modules/ekf2/ekf2_main.cpp#L1135), or [here](https://github.com/PX4/Firmware/blob/master/src/modules/bottle_drop/bottle_drop.cpp#L514). In order to make it easier for the consumer of the estimated wind speed data I propose to set the states to zero if we know that they are invalid, this since that is the best guess anyway if you don't know the wind.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -998,6 +998,8 @@ void Ekf::controlAirDataFusion()
 
 	if (_control_status.flags.wind && airspeed_timed_out && sideslip_timed_out && !(_params.fusion_mode & MASK_USE_DRAG)) {
 		_control_status.flags.wind = false;
+		_state.wind_vel(0) = 0.0f;
+		_state.wind_vel(1) = 0.0f;
 
 	}
 


priseborough:
The negative I can see with this is that for a vehicle that transitions between RW and FW modes multiple time per flight, the wind estimate will go to zero each time and the learned wind will be lost. I would prefer we handle this where the wind message is published in ekf2_main.cpp and either do so in a way that respects the wind-estimate.msg interface specification or modify the specification.

CarlOlsson:
Sounds good. Would be nice to do a rework the estimated airspeed/estimated windspeed topics in the coming weeks, together with getting this in https://github.com/PX4/Firmware/pull/7721

PR #330, opened 2017-10-02
EKF: added get_wind_velocity_var function

CarlOlsson:


priseborough:
Code looks OK, I need to track down the reason for the CI failure.

PR #334, opened 2017-10-10
ekf: fix optical flow bugs

ChristophTobler:
Fixes https://github.com/PX4/Firmware/issues/8102

The lines
```
optflow_sample_new.flowRadXYcomp(0) = optflow_sample_new.flowRadXY(0) - optflow_sample_new.gyroXYZ(0);
optflow_sample_new.flowRadXYcomp(1) = optflow_sample_new.flowRadXY(1) - optflow_sample_new.gyroXYZ(1);
```
result in,
```
optflow_sample_new.flowRadXYcomp(0) = + flow->gyrodata(0) - (- flow->gyrodata(0))
optflow_sample_new.flowRadXYcomp(1) = + flow->gyrodata(1) - (- flow->gyrodata(1))
```
if the vehicle is on ground and flow quality is not good, which is non-zero.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -300,6 +300,7 @@ void Ekf::controlOpticalFlowFusion()
 				// set the flag and reset the fusion timeout
 				_control_status.flags.opt_flow = true;
 				_time_last_of_fuse = _time_last_imu;
+				ECL_INFO("EKF Starting Optical Flow Use");
 
 				// if we are not using GPS then the velocity and position states and covariances need to be set
 				if (!_control_status.flags.gps) {
@@ -365,20 +366,6 @@ void Ekf::controlOpticalFlowFusion()
 
 		}
 
-		// handle the case when we are relying on optical flow fusion and lose it
-		if (_control_status.flags.opt_flow && !_control_status.flags.gps && !_control_status.flags.ev_pos) {
-			// We are relying on flow aiding to constrain attitude drift so after 5s without aiding we need to do something
-			if ((_time_last_imu - _time_last_of_fuse > 5e6)) {
-				// Switch to the non-aiding mode, zero the velocity states
-				// and set the synthetic position to the current estimate
-				_control_status.flags.opt_flow = false;
-				_last_known_posNE(0) = _state.pos(0);
-				_last_known_posNE(1) = _state.pos(1);
-				_state.vel.setZero();
-
-			}
-		}
-
 		// Accumulate autopilot gyro data across the same time interval as the flow sensor
 		_imu_del_ang_of += _imu_sample_delayed.delta_ang - _state.gyro_bias;
 		_delta_time_of += _imu_sample_delayed.delta_ang_dt;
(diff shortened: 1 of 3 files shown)


ChristophTobler:
I added a check to see if the terrain is valid. Otherwise it shouldn't use flow.

priseborough:
I now know why the flow is being rejected on the ground. With very bad quality,  the integration interval is going to zero which is stopping flow measurements being pushed to the buffer.

EKF/control.cpp
```diff
@@ -383,8 +383,8 @@ void Ekf::controlOpticalFlowFusion()
 		_imu_del_ang_of += _imu_sample_delayed.delta_ang - _state.gyro_bias;
 		_delta_time_of += _imu_sample_delayed.delta_ang_dt;
 
-		// fuse the data
-		if (_control_status.flags.opt_flow) {
+		// fuse the data if the terrain/distance to bottom is valid
+		if (_control_status.flags.opt_flow && get_terrain_valid()) {
```
priseborough: This is going to cause problems on ground when the range finder is out of range

ChristophTobler: We could do something like
```
// use rng_gnd_clearance if on ground
				if (!_ekf.get_in_air_status()) {
					range_finder.current_distance = _params->rng_gnd_clearance;

				// check if distance sensor is within working boundarie
				} else if (range_finder.min_distance >= range_finder.current_distance ||
					         range_finder.max_distance <= range_finder.current_distance) {
					range_finder_updated = false;
				}
```
in the Firmware. But then handheld tests wouldn't work.

priseborough: I was thinking that if on ground, an absence of range data for longer than TBD or range data below min or above some reasonable max value for hand carriage would cause rng_gnd_clearance to be used.

ChristophTobler:
I fixed a small bug which made the flow predictions go unstable. I also added the PX4Flow behavior to the simulation https://github.com/PX4/sitl_gazebo/pull/132 (that also revealed the bug)

EKF/control.cpp
ChristophTobler: or actually this way you can still do handheld tests
```
// check if distance sensor is within working boundaries
if (range_finder.min_distance >= range_finder.current_distance ||
	range_finder.max_distance <= range_finder.current_distance) {
	// use rng_gnd_clearance if on ground
	if (_ekf.get_in_air_status()) {
		range_finder_updated = false;

	} else {
		range_finder.current_distance = _params->rng_gnd_clearance;
	}
}
```

ChristophTobler:
Works well on my hardware
px4_flow_ekf_issue_fixed

I guess we can get this in as CI is broken?

priseborough:
I've fixed up the reversion to constant position mode which makes it more robust to handling  when on ground.

priseborough:
SITL testing is showing large innovation spikes on attitude changes, but this is also happening on master.

screen shot 2017-10-12 at 9 07 22 am

ChristophTobler:
Cool, thanks!

PR #339, opened 2017-10-18
Ekf pos ctrl limits wip

ChristophTobler:
@priseborough I changed the fusion flags handling because it wasn't going back from GPS valid which resulted in no velocity restriction.
I also changed that it only resets the states if it was relying on that sensor only.


EKF/control.cpp
```diff
@@ -384,16 +384,17 @@ void Ekf::controlOpticalFlowFusion()
...
-		if ((_time_last_imu - _time_last_of_fuse > 5e6)) {
+		if (!_control_status.flags.gps && !_control_status.flags.ev_pos) {
 			// Switch to the non-aiding mode, zero the velocity states
 			// and set the synthetic position to the current estimate
-			_control_status.flags.opt_flow = false;
 			_last_known_posNE(0) = _state.pos(0);
 			_last_known_posNE(1) = _state.pos(1);
 			_state.vel.setZero();
-			ECL_WARN("EKF Stopping Optical Flow Use");
+			ECL_WARN("EKF Resetting states");
```
priseborough: This should be "EKF Stopping Navigation"

EKF/control.cpp
```diff
@@ -492,16 +493,19 @@ void Ekf::controlGpsFusion()
...
 			// then we need to switch to the non-aiding mode, zero the velocity states
 			// and set the synthetic GPS position to the current estimate
 			_control_status.flags.gps = false;
-			_last_known_posNE(0) = _state.pos(0);
-			_last_known_posNE(1) = _state.pos(1);
-			_state.vel.setZero();
-			ECL_WARN("EKF measurement timeout - stopping navigation");
+			ECL_INFO("EKF Stopping GPS Use");
 
+			if (!_control_status.flags.opt_flow && !_control_status.flags.ev_pos) {
```
priseborough: The concern I have with this is that it removes the ability of a FW aircraft using airspeed fusion to continue to dead-reckon and provide an attitude estimate.

priseborough:
I've modified the fallback from GPS only so that the EKF can continue to provide an attitude reference using air data to constrain drift. This is required for FW.

priseborough:
I'm still working on the FW SITL testing for this

priseborough:
FW SITL test with GPS loss: https://logs.px4.io/plot_app?log=c66b5f18-f4a7-4f7c-9c97-b2ffdcab025a

Quad test using optical flow only: https://logs.px4.io/plot_app?log=51792727-5967-4e16-8b96-969c5ae3e7e3

priseborough:
Quad test with combined GPS and optical flow: https://logs.px4.io/plot_app?log=666670fe-05eb-42bf-b8f6-86e5a6e23de1

ChristophTobler:
@priseborough Now we have the same issue again that I tried to fix which is the GPS flag staying true if flow is active. This leads to having no limits when just using flow https://github.com/PX4/ecl/blob/ekfPosCtrlLimits-wip/EKF/ekf_helper.cpp#L1014.
I commit a new patch which should fix it and still enable dead reckoning with airspeed for FW. Does that make sense?

ChristophTobler:
Ah there is still a case where it doesn't work... Now it will never enter the first if statement. I'll rethink this one

priseborough:
Maybe I can add another control flag for air data aiding. That way the GPS flag can go false.

ChristophTobler:
Looks good to me:+1: The only issue now is that if you don't fuse anything the output will be flooded with https://github.com/PX4/ecl/blob/ekfPosCtrlLimits-wip/EKF/control.cpp#L512

priseborough:
OK, I'll fix that first thing tomorrow.

priseborough:
I fixed the repeating message, and whilst making the mods and testing found some other areas that needed to be fixed for the FW GPS loss handling to work reliably. Here are the velocity estimates and truth values from a SITL FW test with GPS lost at 120 sec and regained at 250 sec.

screen shot 2017-10-20 at 11 02 46 am

Log here: https://logs.px4.io/plot_app?log=0d97e3e0-2725-4409-b6e3-812653d4324b

priseborough:
I've added a couple of patches which make the local and global position validity reporting consistent. This also means the vehicles position estimate will show on the GCS when it is still navigating and in theory, if failsafes were modified, FW vehicles could commence an RTL. In SITL testing, the vehicle could be flown in altitude mode using the GCS map and AHRS display.

Log here: https://logs.px4.io/plot_app?log=87101682-69c6-4684-9fb9-d1c811a28086

ChristophTobler:
@priseborough I did a test flight on our f450 and the velocity limit mostly worked well until it somehow lost local position and then all of a sudden there was no limit any more and with the right stick fully to the left the quad started to tilt. That was a bit scary. Not yet sure why it lost the local position.

log : https://logs.px4.io/plot_app?log=06406bf0-9a50-40a8-a9e5-ddd5a9952357

The height limit didn't seem to work but then I found out that I was not low enough (seems like the range sensor has an offset).

priseborough:
Turning GPS use on an off in flight using the EKF2_AID_MASK parameter is not a valid method. The state machine has not been tested for this so the parameter has a 'reboot_required true' attribute. For SITL testing I used the 'gpssim set -t ' console command to disable the simulated GPS. We need an equivalent for flight testing.

priseborough:
When GPS use is gained the speed limit suddenly turns of which is going to cause problems. I will modify the code so that the speed limit is raised gradually until it is higher than the current speed, after which it will be turned off.

priseborough:
The loss of local position failsafe at the end of the flight was due to the flow sensor going to zero quality for over a second. We can extend that timeout period if required, but given the vehicle was a metre up at the time I don't know what would have caused this.

priseborough:
The problem with sudden release of the ground speed limit when GPS usee commences has been addressed with addition of this commit to the Firmware PR #8157 https://github.com/PX4/Firmware/pull/8157/commits/2096faf19830aca8500b3efc3f88e1bf53d1b100

PR #342, opened 2017-10-21
Create total energy control system implementation

LorenzMeier:
This is a new, clean and streamlined variant of the mathematical derivation created by Paul Riseborough a few years ago.


dagar:
Should we get rid of the vertical velocity complementary filter?

LorenzMeier:
The filter is not just filtering the input, but also damping the control response. So we would need extensive testing to figure out if the raw estimation output is suitable. My guess is that it's not because in the presence of vibration the state estimate needs to track the vibration state closely, but the controller should not.

dagar:
I was referring to this part that's skipped if we pass in a valid vz. https://github.com/PX4/ecl/blob/tecs_cleanup/tecs/tecs.cpp#L99-L114

Currently fw_pos_control_l1 only runs with global position updates (vz always valid), so when used within PX4 this particular complementary filter isn't used.

dagar:
Rebased to fix CI.

PR #343, opened 2017-10-22
additional tecs cleanup

dagar:


dagar:
https://github.com/PX4/ecl/pull/342

tecs/tecs.cpp
```diff
@@ -326,80 +298,81 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const math::Ma
 	// Calculate the throttle demand
 	if (_underspeed_detected) {
 		// always use full throttle to recover from an underspeed condition
-		_throttle_setpoint = 1.0f;
+		_last_throttle_setpoint = 1.0f;
```
LorenzMeier: This doesn't look correct as you're skipping the constrain and rate limit that was originally present below.

dagar: That only happens in the non underspeed case.

dagar:
No problem, we can bring the main tecs_cleanup in first and come back to this later. I found TECS hard to grok with all intermediate state kept in the class. Moving towards something a bit more functional and detangling update_vehicle_state_estimates() from update_pitch_throttle() (updates some state again) would also help in my opinion.

I'd also like to get a standalone ecl build going and setup such the library is built the same standalone or when used within PX4. Each component can build both a static and shared lib. When consumed with cmake you can pick the pieces you want. We'd need to bring geo into ECL. The mathlib dependency can easily be removed from TECS and L1 can be converted to Matrix.

CarlOlsson:
Would be nice if this could go in at some point, I can do some flight testing after new year if needed

dagar:
I can get it rebased and do some additional testing in the meantime.

dagar:
There shouldn't be anything too radical happening here, it's mainly trying to get a handle on passing data as arguments, or accessing it from the object. Many helper methods are actually pretty close to being pure functions.

CarlOlsson:
awesome, after that we can start updating the tecs_status message etc

LorenzMeier:
@bresch Would you mind reviewing this as well?

dagar:
To be honest I've forgotten the details here.

My general goals with TECS architecture (only partially addressed here)
  - detangle the vehicle state estimate updates from setting airspeed and altitude setpoints (fw_pos_ctrl is doing unnecessary work)
  - cleanup the confusing mixture of internal state, argument passing, and keeping a lot of unnecessary intermediate calculations

PR #344, opened 2017-10-22
fix travis-ci

dagar:


dagar:
I've restored the gcc and clang builds and have the python test closer to working again, but marked it as allowed to fail (for now).

PR #347, opened 2017-10-25
EKF: Improve handling of magnetometer errors

priseborough:
This PR contains the following improvements:

1) The use of 3-axis fusion is enforced for the first 30 seconds after the first in-flight magnetic yaw alignment. This reduces the possibility of switching back to magnetic yaw fusion when the mag field state estimates have not stabilised. Doing so can result in  large yaw errors and loss of navigation. 

2) The time for the no movement test to pass before switching back from 3-axis to magnetic yaw fusion has been increased from 2 to 5 seconds to allow the field states to relearn aster a sudden yaw rotation.

3) If magnetometer innovations are failing along with velocity or position, then velocity and position estimates are flagged as invalid because this combination of sensor inconsistency is nearly always caused by an invalid yaw angle that makes inertial navigation impossible until the yaw error is corrected.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1078,6 +1078,7 @@ void Ekf::controlMagFusion()
 	if (!_control_status.flags.in_air) {
 		_last_on_ground_posD = _state.pos(2);
 		_flt_mag_align_complete = false;
+		_time_last_flt_align = 0;
 		_num_bad_flight_yaw_events = 0;
 	}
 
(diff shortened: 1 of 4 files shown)


priseborough:
Closed pending further testing

priseborough:
SITL testing has not indicated that commit 30d95bb is helpful. Further analysis is required.

nicolaerosia:
@priseborough can you please share your testing procedure?

priseborough:
Takeoff in position mode whilst yawing and with large biases on the magnetometer. I was unable to reproduce the fault seen in the log I was analysing. However this testing has revealed that the EKF reporting and commander bad position failsafe logic are not well matched and fly-aways in auto missions are still possible unless the operator can manually select land mode because the bad nav condition momentarily clears when the EKF resets it's states which puts the vehicle back into it's previous mode and results in large rapid speed and buildup before the failsafe operates again.

I was using the following branch to rotate the mag: https://github.com/priseborough/Firmware/tree/testBadMag-wip 
along with changes to the SITL param file so it used default EKF2 parameters

priseborough:
Here is the behaviour seen in the log that I am trying to reproduce in SITL (sorry, the log is private so I am unable to share the link). The copter takes off at 43 seconds and switches to 3-axis mag fusion at 50.5 seconds. There was some magnetic anomaly present at ground level so there was a 0.25 radian yaw reset at the switch event. From 51 to 54 seconds the learned mag bias states rapidly change to a incorrect value as the copter rotates in yaw to adjust. The resulting yaw error results in a high speed fly-away later in the log.

screen shot 2017-10-26 at 4 43 05 pm

It looks like a bad covariance matrix problem, and may be related to the pre-flight message:

0:33:473 | WARNING | [lib__ecl] EKF invalid accel bias - resetting covariance

Unfortunately the log system did not start until the copter arms, so other than the buffered pre-flight message and the fact that the vertical innovations are high at arming, there are no other anomalies.

32035286-d45c208e-ba63-11e7-8f12-02b3ebc130de

I will now try to reproduce the ground level magnetic anomaly and induce a 0.25 rad yaw reset to see if there is something about the setting of the quaternion covariances by the reset function that is causing a problem.

PR #348, opened 2017-10-26
EKF: Add method to enable the IMU bias states to be reset externally

priseborough:
This will be required as part of the fixes to IMU sensor fallback logic raised here https://github.com/PX4/Firmware/issues/8186

Test results will be added when the supporting code on the other side of the interface has been completed.

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -91,6 +91,9 @@ void Ekf::initialiseCovariance()
 	_prev_dvel_bias_var(1) = P[14][14] = P[13][13];
 	_prev_dvel_bias_var(2) = P[15][15] = P[13][13];
 
+	// record IMU bias state covariance reset time - used to prevent resets being performed too often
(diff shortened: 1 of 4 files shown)


TSC21:
@priseborough are we able to adapt this to a usage of a Mavlink command that restarts the estimator in general? Would be very useful in various situations. The current one I think of occurs on SITL, when one wants to drag and drop the model with the sim running.

priseborough:
@TSC21 That is somthing for a separate PR and could be done by:

a) Addition of a global filter reset method that resets all class variables back to startup values. The danger with adding such a method is that when future developers add variables, if they are not initialised in the reset function, the results will be unpredictable and may escape testing.

b) Addition in ekf2_main of a method to stop and re-start the ekf2 program. This is my preferred approach.

This would also be useful after sensor calibration as an alternative to re-powering the autopilot.

priseborough:
Tested using https://github.com/PX4/Firmware/tree/pr-fixImuBiasReset

screen shot 2017-10-30 at 3 50 56 pm

Issue #217, opened 2016-12-09
Improve tolerance to negative pressure altitude offsets caused by ground effect 


priseborough:
Some models of multi-rotors experience a significant negative pressure altitude offset when operating in ground effect. This can cause height overshoot followed by sink immediately after take-off and can cause very low sink rates, pop-ups, land detector failures and tip-overs on landing.

Logic is required to deal with these errors.

priseborough:
Addressed by https://github.com/PX4/ecl/pull/351

priseborough:
Fixed

PR #351, opened 2017-11-04
EKF: Protect against ground effect induced static pressure rise

priseborough:
Addresses #217 

Can be used to reduced the effect of ground effect induced static pressure rise during takeoff and landing. When rotary wing vehicles take off, the rotor wash often induces an increase in static pressure which causes the baro height to read low with a transient that is typically up to 5m for commonly used disc loadings and fuselage and rotor relative positions heights. The symptons observed by the pilot are typically an overshoot of the initial height demand, followed by a slow descent as the estimator recovers.

- This PR adds a method which applies a dead-band to the vertical position innovation if using baro for height and if in the ground effect region during and just after takeoff.
- The method needs to be activated externally.
- The method turns off automatically after 10 seconds or if a specified height is gained.

Requires changes on the firmware side to use. The following code was used to test it on multiple takeoffs: https://github.com/priseborough/Firmware/tree/ekfGndEffectComp-wip

Indoor test log: https://logs.px4.io/plot_app?log=9dfecbf4-6d21-4077-8508-76f54fc280d9

screen shot 2017-11-04 at 6 52 09 pm

### DIFF
--- a/EKF/vel_pos_fusion.cpp
+++ b/EKF/vel_pos_fusion.cpp
@@ -102,6 +102,20 @@ void Ekf::fuseVelPosHeight()
 			// innovation gate size
 			gate_size[5] = fmaxf(_params.baro_innov_gate, 1.0f);
 
+			// Compensate for positive static pressure transients (negative vertical position innovations)
+			// casued by rotor wash ground interaction by applying a temporary deadzone to baro innovations.
+			float deadzone_start = 0.25f * _params.baro_noise;
+			float deadzone_end = deadzone_start + _params.gnd_effect_deadzone;
+			if (_control_status.flags.gnd_effect) {
+				if (_vel_pos_innov[5] < -deadzone_start) {
+					if (_vel_pos_innov[5] <= -deadzone_end) {
+						_vel_pos_innov[5] += deadzone_end;
+					} else {
+						_vel_pos_innov[5] = -deadzone_start;
+					}
+				}
+			}
+
 		} else if (_control_status.flags.gps_hgt) {
 			fuse_map[5] = true;
 			// vertical position innovation - gps measurement has opposite sign to earth z axis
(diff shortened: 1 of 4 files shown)


EKF/vel_pos_fusion.cpp
```diff
@@ -102,6 +102,20 @@ void Ekf::fuseVelPosHeight()
...
 			gate_size[5] = fmaxf(_params.baro_innov_gate, 1.0f);
 
+			// Compensate for positive static pressure transients (negative vertical position innovations)
+			// casued by rotor wash ground interaction by applying a temporary deadzone to baro innovations.
+			float deadzone_start = 0.25f * _params.baro_noise;
+			float deadzone_end = deadzone_start + _params.gnd_effect_deadzone;
+			if (_control_status.flags.gnd_effect) {
+				if (_vel_pos_innov[5] < -deadzone_start) {
+					if (_vel_pos_innov[5] <= -deadzone_end) {
+						_vel_pos_innov[5] += deadzone_end;
```
dagar: This is more of a question, but can you explain the logic here? Why is it shifted by deadzone_end rather than just constrained within the dead zone?

dagar:
Why activate externally and not within the EKF anytime it transitions to in air (and not fixed wing)?

priseborough:
Because the effect starts when motors start to spin, not when it is flying, and the EKF doesn't know about that. Also I want it to be able to be activated externally during landing if necessary.

EKF/vel_pos_fusion.cpp
priseborough: I don't understand the question.

dagar: When _vel_pos_innov[5] (z pos innovation) is below the deadzone end you increase it by deadzone_end. Why bump the innovation by deadzone_end instead of just limiting it to deadzone_end?

priseborough: Because it prevents a possible runaway of the inertial height in that direction and still allows for temporary baro height 'spikes' up to the value set by the parameter.

PR #353, opened 2017-11-08
EKF: use int cast for XeY to avoid float casting of variables

ChristophTobler:
The default type for XeY is float

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -130,7 +130,7 @@ public:
 			int index = (_head - i);
 			index = index < 0 ? _size + index : index;
 
-			if (timestamp >= _buffer[index].time_us && timestamp - _buffer[index].time_us < 100000) {
+			if (timestamp >= _buffer[index].time_us && timestamp - _buffer[index].time_us < (uint64_t)1e5) {
 
 				// TODO Re-evaluate the static cast and usage patterns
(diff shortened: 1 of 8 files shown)


EKF/RingBuffer.h
```diff
@@ -130,7 +130,7 @@ class RingBuffer
 			int index = (_head - i);
 			index = index < 0 ? _size + index : index;
 
-			if (timestamp >= _buffer[index].time_us && timestamp - _buffer[index].time_us < 100000) {
+			if (timestamp >= _buffer[index].time_us && timestamp - _buffer[index].time_us < (int)1e5) {
```
bkueng: The timestamps I see are `uint64_t`, so we should cast to that.
And we need to be careful not to lose accuracy, but it's fine in these cases here.

ChristophTobler:
@bkueng changed it to `uint64_t`

EKF/RingBuffer.h
MaEtUgR: **I agree** because it's theoretically correct although it doesn't change the compile output in the online tool. I assume the compiler chooses the "bigger" of the two data types to do the comparison.

(And **if it would not**, there needs to be a >35 minutes time difference to break it. Would be a nasty bug to find.)

MaEtUgR:
Looks good.
@bkueng How about we introduce a define for this generally in PX4? Something like `#define US(a) (long int)(a * 1e6)`
It works nice even with fractions of seconds: https://godbolt.org/g/z8F8Se

Issue #352, opened 2017-11-08
Yaw error growth after in-flight alignment causes loss of global position


priseborough:
Fault detected in the following FW log: https://review.px4.io/plot_app?log=298fa694-516a-4d76-9d18-a5137adbc0d7

Flight code was flown was forked from https://github.com/PX4/Firmware/commit/5a17f6b2d576e19cbc22f463b08d572667de7b08

Vehicle was launched and develops a 1 rad yaw error after the inflight yaw alignment.

![screen shot 2017-11-08 at 2 03 09 pm](https://user-images.githubusercontent.com/3596952/32529614-94485de4-c48d-11e7-825f-707237731817.png)

The learned mag fields have a large step change when the in-flight yaw reset is performed and the mag field strength is fairly constant:

![screen shot 2017-11-08 at 2 08 24 pm](https://user-images.githubusercontent.com/3596952/32529747-55f13574-c48e-11e7-8e63-675927cba456.png)

The log is good quality and the loss of navigation after takeoff is reproduced during replay. 

![screen shot 2017-11-08 at 2 10 38 pm](https://user-images.githubusercontent.com/3596952/32529807-a7270180-c48e-11e7-9c68-90e69a4f6584.png)

Reducing the magnetic field state variance from the parameter derived value of sq(0.05) to sq(0.001) prevents the buildup in yaw error and subsequent rejection of GPS observations.

![screen shot 2017-11-08 at 2 11 53 pm](https://user-images.githubusercontent.com/3596952/32529836-cd8035fe-c48e-11e7-9e11-bccb084b8d15.png)

Reducing the assumed heading uncertainty after reset from the default value of 0.3 rad to 0.05 rad also eliminated the loss of navigation.

![screen shot 2017-11-08 at 2 17 23 pm](https://user-images.githubusercontent.com/3596952/32529964-8fc22456-c48f-11e7-955e-75c190e61b93.png)

The initial state variances appear to be set to too high a value during the reset.

priseborough:
After reviewing the code and testing hypothesis using replay of the log, it is likely that this was caused by a combination of transient sensor errors and the simultaneous reset the magnetic field and quaternion states and their corresponding covariances. I have created a branch to test changes here: https://github.com/PX4/ecl/tree/ekfMagYawReset-wip which implements the following fixes that should reduce the likelihood of this condition occurring for both RW and FW vehicle types:

[...]

PR #354, opened 2017-11-09
Prevent loss of navigation accuracy after in-flight mag yaw alignment

priseborough:
Fixes https://github.com/PX4/ecl/issues/352

See issue for explanation and preliminary test results using replay of FW logs.

The RW testing is still outstanding and will be completed tomorrow.

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -56,8 +56,12 @@ bool Ekf::resetVelocity()
 		// this reset is only called if we have new gps data at the fusion time horizon
 		_state.vel = _gps_sample_delayed.vel;
 
+		// use GPS accuracy to reset variances
+		setDiag(P,4,6,sq(_gps_sample_delayed.sacc));
+
 	} else if (_control_status.flags.opt_flow || _control_status.flags.ev_pos) {
 		_state.vel.setZero();
+		zeroOffDiag(P,4,6);
 
 	} else {
 		return false;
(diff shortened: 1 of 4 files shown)


dagar:
The CI failures are real.

priseborough:
OK, I have explicitly declared a dynamically sized array which has fixed it.

EKF/ekf_helper.cpp
```diff
@@ -1215,9 +1215,17 @@ void Ekf::zeroCols(float (&cov_mat)[_k_num_states][_k_num_states], uint8_t first
...
 {
-	// store diagonals
+	if (first > last) {
+		return;
+	}
+
+	// create dynamically sized array to save diagonal elemnets
+	const uint8_t array_size = last-first+1;
+	float * variances;
+	variances = new float [array_size];
```
dagar: You'd need to delete variances when you're done with it. How about stack allocating `float variances[_k_num_states]` instead? I think a bit of wasted stack is better than a dynamic allocation.

priseborough:
SITL testing of RW flight with a 0.1Ga bias on the X magnetometer axis showed that the practice of using heading fusion to constrain yaw drift while the filed estimates were stabilising made the recovery from mag bias errors slower because the mag heading fusion was 'fighting' the change in yaw from the GPS observations and vehicle motion. This has been addressed by https://github.com/PX4/ecl/pull/354/commits/3f35ad9a76941b238b1d6e44004a7413efa98682

priseborough:
RW testing has been performed with replay of https://logs.px4.io/plot_app?log=78676267-73d6-4588-a4a1-1cfa1dd30af0

Nominal configuration

screen shot 2017-11-10 at 11 26 50 am

Addition of 0.15Ga bias to mag Y axis

screen shot 2017-11-10 at 11 32 07 am

screen shot 2017-11-10 at 11 32 24 am

priseborough:
Tested on a RW vehicle here: https://logs.px4.io/plot_app?log=2b25ce95-28ea-4636-b7b2-db1bd35084f1

priseborough:
Rebased

Antiheavy:
@priseborough we did some fixed wing test flights with this PR rolled in (logs below).  The aircraft flew great on all flights, but looking at the logs there are some EKF Yaw/Airspeed innovation check bits during the flights - anything to do with this PR or do you think it is something else?:
image

2 flights:
https://review.px4.io/plot_app?log=fe5f07bc-93ba-4ca1-a99d-fd3437847e7d

2 flights:
https://review.px4.io/plot_app?log=8757a903-70cd-4d79-9073-1f43707b4a51

dagar:
I took a quick look at the first log and these are airspeed innovation test failures, not yaw.

image

I'll ask if we can get them displayed differently in flight review.
 https://github.com/PX4/flight_review/issues/70

Antiheavy:
Thanks @dagar .  I bet our pitot tube is reading low due to the proximity of the static ports to the leading edge of our wing.  We will gather more data and possibly look into an improved calibration.

Issue #346, opened 2017-10-22
build system updates


dagar:
ECL should be a standalone cmake project with several libraries.
 - EKF
 - attitude_fw
 - l1
 - tecs
 - validation
 - geo (would need to be fully moved from PX4)

When included as a subdirectory in another cmake project (like PX4) the same cmake configuration would be used. https://github.com/PX4/Firmware/issues/8191

TODO
 - ECL cmake project with each component generating a static and shared library
 - fully move geo from PX4 (https://github.com/PX4/Firmware/issues/8192)
 - remove PX4 mathlib usage (L1 and TECS)
 - basic testing for all components
 - update CI system to build on linux (range of gcc and clang versions), OSX, and windows (appveyor)
 - fix EKF python testing and post plots (https://github.com/PX4/ecl/pull/317)
 - setup astyle or clang-format (need to ignore EKF autogenerated portions)
 - code coverage

dagar:
Mostly implemented.

PR #355, opened 2017-11-14
[WIP] cmake structure for ecl

Stifael:
This PR is based on this [issue](https://github.com/PX4/ecl/issues/346). The goal of this PR is to make each library within ecl as independent from other libraries as possible to make it easier for other projects outside of PX4 to incorporate ecl into their project. Moreover, a cmake structure was chosen that agrees with modern cmake suggestion from [here](https://www.youtube.com/watch?v=bsXLMQ6WgIk) and [here](https://www.youtube.com/watch?v=eC9-iRN2b04&t=1390s). Doing so enables better cmake maintainability within and outside of the project. 

To build the project do the following:

within ecl:
`mkdir build `
`cd build`
`cmake ../src`
`make`
`make test`

The project structure follows a standard cmake structure with having out of source build:

-build
-src
----CMakeLists.txt
----ekflib
----geolib
-----L1lib
-----mathlib
-----tecslib
-----matrixlib
-----tests

where each lib has the following structure following the include name-space construct.
-----namelib
--------CMakeLists.txt
--------include
-------------namelib
----------------name.h
--------src
------------cpp- and h-files

The exception is of course matrixlib, which is a submodule and therefore has its own structure. 

In contrast to the suggestion of [here](https://github.com/PX4/ecl/issues/346), attitude_fw is not a separate library within ecl. The main reason is the strong dependency on PX4 and therefore it makes more sense to add it to PX4 project directly. 

In addition, I created a separate library for math function (mathlib), which contains the math functions from px4 mathlib [here](https://github.com/PX4/Firmware/tree/master/src/lib/mathlib) without the Quaternion and Matrix related files, which is part of matrixlib. The main reason for creating a separate library for math functions is because there are and will be new flight control related SISO math functions which need to live somewhere, but currently it is not clear where to put them. I personally do not think it makes sense to put SISO math functions into matrix, since I still consider it as a library for vector operations. 

I also added google-tests, which allows to make use of mock objects. 

What still needs to be done:

- [ ] finalize CMakeLists.txt. Currently they are at the bare minimum
- [ ] remove tests and include it within each library. This is easy to do. All it requires is to move the code from the CMakeLists.txt within tests to the top CMakeLists.txt
- [ ] make "validation"  a library and verify where it is needed. 
- [ ] add matrix correctly to the project. Currently the path to math.hpp has to be included to make it compile, which of course is not how it should be done.
- [ ] start discussion about matrix CMakeLists.txt. It would be much easier for any other project that uses the matrix lib if it were a header only target (add_library(matrix INTERFACE)) with all the matrix specific compilation flags as part of the target interface. That way it is very simple to incorporate matrix lib within other projects and ensuring that all compilation properties are preserved. 
- [ ] add python tests 
- [ ] write google tests
- [ ] adjust px4 CMakeLists.txt to correctly include ecl


ChristophTobler:
I guess https://github.com/PX4/ecl/pull/432 replaced this one... Please reopen if it's not the case.

Issue #357, opened 2017-11-15
EKF ringbuffer allocation


dagar:
All ringbuffers are allocated in  EstimatorInterface::initialise_interface() https://github.com/PX4/ecl/blob/master/EKF/estimator_interface.cpp#L382-L392

Only allocating each in configurations where they're actually used will save a lot of memory.

priseborough:
How can I check if a buffer has been allocated?

priseborough:
I could add an accessor function to ringBuffer.h that returns false if _buffer == NULL or check that get_length > 0

dagar:
Depends how you'd want to do it, but that could work. Based on EKF configuration do we know which are needed or is it dependant on the actual pushed.

An easy option is to allocate on the first push, but that might be questionable if for some reason data only started publishing midflight.

dagar:
The ringbuffer is in the class, but the buffer within is allocated with a helper. I could change it so that we only keep pointers to the ringbuffer and the constructor allocates.

Let's figure out how and where the EKF decides it needs a particular ringbuffer and then think about what makes the most sense (including error handling).

priseborough:
Another issue is that the ekf2 code assume that all buffers have been allocated and that data (which may be zeros) is always returned. Checking for buffer allocation each time will add to code complexity.

A compromise solution could be to allocate each buffer initially with a length of 1 and then increase the allocation before the first data is pushed to the buffer.

PR #359, opened 2017-11-16
EKF: Reduce memory usage

priseborough:
This PR reduces the amount of memory required by the observation data buffers by initialising each buffer to an initial length of 1. The buffer is not set to it's required length until the first data arrives. This saves over 1kB of memory because typically only half of the available observation types are used.

To prevent repeated allocation failures causing problems, each buffer is given one attempt to allocate the required memory. If memory cannot be allocated, a fault condition is reported and the corresponding observation will not be used.

Hardware testing TBD.

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -151,6 +151,11 @@ public:
 
 
 
+	// return ttrue if allocated
+	unsigned is_allocated()
+	{
+		return (_buffer != NULL);
+	}
 private:
 	data_type *_buffer{nullptr};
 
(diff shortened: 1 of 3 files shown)


dagar:
Looking good.

Here's ecl master compared to this PR on a default pixhawk 2.1 configured as a plane after GPS fusion starts and both local and global are valid.


                 total      used       free       largest
    Mem:        235376     197168      38208      36848          <- ecl master
    Mem:        235376     192848      42528      39648          <- ecl pr-ekfObsBufAlloc

priseborough:
Addresses https://github.com/PX4/ecl/issues/357

PR #360, opened 2017-11-16
EKF RingBuffer allocation minor improvements + print memory usage

dagar:
Expanding the simple status for command line debugging. This is crude, but helps give you an idea where the memory is going.

	nsh> ekf2 status
	INFO  [ekf2] time slip: -40 us
	INFO  [ekf2] Ekf2 class: 6896 Bytes
	INFO  [ekf2] Ekf2 stack size: 6600 Bytes
	INFO  [lib__ecl] local position valid: no
	INFO  [lib__ecl] global position valid: no
	INFO  [lib__ecl] imu buffer: 22 (888 Bytes)
	INFO  [lib__ecl] gps buffer: 14 (680 Bytes)
	INFO  [lib__ecl] mag buffer: 14 (344 Bytes)
	INFO  [lib__ecl] baro buffer: 14 (232 Bytes)
	INFO  [lib__ecl] range buffer: 1 (24 Bytes)
	INFO  [lib__ecl] airspeed buffer: 1 (24 Bytes)
	INFO  [lib__ecl] flow buffer: 1 (56 Bytes)
	INFO  [lib__ecl] ext vision buffer: 1 (56 Bytes)
	INFO  [lib__ecl] output buffer: 22 (1064 Bytes)
	INFO  [lib__ecl] output vert buffer: 22 (536 Bytes)
	INFO  [lib__ecl] drag buffer: 1 (24 Bytes)

Ideas for future memory savings
 - take a pass over the EstimatorInterface, Ekf, Ekf2 classes and remove anything that doesn't need to persist. What can be calculated as needed? What's already available in another form?
 - use relative timestamps where possible?
 - ekf2_main stack size can be reduced a bit with a small structural change.
 - don't keep params on both sides

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -155,6 +155,8 @@ public:
 		return false;
 	}
 
+	int get_total_size() { return sizeof(*this) + sizeof(data_type) * _size; }
+
 private:
 	data_type *_buffer{nullptr};
 
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -45,7 +45,13 @@ template <typename data_type>
 class RingBuffer
 {
 public:
-	RingBuffer() = default;
+	RingBuffer() {
+		if (allocate(1)) {
+			// initialize with one empty sample
+			data_type d = {};
+			push(d);
+		}
+	}
 	~RingBuffer() { delete[] _buffer; }
 
 	// no copy, assignment, move, move assignment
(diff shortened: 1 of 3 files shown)


dagar:
A small simplification in the allocation logic on top of https://github.com/PX4/ecl/pull/359

dagar:
In each case it should be checking the corresponding buffer fail bool next to the initialised flag.

priseborough:
This got closed when I deleted the branch used for #359 after merging it.

PR #362, opened 2017-11-21
EKF: Add additional velocity interface to use landing beacon data

priseborough:
See https://github.com/PX4/Firmware/pull/8160 for details

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -156,6 +156,12 @@ struct dragSample {
 	uint64_t time_us;	///< timestamp of the measurement (uSec)
 };
 
+struct auxVelSample {
+	Vector2f velNE;		///< measured NE velocity relative to the local origin (m/sec)
+	Vector2f velVarNE;	///< estimated error variance of the NE velocity (m/sec)**2
+	uint64_t time_us;	///< timestamp of the measurement (uSec)
+};
+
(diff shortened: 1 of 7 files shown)


TSC21:
@priseborough I ask you if it is possible to extend this velocity interface to include the vertical velocity Z component? The reason is that for the case of the beacon, we are using the X/Y velocities, but when we have full pose estimation of a landing target being visually tracked (which I'll be adding the extension after PX4/Firmware#8160 is merged), we also have the Z component of it.

priseborough:
It is possible, but should be done in conjunction with extending the landing_target_pose message. This  should include the z states, corresponding variances and separate validity for the vertical states.

When this is in place I will create another PR for ecl/EKF.

TSC21:
> It is possible, but should be done in conjunction with extending the landing_target_pose message. This should include the z states, corresponding variances and separate validity for the vertical states.
>
>When this is in place I will create another PR for ecl/EKF.

Sounds good to me. Will add it then when I do I do the upgrade.

priseborough:
SITL test log here: https://logs.px4.io/plot_app?log=97145626-db8c-4743-90f8-c4632a3bcde8

The following plot shows the velocity innovations from the GPS and landing target.

screen shot 2017-11-21 at 11 54 26 am

PR #365, opened 2017-11-24
EKF: Fix vel pos innovation handling bug

priseborough:
Fixes a bug introduced by commit 882f4d4 that incorrectly handles vertical velocity and position innovations.

Before - vertical innovations are not output by the EKF:

before

After - now they are:

after


LorenzMeier:
Did you also send a PR against PX4 upstream to fix?

priseborough:
I'll do that now. The version used by upstream doesn't have the bug that this fixes.

Issue #363, opened 2017-11-23
initial error message "EKF stopping navigation"


dagar:
This is low priority and only a minor annoyance, but the EKF is printing "EKF stopping navigation" at the beginning of many flights.

For example
https://logs.px4.io/plot_app?log=510f330b-db14-4398-9d8a-9fe6f9c28638

![image](https://user-images.githubusercontent.com/84712/33187081-2844a496-d05d-11e7-901d-cc9935c050d0.png)

Can this be initialized differently to skip the error? https://github.com/PX4/ecl/blob/882f4d44e5b9111c4ace4348fa82de5c58c351d1/EKF/control.cpp#L1338

PR #367, opened 2017-12-03
EKF: Don't print navigation lost message on startup

priseborough:
Fixes https://github.com/PX4/ecl/issues/363

screen shot 2017-12-04 at 9 51 26 am


LorenzMeier:
Ready for Firmware PR

PR #369, opened 2017-12-12
EKF: Release flow speed limit progressively

priseborough:
When GPS use is gained whilst flying using optical flow data, the release of the speed limit can cause unexpected change in responsiveness the first time the operator moves the vehicle after the limit is released.

When another source of external aiding (eg GPS) is gained, this patch releases the speed limit as height is gained, but does not reduce it when the vehicle descends,  unless other aiding is lost and optical flow becomes the only source of aiding again.

Flight tested on a px4fmu-v4 board using https://github.com/PX4/Firmware/pull/8223

Test Log: https://logs.px4.io/plot_app?log=4bdc4540-9a52-450a-a14e-afeac2606f3d


priseborough:
Testing was performed by taking off in position mode before GPS checks passed from an under cover location and flying out to where there was limited sky view, verifying that the horizontal speed was being limited. The vehicle height was increased to 3m to improve sky view and enable GPS checks to pass. The vehicle was progressively reduced in height and manoeuvred to confirm that the speed limit did not reduce with reducing altitude.

screen shot 2017-12-12 at 11 29 48 am

PR #371, opened 2017-12-18
ekf: return beta test ratio in get_innovation_test_status

CarlOlsson:
Also requires changes on the [firmware side](https://github.com/PX4/Firmware/pull/8489)


priseborough:
Looks good to me. Can you push to a branch in the upstream repo and reference it in your Firmware PR so that testing and merging can be coordinated.

priseborough:
@CarlOlsson I will merge it now. Please update https://github.com/PX4/Firmware/pull/8489.

PR #372, opened 2017-12-21
TECS initialize throttle integrator at cruise if in air

dagar:


priseborough:
Code looks good. I would want to see a VTOL log (SITL or hardware) before merging to ensure the change in behaviour is as expected. Can you raise a WIP PR on Firmware using this commit so we can get the interactions checked?

PR #377, opened 2018-01-12
Reverse the linked list of data_validator_group

lijieamd:
---

maintain a first node for search from index 0 to fix issue PX4/Firmware#8644

before fix:
I tested with jlink debug on real hardware
I have 2 mag sensor connected to board, mag0 have orb priority 75(ORB_PRIO_DEFAULT), mag1 have orb priority 125(ORB_PRIO_VERY_HIGH). mag0 publish first, and then mag1, image below show screen shot of breakpoint at the moment before adding mag1 to validator group, you can see **_first** is mag0 (which has priority 75)
image

---

this image is screen shot of breakpoint at the first time mag0 data put to validator after adding mag1, see the mess here, **_first** is actually the last one we added to group(mag1), **_priority** field is empty(not set yet), but we are putting mag0 data!
image

---

this image is screen shot after bug fixed
image

### DIFF
--- a/validation/data_validator.h
+++ b/validation/data_validator.h
@@ -48,7 +48,7 @@ class __EXPORT DataValidator {
 public:
 	static const unsigned dimensions = 3;
 
-	DataValidator(DataValidator *prev_sibling = nullptr);
+	DataValidator();
 	virtual ~DataValidator() = default;
 
 	/**
@@ -71,6 +71,12 @@ public:
 	 * @return		the next sibling
 	 */
 	DataValidator*		sibling() { return _sibling; }
+
+	/**
+	 * Set the sibling to the next node in the group
+	 *
+	 */
+	void			sibling(DataValidator* sibling) { _sibling = sibling; }
 
 	/**
 	 * Get the confidence of this validator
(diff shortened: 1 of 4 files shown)


validation/data_validator.h
```diff
@@ -72,6 +72,12 @@ class __EXPORT DataValidator {
 	 */
 	DataValidator*		sibling() { return _sibling; }
 
+	/**
+	 * Set the sibling to the next node in the group
+	 *
+	 */
+	void			sibling(DataValidator* sibling) { _sibling = sibling; }
```
bkueng: Can you rename to `setSibling`?

validation/data_validator_group.cpp
```diff
@@ -45,19 +45,31 @@
...
-	DataValidator *next = _first;
+	DataValidator *next = nullptr;
+	DataValidator *prev = nullptr;
 
 	for (unsigned i = 0; i < siblings; i++) {
-		next = new DataValidator(next);
+		next = new DataValidator();
+		if(i == 0) {
+			_first = next;
+		}
```
bkueng: Style: should be `} else {`

PR #378, opened 2018-01-13
attitude_fw: constrain integrator properly to prevent building it up …

AndreasAntener:
…over the specified maximum

Stuff you find when it's being logged... Could go in with https://github.com/PX4/Firmware/pull/8366 when the rest of the EKF updates are settled.


dagar:
We really need minimal CI on this side.

PR #382, opened 2018-01-23
Revert "EKF: Release flow speed limit with altitude gained"

priseborough:
This reverts commit e70206f74bc4f77a0e000da3c923f4459006d2b1.

The reason for this is that the release in the optical flow speed limit only with altitude after commencing GPS fusion was not liked by testers and has been replaced with a better method implemented inside the position controller.

See https://github.com/PX4/Firmware/pull/8223 for details


priseborough:
The data validator changes have also been reverted because these are causing the following CI errors for some compiler types:

screen shot 2018-01-24 at 4 51 04 pm

@bkueng FYI

Issue #390, opened 2018-01-30
Vertical position derivative responds to z position resets


priseborough:
The state used to calculate vertical derivative estimate returned by get_pos_d_deriv(...) is not being reset when the vertical position is reset:

![screen shot 2018-01-30 at 10 56 21 pm](https://user-images.githubusercontent.com/3596952/35565328-4165b03e-0611-11e8-8d4e-c7d0dcc93c9a.png)

This will cause rapid height change if a reset event occurs at the same time as a pilot climb rate demand due to https://github.com/PX4/Firmware/blob/master/src/modules/mc_pos_control/mc_pos_control_main.cpp#L2372-L2378

See https://review.px4.io/plot_app?log=fe322cb8-ff2d-444e-9d28-96e593179fc5 for an affected flight.

PR #391, opened 2018-01-31
EKF: Add missing reset for output observer vertical position derivative

priseborough:
Required to prevent large transients in value returned by get_pos_d_deriv accessor following a vertical state reset event.

Fixes https://github.com/PX4/ecl/issues/390.

Bench tested by adding a time based baro offset of +100m at 50 seconds in ekf2_main.cpp. vertical derivative estimate now resets to same value as velocity following a reset.

screen shot 2018-01-31 at 11 59 48 am
m


LorenzMeier:
@ChristophTobler Sending the PR for Firmware? Thanks!

ChristophTobler:
@LorenzMeier Yup. Btw. I somehow lost write access to this repo... Can you fix that?

PR #393, opened 2018-01-31
In realignYawGPS() reset the wind estimate and dont use it for t…

CarlOlsson:
…he yaw align

If a fixed wing experiences a bad heading estimate in cruise flight the wind estimate will most likely be very bad, (due to sideslip and airspeed fusion during the period from the bad heading estimate to the reset). Using this bad wind estimate to re-align the heading is thus risky and should be avoided.

@priseborough interested to hear your opinion

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -398,14 +398,11 @@ bool Ekf::realignYawGPS()
 				euler321(2) += courseYawError;
 				_flt_mag_align_complete = true;
 
-			} else if (_control_status.flags.wind) {
-				// we have previously aligned yaw in-flight and have wind estimates so set the yaw such that the vehicle nose is
-				// aligned with the wind relative GPS velocity vector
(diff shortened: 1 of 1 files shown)


priseborough:
That makes sense now that we are doing sideslip fusion. In the past with only airspeed fusion, the wind estimate was not as sensitive sensitive to yaw errors.

priseborough:
Having said that it makes sense, we are still in a situation where, without test data, we do not  know that this is a net benefit. Do you have any SITL or flight test data?

CarlOlsson:
I briefly looked through our data and couldn't find a good example of an in flight reset with a decent amount of wind and sideslip fusion active. Will update here as soon as I find something

priseborough:
Ping?

CarlOlsson:
I could not find any data with a reasonably new version of ecl verifying this so closing here. Lets come back to it if it is needed

PR #395, opened 2018-02-05
EKF: fuse EV with offset instead of incrementally

ChristophTobler:
@priseborough Why can't we fuse EV like in this PR? Specially when EV is in NED and we use EV and GPS. In that case the frame is the one from the EV anyways, right?

The reason I want to change it is that the performance with the incremental fusion isn't very good. You can test it by hardcoding https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L262 to `true` and use the simulation `make posix_sitl_default gazebo_iris_opt_flow` with `EKF2_AID_MASK` set to 24. It is drifting heavily.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -252,7 +252,7 @@ void Ekf::controlExternalVisionFusion()
 			_ev_sample_delayed.posNED(1) -= pos_offset_earth(1);
 			_ev_sample_delayed.posNED(2) -= pos_offset_earth(2);
 
-			// Use an incremental position fusion method for EV data if using GPS or if the observations are not in NED
+			// Check if we need to apply an offset before fusing EV data
 			if (_control_status.flags.gps || (_params.fusion_mode & MASK_ROTATE_EV)) {
 				_fuse_hpos_as_odom = true;
 			} else {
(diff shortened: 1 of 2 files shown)


ChristophTobler:
If the EV data is not in NED we would probably need to wait untel `_ev_rot_mat` has converged.

ChristophTobler:
Info: if the EV data is not in NED, this version nor the current master version is working. I tested this by changing https://github.com/PX4/Firmware/blob/master/src/modules/simulator/simulator_mavlink.cpp#L1131-L1132 to 
```
vision_position.x = -ev_mavlink->y;
vision_position.y = ev_mavlink->x;
```
(270 degrees rotation about z)

But do we actually need to support data that is not in local NED?

mhkabir:
> But do we actually need to support data that is not in local NED?

Yes. VIO normally does not have a global heading reference to init to.

ChristophTobler:
@mhkabir Yes, I know (almost never). I got confused with the use of NED. A lot of times this is used instead of forward-right-down even if there's no actual north/east. And because a VIO almost never has a geographical heading, I was always thinking of frd...

PR #396, opened 2018-02-08
ekf fix ev rotation calculation

ChristophTobler:
This fixes the external vision rotation matrix calculation. There were two issues:
* The spike handling added a fraction of `rot_vec` instead of `_input_delta_vec` to `_ev_rot_vec_filt`
* The vector from quaternion calculation must have had a bug somewhere. I just swapped to using the matrix library which also reduces duplicate code.

This has been tested in sitl using `make posix_sitl_default gazebo_iris_opt_flow` with `EKF2_AID_MASK` set to `72`.


ChristophTobler:
This has also been tested as part of https://github.com/PX4/ecl/issues/397

PR #401, opened 2018-02-27
EKF: Fix bug in use of gps velocity noise parameter

priseborough:
This fixes a bug introduced by an earlier feature request PR. The noise parameter is supposed to define the lower limit on the observation noise.

Bug spotted by Deon Blaauw https://px4.slack.com/archives/C643UJTB7/p1519744752000030


priseborough:
screen shot 2018-02-28 at 11 47 24 am

Velocity innovation variance now scales with reported speed accuracy. Test performed by covering GPS antenna to induce high nose levels.

PR #404, opened 2018-03-05
attitude_fw: move angular rate limits to body angular rates

tstastny:
Angular rate limits should be applied to body angular rate references, as opposed to the non-transformed Euler rate references.

(follow-up from #402)


dagar:
Reviewed and tested in https://github.com/PX4/Firmware/pull/9023.

PR #405, opened 2018-03-14
added airdata module which contains wind & airspeed scale estimator

RomanBapst:
I would like to add the wind & airspeed scale estimator to the ecl library since I think it's the best place for it to live.
Just like the ekf it has a clear interface and could easily be deployed on other platforms.

On the Firmware side we still need to discuss if we want to integrate this into the ekf2 module or if we implement a standalone work queue which uses this library.

TODO:
 - [ ] documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ px4_add_module(
 	STACK_MAX 6000
 	COMPILE_FLAGS
 	SRCS
+		airdata/WindEstimator.cpp
 		attitude_fw/ecl_controller.cpp
 		attitude_fw/ecl_pitch_controller.cpp
 		attitude_fw/ecl_roll_controller.cpp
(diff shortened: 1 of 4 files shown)


dagar:
We should have this building standalone in the ecl project. All of these modules should build as a separate lib (not using px4_add_module), then on the PX4 side it can include the entire thing and link to the specific pieces required (ekf, tecs, etc)

mhkabir:
@RomanBapst How does this work tie in with the multirotor wind estimator?

RomanBapst:
@mhkabir The multirotor wind estimator is part of ekf2, this is a standalone library. While ekf2 can estimate wind for both multirotor and fixed wing this library specifically targets fixed wings. In future we want to use it for three specific tasks:

- estimate wind
- estimate airspeed scale factor
- estimate airspeed failure, possible using two separate source of airspeed

dagar:
@RomanBapst there's no accompanying standalone build with this, so unless you want to add one now you might as well merge.

PR #407, opened 2018-03-15
EKF decrease synthetic sideslip interval 1s -> 150ms

dagar:


priseborough:
Have you tested the effect of the faster rate in SITL?

dagar:
Yes while testing https://github.com/PX4/Firmware/pull/8771 I had it set to 10Hz and it seemed to work fine (minus the timeouts marking xy_valid = false). I could try it again with this specific setting relative to master if there's anything specific to examine.

priseborough:
No, that will be adequate if you have tested at a slightly higher speed already.

PR #408, opened 2018-03-16
attitude_fw delete unused and cleanup

dagar:
TODO: review gain scheduling

### DIFF
--- a/attitude_fw/ecl_controller.cpp
+++ b/attitude_fw/ecl_controller.cpp
@@ -1,6 +1,6 @@
 /****************************************************************************
  *
- *   Copyright (c) 2013 Estimation and Control Library (ECL). All rights reserved.
+ *   Copyright (c) 2013-2018 Estimation and Control Library (ECL). All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -48,98 +48,30 @@
 
 #include "ecl_controller.h"
 
-#include <stdio.h>
-#include <mathlib/mathlib.h>
+using math::constrain;
+using math::max;
+using math::min;
 
-ECL_Controller::ECL_Controller(const char *name) :
-	_last_run(0),
-	_tc(0.1f),
-	_k_p(0.0f),
-	_k_i(0.0f),
-	_k_ff(0.0f),
-	_integrator_max(0.0f),
-	_max_rate(0.0f),
-	_last_output(0.0f),
-	_integrator(0.0f),
-	_rate_error(0.0f),
-	_rate_setpoint(0.0f),
-	_bodyrate_setpoint(0.0f)
+void ECL_Controller::integrate(const float rate_error)
 {
-}
+	const float dt = ecl_elapsed_time(&_last_run) * 1e-6f;
+	_last_run = ecl_absolute_time();
 
-void ECL_Controller::reset_integrator()
-{
-	_integrator = 0.0f;
-}
+	// lock integral for long intervals
+	if ((_k_i > 0.0f) && (dt < 0.5f)) {
+		float id = rate_error * dt;
 
-void ECL_Controller::set_time_constant(float time_constant)
-{
-	if (time_constant > 0.1f && time_constant < 3.0f) {
-		_tc = time_constant;
+		// anti-windup: do not allow integrator to increase if actuator is at limit
+		if (_last_output < -1.0f) {
+			// only allow motion to center: increase value
+			id = max(id, 0.0f);
+
+		} else if (_last_output > 1.0f) {
+			// only allow motion to center: decrease value
+			id = min(id, 0.0f);
+		}
+
+		// add and constrain
+		_integrator = constrain(_integrator + id * _k_i, -_integrator_max, _integrator_max);
 	}
-}
-
-void ECL_Controller::set_k_p(float k_p)
-{
-	_k_p = k_p;
-}
-
-void ECL_Controller::set_k_i(float k_i)
-{
-	_k_i = k_i;
-}
-
-void ECL_Controller::set_k_ff(float k_ff)
-{
-	_k_ff = k_ff;
-}
-
-void ECL_Controller::set_integrator_max(float max)
-{
-	_integrator_max = max;
-}
-
-void ECL_Controller::set_max_rate(float max_rate)
-{
-	_max_rate = max_rate;
-}
-
-void ECL_Controller::set_bodyrate_setpoint(float rate)
-{
-	_bodyrate_setpoint = math::constrain(rate, -_max_rate, _max_rate);
-}
-
-float ECL_Controller::get_rate_error()
-{
-	return _rate_error;
-}
-
-float ECL_Controller::get_desired_rate()
-{
-	return _rate_setpoint;
-}
-
-float ECL_Controller::get_desired_bodyrate()
-{
-	return _bodyrate_setpoint;
-}
-
-float ECL_Controller::get_integrator()
-{
-	return _integrator;
-}
-
-float ECL_Controller::constrain_airspeed(float airspeed, float minspeed, float maxspeed)
-{
-	float airspeed_result = airspeed;
-
-	if (!ISFINITE(airspeed)) {
-		/* airspeed is NaN, +- INF or not available, pick center of band */
-		airspeed_result = 0.5f * (minspeed + maxspeed);
-
-	} else if (airspeed < minspeed) {
-		airspeed_result = minspeed;
-	}
-
-	return airspeed_result;
 }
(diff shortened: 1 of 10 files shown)


tstastny:
I cannot think of a reason to have linear airspeed scalers on the feedforward terms, while quadratic are on the proportional and integral signals. Unless someone has a specific reason to leave it this way, I would be favor of making it all one quadratic airspeed scaler on the output of the sum of the ff p and i signals.

dagar:
Updated.

bresch:
@tstastny The reason is that the feedforward is used to compensate for the natural rate damping. For example, in roll, the nondimensional roll moment derivative can be modeled using (1) and generates a roll moment by (2).
Assuming a symmetric aircraft and zero sideslip, the equation can be simplified using only the rollrate damping and the roll moment produced by the ailerons (3).
During a roll at constant speed, the torque produced by the ailerons should exactly compensate for the damping (4).
Simplifying (5) and rearranging to extract the ideal ailerons deflection gives (6).
The first fraction gives the value of the ideal feedforward and we can see that the scaling is linear to airspeed. (The negative sign is then absorbed by the aileron control derivative which is also negative)

<a href="https://www.codecogs.com/eqnedit.php?latex=\\&space;C_\ell&space;=&space;C_{\ell_0}&space;&plus;&space;C_{\ell_\beta}\:\beta&space;&plus;&space;C_{\ell_p}\:\frac{b}{2V_T}\:p&space;&plus;&space;C_{\ell_{\delta_a}}&space;\:\delta_a&space;\quad(1)&space;\\&space;\ell&space;=&space;\frac{1}{2}\rho&space;V_T^2&space;S&space;C_\ell&space;\quad&space;(2)&space;\\&space;\ell&space;=&space;\frac{1}{2}\rho&space;V_T^2&space;S&space;\left&space;[C_{\ell_{\delta_a}}&space;\:\delta_a&space;&plus;&space;C_{\ell_p}\:\frac{b}{2V_T}&space;\:&space;p&space;\right&space;]&space;\quad&space;(3)&space;\\&space;-&space;\frac{1}{2}\rho&space;V_T^2&space;S&space;C_{\ell_{\delta_a}}&space;\:\delta_a&space;=&space;\frac{b}{4}\rho&space;V_T&space;S&space;\:&space;p&space;\quad&space;(4)&space;\\&space;-&space;V_T&space;C_{\ell_{\delta_a}}&space;\:\delta_a&space;=&space;\frac{b}{2}&space;\:&space;p&space;\quad&space;(5)&space;\\&space;\delta_a&space;=&space;-\frac{b}{2&space;C_{\ell_{\delta_a}}}&space;\frac{1}{V_T}&space;\:&space;p&space;\quad&space;(6)" target="_blank"></a>

Where 
V_T : trim airspeed of the model
S : ref surface
b : wing span
p : rollrate
delta_a : aileron deflection
beta : angle of sideslip
rho : air density
l (\ell) : roll torque
C_x : nondimentional derivative of x

tstastny:
> The first fraction gives the value of the ideal feedforward and we can see that the scaling is linear to airspeed.

Thanks for the clarification @bresch -- however, I'm curious if the airspeed in the denominator of the coefficient actually has any *dynamic* physical meaning, or is only meant to be a relative scaling term chosen for comparing coefficients between various aircraft of different configuration and flight operating points; e.g., comparing the efficiency of a lift coefficient for a glider with large span at slow speed vs a lift coefficient for a cessna with small span, and higher speed. In my flight dynamics courses (many years ago) this is how it was always described to us. 

I suppose if one wanted to use a similar approach with the scalar here to keep the feedforward gains themselves "comparable" between various aircraft, this could also be ok. But if you can point to an actual physical derivation of why the VT is in the denominator of the coefficient (besides simply a non-dimensionalization choice for units and characteristic parameters), I am always happy to learn something new. Otherwise, I guess the only truly physical term needed for scaling comes from the VT^2 -- i.e. dynamic pressure scaling.

wdyt?

bresch:
@tstastny Thanks for the question, I actually never questioned this formula and had to read a bit to find an answer. 

- Firstly, (b/2) * p is the vertical wingtip speed and thus, the ratio (b * p)/(2 * V) is the angle that the wingtips make with the flight path and is called the _helix angle_.
- Secondly, the roll damping coefficient of a rigid wing is constant regardless the airspeed. This coefficient only changes at higher Mach number (> 0.6) due to aeroelasticity (wing twist).
- Thirdly, a constant aileron deflection produces a constant helix angle

This together shows that there is a physical linear airspeed relation between the roll rate and the aileron deflection.
The following reference explains quite well why I just summarize and gives nice figures: 
Bill Crawford, [Flightlab Ground School](http://www.flightlab.net/Flightlab.net/Download_Course_Notes_files/9_RollingDynamics.pdf), 9.5-9.9, "Rolling Dynamics"

The references below also helped me to find the answer:
**Helix angle** : Stevens & Lewis, _Aircraft Control ans Simulation_, 3rd edition, p.80-81, "The aerodynamic derivatives"
**Constant roll damping** : Stevens & Lewis, _Aircraft Control ans Simulation_, 3rd edition, p.93-94, "Rolling moment, Figure 2.3-6"
**Lateral control power, helix angle**: Perkins & Hage, _Airplane Performance Stability and Control_, p.352-353

dagar:
It would be great to include a brief explanation and references in the comments.

priseborough:
That was the original reasoning. Steady state roll rate for a given aileron deflection is proportional to true airspeed, whereas instantaneous change in roll acceleration due to a change in aileron deflection is proportional to dynamic pressure (indicated airspeed ^2).

Keeping the distinction between true and indicated airspeed in the calculation of the different gain paths is important.

hamishwillee:
> It would be great to include a brief explanation and references in the comments.

@bresch @priseborough Is this something that should/could live in the guide too?

bresch:
@priseborough Thanks for the precision, I completely ignored that point!

I needed a mathematical proof to convince myself, so I share the equations here:

At low altitude and speed, the indicated airspeed (IAS, V_I) can be converted to true airspeed (TAS, V_T) using a _density error_ factor (1). Squaring, rearranging and adding a 1/2 factor to both sides makes the dynamic pressure expression appear (2). We can now easily see that the dynamic pressure is proportional to the IAS squared (3)

<a href="https://www.codecogs.com/eqnedit.php?latex=\\&space;V_T&space;=&space;V_I&space;\sqrt{\frac{\rho_0}{\rho}}&space;\quad(1)\\&space;\bar{q}&space;=&space;\frac{1}{2}&space;\rho&space;V_T^2&space;=&space;\frac{1}{2}&space;V_I^2&space;\rho_0&space;\quad(2)\\&space;\bar{q}&space;\propto&space;V_I^2&space;\quad(3)&space;\\&space;\\&space;V_T&space;:&space;\mathrm{True&space;\:&space;airspeed}&space;\\&space;V_I&space;:&space;\mathrm{Indicated&space;\:&space;airspeed}&space;\\&space;\bar{q}&space;:&space;\mathrm{Dynamic&space;\:&space;pressure}&space;\\&space;\rho&space;:&space;\mathrm{Air&space;\:&space;density}&space;\\&space;\rho_0&space;:&space;\mathrm{Air&space;\:&space;density&space;\:&space;at&space;\:&space;sea&space;\:&space;level}&space;\\" target="_blank"></a>

I only found this [ref](https://www.experimentalaircraft.info/flight-planning/aircraft-navigation-speed.php) for the IAS to TAS formula used in PX4, does someone have a better reference?

@hamishwillee Absolutely! :)

hamishwillee:
> @hamishwillee Absolutely! :)

@bresch Great, you've scored yourself a job then [evil snigger]. Is that OK?

The question is where? The only ECL doc we have is [Using the ecl EKF](https://dev.px4.io/en/tutorials/tuning_the_ecl_ekf.html). Is this a good place?

Looking at this doc it appears to me that we use the term ECL and EKF largely interchangeably. How exactly are they related, along with LPE? (It could be that ECL implements the EKF and the terms are synonymous, or it could be that the ECL is a higher level library that owns other functions and can interchange the control system used - ie ECL plugs in either EKF or LPE)

Either way, I think this doc should move under [Concepts](https://dev.px4.io/en/concept/), possibly under a topic Estimators, that also includes https://dev.px4.io/en/advanced/switching_state_estimators.html

LorenzMeier:
@bresch Any news on this?

bresch:
@LorenzMeier About the documentation, I have a draft version that Hamish reviewed, I should be able to make a PR this week. I should make that scaling technique clear and understandable to everyone.
I'm currently reviewing this PR and I'll make sure that the IAS and TAS scalers used correctly.

priseborough:
I'm closing this. The documentation introduced by https://github.com/PX4/Devguide/pull/555 explains the scaling.

PR #411, opened 2018-03-19
tecs.cpp: tecs initialisation fixes

RomanBapst:
When initialised in air the throttle integrator was set to cruise throttle which does not make any sense.
The cruise throttle value is separately added to the throttle value and does not need to be part of the integrator.
However, for the last throttle value variable (used for slew rate limiting) it makes sense to initialise it with cruise throttle when TECS is initialised in air.

### DIFF
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -522,9 +522,9 @@ void TECS::_initialize_states(float pitch, float throttle_cruise, float baro_alt
 		_vert_pos_state = baro_altitude;
 		_tas_rate_state = 0.0f;
 		_tas_state = _EAS * EAS2TAS;
-		_throttle_integ_state = (_in_air ? throttle_cruise : 0.0f);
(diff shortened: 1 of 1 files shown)


RomanBapst:
It might seem that this PR reverses the changes done in #372 to fix https://github.com/PX4/Firmware/issues/8508

However, after looking at the log in https://github.com/PX4/Firmware/issues/8508
I think that the throttle was cut by TECS itself because when it took control over the vehicle there was a large negative energy and energy rate error. As a result TECS reduced throttle rapidly to zero in order to reduce the energy rate into the system.
You can reproduce this in SITL by flying plane in stabilized, climbing as fast you can and then suddenly switch into altitude control. The system will have an excessive amount of energy and thrust will go to 0 for a while.

I tested this PR in SITL using the plane and the standard VTOL simulation, in both cases the throttle initialises just fine with this PR.

PR #413, opened 2018-03-21
EKF: Enable extended dead reckoning after GPS loss for FW vehicles

priseborough:
Fixes https://github.com/PX4/ecl/issues/414 enabling extended dead reckoning after GPS loss

Fixes an issue with excessively  short time from loss of GPS to reported loss of nav validity encountered during testing of https://github.com/PX4/Firmware/pull/8771#issuecomment-374826607

- Removes duplicate checking for dead reckoning and consolidate into a single function.
- Use separate timers to check for start of dead reckoning and check when dead reckoning has been performed for too long for the nav solution to be valid.
- Allows the timeout for validity reporting to be adjusted externally.
- Separates external reporting of dead reckoning from internal checks.

Recommend testing in combination with https://github.com/PX4/Firmware/pull/8771


dagar:
Looks good. The other thing we should check is the vehicle_global_position deadreckoning flag. In https://github.com/PX4/Firmware/pull/8771#issuecomment-373228891 it wasn't updating. I'm also wondering if we should move that flag from global to local.

priseborough:
The only difference between what is returned by the global_position_is_valid() and local_position_is_valid() accessor functions is that global position requires an origin.

dagar:
Rebased on master.

priseborough:
@dagar If you can update your review, then we can get this merged.

priseborough:
The last patch fixes the issue that was causing global position eph to go to zero before global position stopped publishing

priseborough:
Here is the log: https://logs.px4.io/plot_app?log=abbb7ebf-259d-432a-bf08-b137ce88df90

Reported global position error no longer falls before global position topic stops publishing.

screen shot 2018-04-19 at 11 26 04 am

dagar:
@priseborough given we need a similar change for the sync between local position eph/epv and local_position_is_valid() could we avoid it entirely? Why not just skip zeroing the reported errors?

Any consumer can independently check the validity flags, or in the case of global position won't even have a publication.

PR #417, opened 2018-04-03
Optical flow doreset

huangwen0907:
@priseborough  when the optical is not fused for a long time, can it do the reset process?


priseborough:
This will not work because:

1) The velocity and position reset functions cannot handle a reset using optical flow when moving (they need to be modified)
2) We should not reset if optical flow is lost if we have other aiding sources.

I will create a PR ready for you to test tomorrow.

PR #418, opened 2018-04-04
EKF: Fix non GPS aiding data reset logic

priseborough:
This fixes the following problems with use of optical flow and external vision data.

Both the optical flow and external vision data fusion logic were missing functionality required to ensure that the EKF would reset to the observation if innovation checks failed for too long and the data was the only source of aiding.

The velocity reset method did not handle optical flow measurements and instead initialised velocity to zero if optical flow data was the aiding source.

The velocity reset function did not separate the optical flow and external vision use cases properly.

The position reset function selected optical flow data before external vision data, even though optical flow data does not provide position information.

The timeouts used to declare stoppage of optical flow and external vision use were hardcoded.

If the vehicle is on ground and shaken and / or rotated rapidly whilst relying on optical flow for aiding, the estimator would diverge and have to reset.

Testing https://logs.px4.io/plot_app?log=bbbeaf62-23a6-4980-9bc8-991386e6f469

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -289,6 +289,15 @@ void Ekf::controlExternalVisionFusion()
 				// use the absolute position
 				_vel_pos_innov[3] = _state.pos(0) - _ev_sample_delayed.posNED(0);
 				_vel_pos_innov[4] = _state.pos(1) - _ev_sample_delayed.posNED(1);
+
+				// check if we have been deadreckoning too long
+				if (_time_last_imu - _time_last_pos_fuse > _params.no_gps_timeout_max) {
+					// don't reset velocity if we have another source of aiding constraining it
+					if (_time_last_imu - _time_last_of_fuse > (uint64_t)1E6) {
+						resetVelocity();
+					}
+					resetPosition();
+				}
 			}
 
 			// observation 1-STD error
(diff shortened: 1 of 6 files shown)


EKF/control.cpp
```diff
@@ -401,12 +367,22 @@ void Ekf::controlOpticalFlowFusion()
...
 		}
 
+		// handle the case when we have optical flow, are reliant on it, but have not been using it for an extended period
+		if (_control_status.flags.opt_flow
+						&& !_control_status.flags.gps
+						&& !_control_status.flags.ev_pos) {
+			bool do_reset = _time_last_imu - _time_last_of_fuse > _params.no_gps_timeout_max;
+			if (do_reset) {
+				resetVelocity();
+				resetPosition();
```
huangwen0907: Do we need print a info here?

EKF/ekf_helper.cpp
```diff
@@ -125,6 +155,20 @@ bool Ekf::resetPosition()
 		// use EV accuracy to reset variances
 		setDiag(P,7,8,sq(_ev_sample_delayed.posErr));
 
+	} else if (_control_status.flags.opt_flow) {
+		if (!_control_status.flags.in_air) {
+			// we are likely starting OF for the first time so reset the horizontal position and vertical velocity states
```
mhkabir: Comment is inconsistent with what is done below. Where are we resetting vertical velocity?

huangwen0907: yes, I found that when horizontal velocity is diverged, the vertical velocity is also not correct sometimes.

priseborough:
Replay test log here: https://logs.px4.io/plot_app?log=bbbeaf62-23a6-4980-9bc8-991386e6f469

huangwen0907:
@priseborough could you merge this pr? I have test it, everything work as expected.

EKF/terrain_estimator.cpp
```diff
@@ -134,40 +137,59 @@ void Ekf::fuseHagl()
...
+		  ((_time_last_imu - _time_last_hagl_fuse) < (uint64_t)5e6)) {
 		return true;
 
 	} else {
 		return false;
 	}
 }
 
+// determine terrain validity
+void Ekf::update_terrain_valid()
```
ChristophTobler: This function does almost the same as `get_terrain_valid()`. Can we reuse that one to set `_hagl_valid` otherwise we need to add/remove checks in both locations if we change something.

PR #425, opened 2018-04-21
EKF always fill position and velocity accuracy

dagar:
https://github.com/PX4/ecl/pull/413#issuecomment-383314867

Any consumer can independently check the validity flags, or in the case of global position won't even have a publication.


dagar:
Alternatively I would propose setting these to positive infinity rather than 0.

priseborough:
We aren't using the dead reckoning flag returned so i am OK with removing that form those interfaces. Returning +inf is a pain when plotting data so do not want to go there. I'm happy with these changes as is.

PR #428, opened 2018-05-02
validation add consistency checks

dagar:


priseborough:
No supporting or contextual information provided and this is now out of date.

PR #429, opened 2018-05-03
EKF: Fix bug causing simultaneous range and baro height fusion

priseborough:
This fixes a bug that causes simultaneous fusion of range finder and baro measurements when EKF2_RNG_AID is enabled.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -836,8 +836,7 @@ void Ekf::controlHeightFusion()
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


mhkabir:
@dagar Can we get a submodule update on Firmware for this?

dagar:
https://github.com/PX4/Firmware/pull/9400

Ahrovan:
Is there a solution to this problem? Or unrelated
EKF baro hgt timeout - reset to baro #9060

Issue #430, opened 2018-05-03
elseif never reached


bartslinger:
Was going through the code and noticed that this is never called:
https://github.com/PX4/ecl/blob/master/EKF/ekf.cpp#L221

Because of https://github.com/PX4/ecl/pull/388

Not exactly sure yet what the implication is.

ChristophTobler:
@bartslinger I guess you are right. That got forgotten in https://github.com/PX4/ecl/pull/388
Can you create a PR?

bartslinger:
Sure, just didn't do it yet because I'm not 100% sure if removing it was the correct solution. PR is here https://github.com/PX4/ecl/pull/431

PR #431, opened 2018-05-03
remove superfluous elseif

bartslinger:
Fixes https://github.com/PX4/ecl/issues/430

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -189,41 +189,30 @@ bool Ekf::initialiseFilter()
 		_primary_hgt_source = _params.vdist_sensor_type;
 	}
 
-	// accumulate enough height measurements to be confident in the qulaity of the data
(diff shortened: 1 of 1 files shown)


bartslinger:
Actually just realized that the entire check is superfluous, because the only four possibilities are already covered in the if-statement. What do you think @ChristophTobler ?

ChristophTobler:
>Actually just realized that the entire check is superfluous, because the only four possibilities are already covered in the if-statement.

True. Can you update the PR accordingly?

ChristophTobler:
@bartslinger Can you please resolve the conflicts?

bartslinger:
Also added `_control_status.flags.ev_hgt = false;` because otherwise you could end up in a situation where both `baro_hgt` and `ev_hgt` are set to true. I assumed that that's not desirable.

mhkabir:
@priseborough Can you review this please?

PR #432, opened 2018-05-03
cmake standalone

dagar:
This updates ECL cmake so that it's the same if built standalone or included in another project like PX4.


priseborough:
This is outside my competence to review as I do not have the necessary understanding of the make system.

dagar:
It was more of an FYI, from the PX4 workflow perspective this doesn't actually change anything.

A few notes
 - ECL now has a standard cmake build, and it's not built the same way when included in other projects (no special PX4 cmake)
 - each top level module (EKF, tecs, etc) creates a separate library, it's up to the consuming project to link and use (libraries named ecl_EKF, ecl_tecs, etc)
 - the EKF python testing is slightly more accessible now, try running `make test` and `make test_EKF` from within ecl (even as a submodule)

image

priseborough:
This has broken build for ekf2 replay preventing testing of subsequent changes. I will revert it until that issue has been solved.

priseborough:
Why was this merged without review?   It breaks a build I am working on for another project.

dagar:
Where is it broken? Those types of concerns were why I tagged you for review.

There actually isn't much risk in this PR, it's a lot of cmake to replace this single file https://github.com/PX4/ecl/blob/1bd1809d6e5c790f3eeedb1cb967d67dfc9bec54/CMakeLists.txt.
I still verified the actual symbols within PX4 binaries before and after.

priseborough:
There was less than half a day from when i got up this morning and saw the PR to when it was merged without review. I indicated i did not have the competence in the are of build systems to review it. When I got back to it this afternoon to try building it with my  project it had already been merged. Normally this sort of urgency is reserved for hot fixes.

priseborough:
I am getting multiple undefined reference errors, eg:

/Users/paul/src/Firmware/build/tap-v3_default_replay/../../src/modules/ekf2/ekf2_main.cpp:1302: undefined reference to `Ekf::get_hagl_innov_var(float*)'

I am currently working through incorporating the cmake changes in https://github.com/PX4/Firmware/pull/9406/files.

priseborough:
I will get some asistance on the project side to try and resolve the build errors @simonegu FYI

dagar:
This PR was a combination of both trivial (certain that the generated output was identical) and large (many files moved around). There was no urgency to merge it, but I'd also rather not let it sit around requiring multiple rebases.

Please let me know if and when active development is underway and I will happily hold off on large changes like this.

dagar:
@priseborough in your branch (output above) what do you have in the ekf2 CMakeLists.txt?

https://github.com/PX4/Firmware/blob/master/src/modules/ekf2/CMakeLists.txt#L43-L45

priseborough:
px4_add_module(
	MODULE modules__ekf2
	MAIN ekf2
	COMPILE_FLAGS
		-Wno-sign-compare # TODO: fix all sign-compare
	STACK_MAIN 2500
	STACK_MAX 4000
	SRCS
		ekf2_main.cpp
	DEPENDS
		platforms__common
		git_ecl
	)

dagar:
Are you able to cherry-pick this commit? https://github.com/PX4/Firmware/commit/2e92484325ee99ea7538a39e2f3fa623e8d974ab

image

You're missing the ecl_EKF and ecl_geo libraries as dependencies (DEPENDS).

priseborough:
Not without generating conflicts in almost every file. I need to work through them with the project developers.

priseborough:
screen shot 2018-05-04 at 4 58 43 pm

dagar:
The fact that it didn't cherry-pick cleanly makes me think there must be some other large changes in the way.

Without the cherry-picked commit all you need to do is add the appropriate library to the DEPENDS section of px4_add_module if you get an undefined reference when building.

The libraries (ecl_EKF, ecl_geo, ecl_tecs) correspond to the top level modules/folders in the ECL project.

For example...
  - ekf2 uses ecl EKF and geo, so under DEPENDS it has ecl_EKF and ecl_geo
  - fw_pos_ctrl_l1 uses tecs and l1, under DEPENDS it needs ecl_tecs and ecl_l1

PR #435, opened 2018-05-04
Revert "cmake standalone"

priseborough:
Reverts PX4/ecl#432

Requested because PX4/ecl#432 breaks builds for projects forked prior to https://github.com/PX4/Firmware/commit/2e92484325ee99ea7538a39e2f3fa623e8d974ab that wish to use latest ecl and has not been peer reviewed.


dagar:
PX4/Firmware master ECL was updated ~ 16 hours ago to the latest PX4/ecl master commit before the cmake changes, so projects forked prior to PX4/Firmware@2e92484 aren't stuck.

image

Is there something broken that can't be fixed with updating both Firmware and ecl?

dagar:
Closing the PR, but keeping the branch for reference.

PR #335, opened 2017-10-13
EKF: Enable operation where mag heading is unreliable

priseborough:
This PR adds a parameter controlled option that enables operation in environments where the earths magnetic field is either distorted or has insufficient horizontal strength to provide a reliable yaw estimate.

It has been tested using  https://github.com/priseborough/Firmware/tree/ekfHighLatMag-wip 

and 

'make posix_sitl_default gazebo_standard_vtol'

using EKF2_MAGEF_BAD = 0, 1 or 2 and EKF2_MAG_YAWGND = 90.0 to match the simulation initial yaw.

SITL TEST RESULTS

https://logs.px4.io/plot_app?log=399b212a-0a3e-4d8c-b0a9-5256adb2678a
EKF2_MAGEF_BAD = 0 : Normal mag heading fusion in the ground with a mixture of heading and 3D fusion in the air.
screen shot 2017-10-13 at 12 31 12 pm

https://logs.px4.io/plot_app?log=be4e651e-00dc-45f2-af9e-c89ee91ddf03
EKF2_MAGEF_BAD = 1 : Static heading fusion in the ground with no fusion of mag data in air.
screen shot 2017-10-13 at 12 31 26 pm

https://logs.px4.io/plot_app?log=a13c0c5f-9278-4900-b888-d2e1d0a56ed2
EKF2_MAGEF_BAD = 2 : Static heading fusion in the ground. 3-axis mag fusion is used when selected, mag heading fusion is blocked.
screen shot 2017-10-13 at 12 31 38 pm

PR #445, opened 2018-05-11
EKF: Enable operation where mag heading is unreliable

mhkabir:
Rebased version of https://github.com/PX4/ecl/pull/335.


LorenzMeier:
@priseborough Good to go?

priseborough:
We need to do some testing. I didn't have tine to do any last week.

priseborough:
It failed to commence using GPS on my limited testing this afternoon. Test used https://github.com/priseborough/Firmware/tree/ekfHighLatMag-wip with EKF2_MAG_BAD=1 and EKF2_MAG_YAWGND=90

I will investigate further.

priseborough:
I've rebased and using latest PX4/Firmware master was able to arm in POSCTL. Will test further tomorrow. 

I think it makes sense to try and integrate this better into the logic that has been added to enable a non-magnetomer mode for indoor use which is selected using the EKF2_MAG_TYPE parameter.

The EKF2_MAGEF_BAD parameter added by this PR is going to be confusing, so I will try to add it as another option selected by EKF2_MAG_TYPE.

Also the movement check used on ground when EKF2_MAG_TYPE = 4 which enables the vehicle to be picked up and moved can also be used to avoid the restriction imposed by this PR which requires the vehicle to be not rotated in yaw when on ground.

priseborough:
Closing this until it is reworked to integrate with the _mag_use_inhibit flag and EKF2_MAG_TYPE = 4 behaviour.

PR #448, opened 2018-05-19
EKF : introduce new architechture for navigation limits

mhkabir:
This reworks the navigation limits architecture to be more flexible in terms of sensing sources, and does not require the estimator to access sensor-specific parameters. Instead, the ekf2 frontend gets it from the sensor drivers directly, allowing for more advanced limiting for flow and visual odometry e.g based on scene texture.

PX4 branch : https://github.com/PX4/Firmware/commits/pr-optFlowFixes
Testing : Untested, WIP


priseborough:
@mhkabir when/how do you plan to test this?

mhkabir:
This commit is already in https://github.com/PX4/ecl/pull/452. It has been tested in SITL and in real flights as a part of the Firmware PR.

I'm closing this standalone PR, since it will be going in with #452.

PR #449, opened 2018-05-21
geo remove use of DBL_EPSILON

dagar:
-this doesn't seem to be (obviously?) available on qurt

@RomanBapst FYI


RomanBapst:
@dagar Man this was actually an easy, bummer we did not find it.

PR #450, opened 2018-05-22
fix negative vision timeout

EliaTarasov:
This pr fixes negative vision timeout. This can happen if last vision timestamp is greater then last imu timestamp and it would trigger false ev timeout.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -326,6 +326,7 @@ void Ekf::controlExternalVisionFusion()
 		}
 
 	} else if (_control_status.flags.ev_pos
+		   && (_time_last_imu >= _time_last_ext_vision)
 		   && (_time_last_imu - _time_last_ext_vision > (uint64_t)_params.no_gps_timeout_max)) {
 
 		// Turn off EV fusion mode if no data has been received


mhkabir:
The correct fix would be to make sure you can't have future timestamps in the system. The new timesync implementation on PX4 master makes sure of this. Have you tried it?

EliaTarasov:
I just tested latest master and still got the same error.

mhkabir:
How are you time synchronising?

EliaTarasov:
Nohow. I thought i was somehow done by mavlink, but i didn't find and didn't specify any particular option to force mavlink or ekf2 to do that. Can you give some hints on it?

LorenzMeier:
I think hardening is always important and needed.

mhkabir:
Hardening is important, but this might just hide a much bigger problem with the system design and the user would never find out.

@EliaTarasov If you are using mavros with PX4, time synchronisation is done automatically. It takes about 50s to converge and should then give timestamps accurate to 50 microseconds. Can you send a ULog so that I can see what's going on?

EKF/control.cpp
```diff
@@ -326,7 +326,7 @@ void Ekf::controlExternalVisionFusion()
 		}
 
 	} else if (_control_status.flags.ev_pos
-		   && (_time_last_imu - _time_last_ext_vision > (uint64_t)_params.no_gps_timeout_max)) {
+		   && ((uint64_t)(std::llabs(_time_last_imu - _time_last_ext_vision)) > (uint64_t)_params.no_gps_timeout_max)) {
```
dagar: I'm not sure if this does what you're intending. These are unsigned types. Do you also want to treat a sufficiently large future timestamp as a timeout as well?

What about a simple check that `_time_last_imu >= _time_last_ext_vision`?

EliaTarasov:
@dagar 

> What about a simple check that _time_last_imu >= _time_last_ext_vision?

That's look better.

EliaTarasov:
@mhkabir 

> I think that it will be most prudent to make sure that samples with future timestamps don't make it into the EKF. This can probably be done effectively in the Firmware ekf2 wrapper.

For the long-term solution i agree. Close this pr then now?

mhkabir:
This looks good. I still need to check why the timesync hard-resets in SITL.

EliaTarasov:
@mhkabir Is there any news on this issue?

PR #452, opened 2018-05-24
Miscellaneous optical flow use improvements

priseborough:
Changes to make use of optical flow more reliable in an indoor environment where GPS signal and magnetic field are inconsistent.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -340,6 +340,9 @@ struct parameters {
 	// auxilliary velocity fusion
 	float auxvel_noise{0.5f};		///< minimum observation noise, uses reported noise if greater (m/s)
 	float auxvel_gate{5.0f};		///< velocity fusion innovation consistency gate size (STD)
+
+	// control of on-ground movement check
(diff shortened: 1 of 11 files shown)


mhkabir:
This is awesome! I had started working on a similar feature locally to support VIO/OF with indoor detection based on GPS quality checks.

mhkabir:
Does this PR fuse magnetic heading while on ground without GPS?

priseborough:
It there are no body frame observations (GPS or external vision), then indoor operation is assumed and magnetometer is not used because it is not required in that mode of operation. The reasoning is that the large magnetometer innovations generated flying indoors can also cause significant height and other state errors due to the linearisation assumptions inherent in the covariance matrix. 

I am thinking about enabling that functionality being a parameter choice because there may be some applications where we are not navigating in NED, but where a magnetic referenced yaw would be useful.

mhkabir:
That would be really good to have. Can you add something? I would generally like it in order to have a mostly-correct magnetic heading when taking off in GPS shadowed areas.

dagar:
PX4/Firmware master is currently pointing at this branch.

PR #453, opened 2018-05-24
Pr bitcraze optical flow ekf interface

DanielePettenuzzo:
This PR enables optical flow on crazyflie. The optical flow deck doesn't have a gyroscope therefore the gyro data should be taken from the imu data in the estimator interface. The pmw3901 driver (optical flow for crazyflie) will publish gyro data equal to NAN to make this check in the ekf.

Had to add a read function in the ringbuffer to read imu data without deleting any older data from it.

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -158,6 +158,29 @@ public:
 		return false;
 	}
 
+	bool read_first_older_than(const uint64_t& timestamp, data_type *sample)
+	{
+		// start looking from newest observation data
+		for (uint8_t i = 0; i < _size; i++) {
+			int index = (_head - i);
+			index = index < 0 ? _size + index : index;
+
+			if (timestamp >= _buffer[index].time_us && timestamp - _buffer[index].time_us < (uint64_t)1e5) {
+
+				*sample = _buffer[index];
+
+				return true;
+			}
+
+			if (index == _tail) {
+				// we have reached the tail and haven't got a match
+				return false;
+			}
+		}
+
+		return false;
+	}
+
 	int get_total_size() { return sizeof(*this) + sizeof(data_type) * _size; }
 
 private:
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -391,21 +391,35 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
 
 			// NOTE: the EKF uses the reverse sign convention to the flow sensor. EKF assumes positive LOS rate is produced by a RH rotation of the image about the sensor axis.
 			// copy the optical and gyro measured delta angles
-			optflow_sample_new.gyroXYZ = - flow->gyrodata;
+			//optflow_sample_new.gyroXYZ = - flow->gyrodata;
+
+			imuSample matching_imu_sample;
+
+			_imu_buffer.read_first_older_than(optflow_sample_new.time_us, &matching_imu_sample);
+
+			Vector3f matching_gyro_sample = matching_imu_sample.delta_ang / matching_imu_sample.delta_ang_dt;
+
+			optflow_sample_new.gyroXYZ = matching_gyro_sample;
+
+			flow_quality_good = true;
 
 			if (flow_quality_good) {
-				optflow_sample_new.flowRadXY = - flow->flowdata;
+				//optflow_sample_new.flowRadXY = - flow->flowdata;
+				optflow_sample_new.flowRadXY = flow->flowdata / delta_time;
 
 			} else {
 				// when on the ground with poor flow quality, assume zero ground relative velocity
-				optflow_sample_new.flowRadXY(0) = - flow->gyrodata(0);
-				optflow_sample_new.flowRadXY(1) = - flow->gyrodata(1);
+				//optflow_sample_new.flowRadXY(0) = - flow->gyrodata(0);
+				//optflow_sample_new.flowRadXY(1) = - flow->gyrodata(1);
+
+				optflow_sample_new.flowRadXY(0) = - matching_gyro_sample(0);
+				optflow_sample_new.flowRadXY(1) = - matching_gyro_sample(1);
 
 			}
 
 			// compensate for body motion to give a LOS rate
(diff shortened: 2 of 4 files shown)


EKF/estimator_interface.cpp
```diff
@@ -391,21 +391,56 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
...
+
+			bool no_gyro = false;
+			imuSample matching_imu_sample;
+			Vector3f matching_gyro_sample;
+
+			/* for optical flow boards with no gyro. The driver will set the gyro x, y and z fields in the optical_flow message to NAN so that  
+			 * we can check this here and get the gyro data with the matching timestamp from the imu buffer. The gyro data is divided by dt to 
+			 * convert it into a rate before removing it from the flow to remove the rotation component. This is done to prevent errors when the flow
+			 * and gyro data are integrated over different times */
+			if (!ISFINITE(flow->gyrodata(0)) && !ISFINITE(flow->gyrodata(1)) && !ISFINITE(flow->gyrodata(2))) {
```
dagar: Would it be worth doing OR here to catch any bogus output in the regular optical flow case?

EKF/estimator_interface.cpp
```diff
@@ -391,21 +391,56 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
...
+				no_gyro = true;
+				_imu_buffer.read_first_older_than(optflow_sample_new.time_us, &matching_imu_sample);
+				matching_gyro_sample = matching_imu_sample.delta_ang / matching_imu_sample.delta_ang_dt;
+				optflow_sample_new.gyroXYZ = matching_gyro_sample;
```
dagar: Why not populate optflow_sample_new so that the remaining logic in this function is identical with or without gyro?

EKF/estimator_interface.cpp
priseborough: @dagar can you please explain the specific error condition you would like to see addressed here.

EKF/RingBuffer.h
```diff
@@ -158,6 +158,29 @@ class RingBuffer
 		return false;
 	}
 
+	bool read_first_older_than(const uint64_t& timestamp, data_type *sample)
```
priseborough: How have you verified that this new function is retrieving the correct data? The system level effect of retrieving data at the wrong time will be subtle and could easily be missed.

EKF/estimator_interface.cpp
```diff
@@ -404,21 +404,56 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
...
+			imuSample matching_imu_sample;
+			Vector3f matching_gyro_sample;
+
+			/* for optical flow boards with no gyro. The driver will set the gyro x, y and z fields in the optical_flow message to NAN so that  
+			 * we can check this here and get the gyro data with the matching timestamp from the imu buffer. The gyro data is divided by dt to 
+			 * convert it into a rate before removing it from the flow to remove the rotation component. This is done to prevent errors when the flow
+			 * and gyro data are integrated over different times */
+			if (!ISFINITE(flow->gyrodata(0)) && !ISFINITE(flow->gyrodata(1)) && !ISFINITE(flow->gyrodata(2))) {
+				no_gyro = true;
+				_imu_buffer.read_first_older_than(optflow_sample_new.time_us, &matching_imu_sample);
```
priseborough: This introduces point sampling of  gyro data and with flow sensors that output at a lower rate, eg 10Hz which is undesirable as it can introduce significant errors due to aliasing of body angular motioin. If we are going to do the reconstruction of flow->gyrodata inside the EKF, then it should be across the integration time interval of the flow sensor. 

That means we need a method of sampling and accumulating the gyro data between optical flow measurements.

priseborough: I will have a look to see how we can achieve this cleanly using the high rate IMU data. IMU data  is accumulated and downsampled to 8msec before prior to being written to the EKF's internal buffer, so we will need to run the accumulation for optical flow as a separate function called from inside the setIMUData function before the downsampling.

EKF/estimator_interface.cpp
dagar: The current logic switches to using gyro from the EKF if all 3 flow->gyrodata fields are not finite. I was asking about switching it to ANY flow->gyrodata fields being non-finite, which would prevent bogus data from entering in regular usage without additional checks.

priseborough: OK, that makes sense

EKF/estimator_interface.cpp
```diff
@@ -404,19 +404,31 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
...
+			imuSample matching_imu_sample;
+			Vector3f matching_gyro_sample;
+
+			/* for optical flow boards with no gyro. The driver will set the gyro x, y and z fields in the optical_flow message to NAN so that  
+			 * we can check this here and get the gyro data with the matching timestamp from the imu buffer. The gyro data is divided by dt to 
+			 * convert it into a rate before removing it from the flow to remove the rotation component. This is done to prevent errors when the flow
+			 * and gyro data are integrated over different times */
+			if (!ISFINITE(flow->gyrodata(0)) && !ISFINITE(flow->gyrodata(1)) && !ISFINITE(flow->gyrodata(2))) {
+				_imu_buffer.read_first_older_than(optflow_sample_new.time_us, &matching_imu_sample);
+				flow->gyrodata = matching_imu_sample.delta_ang / matching_imu_sample.delta_ang_dt * delta_time;
```
priseborough: This introduces point sampling of gyro data which is going to be problematic with lower frequency optical flow data, eg 10Hz due to aliasing of body motion. A different approach is going to be needed. We already have _imu_del_ang_of and _delta_time_of class variables which are the accumulated gyro data between the current and previous optical flow observation. I think we should move the motion compensation into controlOpticalFlowFusion() and use those variables

priseborough:
See https://github.com/PX4/ecl/pull/458 which eliminates the problem with point sampling of gyro data.

PR #454, opened 2018-05-27
Pr no gyro opt flow

RomanBapst:
@DanielePettenuzzo I tried to change the logic so that we can remove multiple checks for no_gyro flag.
Basically I think if we figure out that we have no gyro data from the flow sensor we can get a gyro sample from the buffer, convert it into a gyro rate, and then back to a delta angle using the dt from the flow sensor.
I hope I did not oversee something here.
@priseborough FYI, I corrected the units in the comment for the flow message (common.h)

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -71,8 +71,8 @@ struct gps_message {
 
 struct flow_message {
 	uint8_t quality;	///< Quality of Flow data
-	Vector2f flowdata;	///< Optical flow rates about the X and Y body axes (rad/sec)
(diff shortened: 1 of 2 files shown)


DanielePettenuzzo:
@RomanBapst Everything looks good. We can test it this week. Can you just remove the no_gyro variable so that it builds properly?

priseborough:
I have no issue with these changes, but  after further review of https://github.com/PX4/ecl/pull/453 I think we need a different approach to sampling the IMU rate data. The method used by  https://github.com/PX4/ecl/pull/453 introduces point sampling which is undesirable when using low frequency optical flow rates.

PR #455, opened 2018-05-27
set MODULE define for each library

dagar:
- this is needed for INFO, WARN, ERR messages when used within PX4

image


dagar:
Should be `MODULE_NAME`, not `MODULE`.

dagar:
image

LorenzMeier:
Makes sense!

PR #458, opened 2018-06-12
EKF: Enable use of flow sensors that do not have gyros

priseborough:
Replaces https://github.com/PX4/ecl/pull/453

If optical flow gyro delta angles are set to NAN, then the EKF's internal IMU sample will be used.

TODO testing

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -141,7 +141,7 @@ struct flowSample {
 	Vector2f flowRadXY;	///< measured delta angle of the image about the X and Y body axes (rad), RH rotaton is positive
 	Vector3f gyroXYZ;	///< measured delta angle of the inertial frame about the body axes obtained from rate gyro measurements (rad), RH rotation is positive
 	float    dt;		///< amount of integration time (sec)
-	uint64_t time_us;	///< timestamp of the integration period mid-point (uSec)
+	uint64_t time_us;	///< timestamp of the integration period leading edge (uSec)
 };
 
 struct extVisionSample {
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -139,7 +139,6 @@ struct airspeedSample {
 struct flowSample {
 	uint8_t  quality;	///< quality indicator between 0 and 255
 	Vector2f flowRadXY;	///< measured delta angle of the image about the X and Y body axes (rad), RH rotaton is positive
-	Vector2f flowRadXYcomp;	///< measured delta angle of the image about the X and Y body axes after removal of body rotation (rad), RH rotation is positive
 	Vector3f gyroXYZ;	///< measured delta angle of the inertial frame about the body axes obtained from rate gyro measurements (rad), RH rotation is positive
 	float    dt;		///< amount of integration time (sec)
 	uint64_t time_us;	///< timestamp of the integration period mid-point (uSec)
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -455,16 +455,28 @@ void Ekf::controlOpticalFlowFusion()
 			}
 		}
 
-		// fuse the data if the terrain/distance to bottom is valid but use a more relaxed check to enable it to survive bad range finder data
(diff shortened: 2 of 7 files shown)


LorenzMeier:
Firmware PR is merged, so you should be able to consolidate this all.

mhkabir:
I did not look deeply into the diff, but does this take into account the delay of the optical flow measurement? For different sensors this could be much higher, upto 40-50ms.

priseborough:
That is handled intrinsically because the accumulated gyro and optical flow measurements are brought together at the fusion time horizon

priseborough:
I've done another review pass on the exisiting sensor buffering code and I think we still have a half optical flow frame time offset mismatch between the flow data and the gyro data because the time stamp for the flow data is set to the middle of the integration time period, not the end. 

This means the gyro data that is used to do the accumulation for the _imu_del_ang_of class variable will need to retrieved from the buffer separately using a time stamp that is the same as the end of the optical flow integration period instead of the midpoint. This won't take long to do.

priseborough:
I've pushed a commit which should fix the difference in definition of the sampling time used in the optical flow buffer vs the sampling time required to accumulated gyro data. 

It also fixes  a pre-exisiting bug that meant the gyro data used to calculate the flow sensor gyro bias corrections was only accumulating EKF gyro data when new flow data arrived.

 I will go through it again tomorrow to see if I can spot anything else that has been missed.

RomanBapst:
@priseborough Can you please have a look at the commit I added? I think the sign of the gyro data when it was taken from the ekf was not correct. I tested in SITL and also on the crazyfly.
On the crazyfly I additionally had the issue that the flow data seemed to have signs inverted. However, I think this may be a problem in the optical flow driver used for the crazyfly which I need to have a look at with @DanielePettenuzzo 
@DanielePettenuzzo  We also need to add the newly introduced flow constraints to the driver, otherwise the ekf will not accept the flow data. I will do a pull request for that tomorrow.

priseborough:
@RomanBapst Thanks for spotting that.

On the subject of flow constraints, are you talking about the maximum flow rate? If so we need to find the value at which the sensor stops being accurate and set the associated EKF2 parameter appropriately in the crazyflie config file, rather than applying a flow limit in the sensor.

RomanBapst:
@priseborough Ok, I'll have a look and update those parameters.

mhkabir:
These are the parameters your crazyflie driver needs to read in and publish to uORB : https://github.com/PX4/Firmware/blob/master/src/modules/sensors/sensor_params_flow.c

priseborough:
@RomanBapst and @mhkabir , can one of you please repeat your review. The last one you did has been dismissed by recent changes.

RomanBapst:
@priseborough Here is a log from a flight with optical flow. Since we did not get the crazyzfly to work we decided to put the bitcraze optical flow deck on the S500 frame (similar to F450). The tests went very well, very stable flight and low drift. Unfortunately, we were constrained to fly at low altitude due to the teraranger's inability to measure in excess of 1.5 m.
https://logs.px4.io/plot_app?log=3fd36665-7441-4d21-b2a8-dc1e499f17f2

priseborough:
Thanks Roman, the logs looks good. I like the idea of using that sensor on a a different airframe for testing - do you have a description of how that was wired up?

RomanBapst:
@priseborough We were using a prototype version of the board that has a different pinout.
I found this on the web for the breakout board
expansion_board_pinout

The flow sensor is connected via SPI so the relevant pins would be
VCC, SCK, MISO, MOSI and GND

Here are the schematics
https://wiki.bitcraze.io/_media/projects:crazyflie2:expansionboards:flow-deck_reve.pdf

We will start documenting this properly as we plan to use those widely.

RomanBapst:
@priseborough I forgot one pin, you need to also connect I03 pin which is the chip select pin (SS).

Issue #459, opened 2018-06-13
Timestamp comparison issue


bkueng:
[here](https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L615) we have the check:
```
_time_last_imu - _time_last_gps > (uint64_t)10e6
```
And from `ekf2_main` the gps data is copied and set via `setGpsData`, after `setIMUData` is called. It means that `_time_last_gps > _time_last_imu` can happen. And the comparison is unsigned, so the check will result true.

This happens in SITL for example, the timestamp difference is only a few microseconds.
I see more of these checks, so the problem exists elsewhere too.

Possible solutions:
- convert all checks to use signed logic
- convert all checks like this: `_time_last_imu > _time_last_gps + (uint64_t)10e6`
- call `setIMUData` after all other data is already set. This might have other side-effects
- make sure `_time_last_imu` is always the last timestamp. We would then set the gps timestamp like this:
  ```
  _time_last_gps = min(_time_last_imu, time_usec)
  ```

The last one is the simplest solution with probably the least side-effects.

@priseborough what do you think?

mhkabir:
Also this was merged without much discussion https://github.com/PX4/ecl/pull/450, but I don't have clarity on what it fixes/attempts to fix. Looks like a similar issue.

LorenzMeier:
The last one looks good.

priseborough:
This check requires GPS use to be true which means previous use of GPS. This can only happen if GPS checks have been passed which requires 10 seconds of valid GPS data retrieved from the buffer.  Also data is only retrieved from the buffer after it has fallen behind the fusion time horizon which is delayed relative to _time_last_imu by the length of the IMU buffer.

priseborough:
There are three timestamp error conditions that will affect the operation of the EKF:

1) The data timestamp is greater than than _imu_sample_delayed.time_us + 1000 * FILTER_UPDATE_PERIOD_MS. If this happens data will be fused at the wrong time because the fusion time horizon will have already passed before the EKF has opportunity to retrieve the data from the buffer.

2) Data is out of order, This will cause data to be fused at the incorrect time because the buffers require the timestamps to be always increasing.

3) The time stamp is small enough or the arrival data rate is high enough such that the data has been overwritten with new data before it can be retrieved from the buffer.

priseborough:
Fixed by https://github.com/PX4/ecl/pull/462

Will keep this open until the other instances, eg range finder have been fixed.

[...]

PR #461, opened 2018-06-14
fix timestamp comparison issue

bkueng:
There are checks in the code such as this:
```
 _time_last_imu - _time_last_gps > (uint64_t)10e6
```
If the gps time is newer (by a couple of microseconds), the above check will evaluate to true due to unsigned wrap-around logic.

Ensuring at the interface level that no other timestamp > _time_last_imu is fed into the ecl library fixes this.

However this could potentially hide timestamp problems.
For this reason, and since there was already a fix in https://github.com/PX4/ecl/pull/450, the check is not added to the external vision data.

This PR is a solution for #459 that can be integrated into the release. We can still discuss & implement a different solution later.

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -173,6 +173,12 @@ void EstimatorInterface::setMagData(uint64_t time_usec, float (&data)[3])
 		return;
 	}
 
+	// There are various code places that assume _time_last_imu is the latest timestamp,
+	// so make sure the mag time is not newer.
+	if (time_usec > _time_last_imu) {
+		time_usec = _time_last_imu;
+	}
+
 	// Allocate the required buffer size if not previously done
 	// Do not retry if allocation has failed previously
 	if (_mag_buffer.get_length() < _obs_buffer_length) {
@@ -203,6 +209,12 @@ void EstimatorInterface::setGpsData(uint64_t time_usec, struct gps_message *gps)
 {
 	if (!_initialised || _gps_buffer_fail) {
 		return;
+	}
+
+	// There are various code places that assume _time_last_imu is the latest timestamp,
+	// so make sure the gps time is not newer.
+	if (time_usec > _time_last_imu) {
+		time_usec = _time_last_imu;
 	}
 
 	// Allocate the required buffer size if not previously done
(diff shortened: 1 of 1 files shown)


mhkabir:
@bkueng can you revert the fix from #450 and apply this fix to vision too? This is just to stay consistent.

dagar:
Why don't we just switch to signed 64 bit integers across the board? That's still plenty of time in microseconds (or nanoseconds for that matter). We could also get consistent with typedef usage.

LorenzMeier:
@dagar Because that doesn’t guarantee that horrible things won’t happen with negative timestamps.

priseborough:
I do not agree with this change because if an off-bard IMU data and a low latency sensor is being used, then the time stamp of the aiding data my be newer than the current IMU time. This is is not a problem provided the time difference does not exceed the length of the IMU data buffer.

It must be remembered that internally the EKF will only retrieve observation data when it's time stamp has fallen behind the fusion time horizon as indicated by _imu_sample_delayed.time_us, which is delayed relative to _time_last_imu.

priseborough:
I think the way to fix the loss of data check issue is to use change the check to use the time stamps for data that has been been retrieved from the buffer, not the most recent data that has arrived at the interface. We then rely on

_gps_data_ready = _gps_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_gps_sample_delayed);

to ensure that _gps_sample_delayed.time_us never greater than _imu_sample_delayed.time_us

LorenzMeier:
@priseborough If you feel that's better and a 100% fix, could you add commits to this PR or put up a new PR? We really would need to get this in for v1.8 with *minimal changes*. Thanks!

priseborough:
If we wanted to add additional protection against outrageous time stamps, then in the interface we can reject data for any corrected time stamp <  _imu_sample_delayed.time_us - FILTER_UPDATE_PERIOD_MS * (_imu_buffer_length - 1) because that data is already behind the fusion time horizon and cannot be time aligned.

priseborough:
@LorenzMeier  OK - will submit a minimum change PR

priseborough:
Replaced by https://github.com/PX4/ecl/pull/462

dagar:
> @dagar Because that doesn’t guarantee that horrible things won’t happen with negative timestamps.

Yes we still need additional checks preventing bogus data from entering the EKF in the first place. I'm talking about avoiding all kinds of relatively subtle issues (like this one) with integer overflow and signed conversions that are often ignored. Somewhat surprisingly signed integers also end up being faster in many cases. https://www.youtube.com/watch?v=equbUrX-ZWQ

It would fix all of the comparisons like ` _time_last_imu - _time_last_gps > 10e6`.

PR #464, opened 2018-06-23
EKF : allow init without external vision measurements

mhkabir:
This allows initialization of the attitude estimate without receiving vision messages first. It also automatically resets the calculated rotation matrix from navigation frame to vision frame in the event of a vision (re)-init, since the visual estimator will have a new origin whenever it resets.

This is a first in a series of PRs which will be robustifying vision-GPS-OF flight over the next few weeks.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -171,6 +171,11 @@ void Ekf::controlExternalVisionFusion()
 				_control_status.flags.ev_pos = true;
 				ECL_INFO("EKF commencing external vision position fusion");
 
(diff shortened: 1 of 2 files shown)


mhkabir:
@priseborough Does this look good to you? I flew this a few times, seems to work fine.

mhkabir:
Sorry, will do tomorrow.

mhkabir:
@priseborough Test log : https://logs.px4.io/plot_app?log=84e5cd4e-e5f8-435c-9938-1ac77e924529

PR #465, opened 2018-06-25
EKF angle constants in degrees for readability

dagar:
Minor readability change. These can all be evaluated at compile time (math::radians and sq), so there's no functional or even minor performance change.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -516,8 +516,7 @@ void Ekf::controlGpsFusion()
 					if (_mag_inhibit_yaw_reset_req) {
 						_mag_inhibit_yaw_reset_req = false;
 						// Zero the yaw bias covariance and set the variance to the initial alignment uncertainty
-						float dt = 0.001f * (float)FILTER_UPDATE_PERIOD_MS;
-						setDiag(P, 12, 12, sq(_params.switch_on_gyro_bias * dt));
+						setDiag(P, 12, 12, sq(_params.switch_on_gyro_bias * FILTER_UPDATE_PERIOD_S));
 					}
 				}
 
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -58,7 +58,7 @@ void Ekf::controlFusionModes()
 
 		// Once the tilt variances have reduced to equivalent of 3deg uncertainty, re-set the yaw and magnetic field states
 		// and declare the tilt alignment complete
-		if ((angle_err_var_vec(0) + angle_err_var_vec(1)) < sq(0.05235f)) {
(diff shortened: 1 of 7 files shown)


EKF/ekf.h
```diff
@@ -598,10 +598,7 @@ class Ekf : public EstimatorInterface
 	void checkForStuckRange();
 
 	// return the square of two floating point numbers - used in auto coded sections
-	inline float sq(float var)
-	{
-		return var * var;
-	}
+	static constexpr float sq(float var) { return var * var; }
```
priseborough: Why has the inline attrribute been removed function? inlining has been used in the past in the past to preserve speed of autocode. What effect will this change have on the compiled code? My intention was that the compiled code usign this function would be the same as if I had used var*var instead.

dagar: Adding constexpr implies inline, but also means it can be fully evaluated at compile time (if the input is known at compile time).

The mathlib helper functions for degrees <-> radians, constrain, min, max are all constexpr as well. Overall that means statements like `sq(math::radians(10.0f))` really do turn into `0.0304617f` at compile time. In simple cases this might happen if they were all inline as well, but it's dependant on compiler optimization level, function size, etc.

EKF/control.cpp
```diff
@@ -346,11 +346,13 @@ void Ekf::controlOpticalFlowFusion()
...
 		// When on ground check if the vehicle is being shaken or moved in a way that could cause a loss of navigation
-		float accel_norm = _accel_vec_filt.norm();
-		bool motion_is_excessive = ((accel_norm > 14.7f) // accel greater than 1.5g
-					    || (accel_norm < 4.9f) // accel less than 0.5g
+		const float accel_norm = _accel_vec_filt.norm();
+
+		const bool motion_is_excessive = ((accel_norm > (CONSTANTS_ONE_G * 1.5f)) // accel greater than 1.5g
+					    || (accel_norm < (CONSTANTS_ONE_G * 0.5f)) // accel less than 0.5g
 					    || (_ang_rate_mag_filt > _flow_max_rate) // angular rate exceeds flow sensor limit
-					    || (_R_to_earth(2,2) < 0.866f)); // tilted more than 30 degrees
```
dagar: Note the comment didn't match the code here.

priseborough: The code was correct, comment was wrong. The value had been updated during development and the comment got left behind. I will push a commit to fix it and remove the numerical values from the other check comments. The code makes the limit used clear.

priseborough: Also it is the cosine of 30 degrees, so the way it has been done will not work.

PR #466, opened 2018-06-28
terrain_estimator: add vehicle_variance_scaler

CarlOlsson:
_params.vehicle_variance_scaler was never used


LorenzMeier:
@priseborough Could you please review?

PR #467, opened 2018-06-29
TECS: Fix undesired behaviour when integrator gain is zero

philipoe:
### Issue:
For testing purposes or also if one just does not want to use an integrator, users can set the TECS `_integrator_gain` to zero. However, this results in undesired behavior:

- If one starts with `_integrator_gain!=0` but then sets `_integrator_gain=0` during flight for testing/tuning, then `_pitch_integ_state` and `_throttle_integ_state` can potentially remain at large values, thus heavily disturbing the control loop.
- Even if one starts with `_integrator_gain=0`, a climbout (e.g. autonomous takeoff via Pixhawk autopilot) will force certain integrator state values e.g. [here](https://github.com/PX4/ecl/blob/master/tecs/tecs.cpp#L373). When the climbout is finished, the integrator will then just remain at that value, again disturbing the control loop! 

### Example / logs: 
See image below: This was a test flight with PX4 Firmware v1.8.0 (plus some non-TECS related modifications), where we performed an automated takeoff and landing with `FW_T_INTEG_GAIN = 0`. Both throttle and pitch integrator states (Pitch Ref I and uThrot,I in the plots) are != 0 during both the climbout and the following flight. It obviously does not harm during the climbout, but during the descent it does harm because even though the altitude is too high we still have uThrot=0.11 due to the integrator.

image

The flights logs are available here : https://review.px4.io/plot_app?log=5b933652-cc00-4e3c-8e06-b797cd9dfb66 . The plot above shows the first takeoff in that log file.

### Solution:
Just force the integrator state for both pitch and throttle to zero when the integrator gain is zero. See commits. I will submit a separate PR to Pixhawk which clarifies this in the description of the FW_T_INTEG_GAIN parameter.

@dagar @priseborough 
@tstastny FYI

### DIFF
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -359,9 +359,11 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 		_last_throttle_setpoint = _throttle_setpoint;
 
 		// Calculate throttle integrator state upper and lower limits with allowance for
-		// 10% throttle saturation to accommodate noise on the demand
-		float integ_state_max = (_throttle_setpoint_max - _throttle_setpoint + 0.1f);
-		float integ_state_min = (_throttle_setpoint_min - _throttle_setpoint - 0.1f);
+		// 10% throttle saturation to accommodate noise on the demand. However, if the
+		// integrator gain is zero, also force integrator state to zero, otherwise it could
+		// stay at a large constant value
+		float integ_state_max = _integrator_gain > 0.001f ? (_throttle_setpoint_max - _throttle_setpoint + 0.1f) : 0.0f;
+		float integ_state_min = _integrator_gain > 0.001f ? (_throttle_setpoint_min - _throttle_setpoint - 0.1f) : 0.0f;
 
 		// Calculate a throttle demand from the integrated total energy error
 		// This will be added to the total throttle demand to compensate for steady state errors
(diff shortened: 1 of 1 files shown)


tecs/tecs.cpp
```diff
@@ -359,9 +359,11 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
...
 
 		// Calculate throttle integrator state upper and lower limits with allowance for
-		// 10% throttle saturation to accommodate noise on the demand
-		float integ_state_max = (_throttle_setpoint_max - _throttle_setpoint + 0.1f);
-		float integ_state_min = (_throttle_setpoint_min - _throttle_setpoint - 0.1f);
+		// 10% throttle saturation to accommodate noise on the demand. However, if the
+		// integrator gain is zero, also force integrator state to zero, otherwise it could
+		// stay at a large constant value
+		float integ_state_max = _integrator_gain > 0.001f ? (_throttle_setpoint_max - _throttle_setpoint + 0.1f) : 0.0f;
+		float integ_state_min = _integrator_gain > 0.001f ? (_throttle_setpoint_min - _throttle_setpoint - 0.1f) : 0.0f;
```
dagar: What about skipping the entire block instead of calculating _throttle_integ_state and then clamping it to zero?

``` C++
if (_integrator_gain > 0.0f) {
    lines ~361-380
} else {
    _throttle_integ_state = 0.0f;
}
```

tecs/tecs.cpp
```diff
@@ -473,8 +475,9 @@ void TECS::_update_pitch_setpoint()
 					max((_pitch_setpoint_min - _pitch_setpoint_unc) * climb_angle_to_SEB_rate / _pitch_time_constant, 0.0f));
 	}
 
-	// Update the pitch integrator state
+	// Update the pitch integrator state. If gain is zero, force state to zero, too, to avoid that it stays at large constant values
 	_pitch_integ_state = _pitch_integ_state + pitch_integ_input * _dt;
+	if(_integrator_gain < 0.001f) _pitch_integ_state = 0.0f;
```
dagar: Same as above. If you already know you're going to zero something you might as well save the effort of calculating it.

``` C++
if (_integrator_gain > 0.0f) {
    _pitch_integ_state = _pitch_integ_state + pitch_integ_input * _dt;
} else {
    _pitch_integ_state = 0.0f;
}
```

philipoe:
@dagar Changed and re-tested in HIL.

dagar:
I pushed a minor style fix.

PR #470, opened 2018-06-30
geo get_bearing_to_next_waypoint() small optimization

dagar:
Saves a few double floating point operations.


dagar:
Each double precision cos or sin on a typical STM32F4 flight controller is nearly 20 microseconds. The inverses are even slower.

PR #471, opened 2018-07-03
ekf run output predictor immediately with new IMU data

dagar:


LorenzMeier:
Nice!

priseborough:
@LorenzMeier This arose from a a discussion between myself and @dagar  yesterday and will help clear the path to use of multiple EKF instances without a significant time delay penalty.

LorenzMeier:
@priseborough I thought so. It was too directed to not be based on your guidance 8)

priseborough:
Here is the replayed output tracking error using the  ecl version from master:

screen shot 2018-07-04 at 11 36 05 pm

Here it is using the version from this PR:

screen shot 2018-07-04 at 11 34 22 pm

priseborough:
A general observation is that there is a lot of change in this PR when the only functional addition required for immediate update of the output attitude quaternion is addition of the calculate_quaternion() function

priseborough:
Commit https://github.com/PX4/ecl/pull/471/commits/aaff047e5852761f8d15bc695c7a1ca70beeeeb1 has fixed the difference:

This PR at commit aaff047e5852761f8d15bc695c7a1ca70beeeeb1 :

screen shot 2018-07-05 at 12 12 58 am

Master:

screen shot 2018-07-05 at 12 13 59 am

dagar:
The CI test is a bit bogus and calls update() multiple times without new IMU data.

It's now failing because update has been changed to returns true or false based on actually updating or not. Previously the return was based on tilt_align, yaw_align, and finite quaternion.

mhkabir:
Does this generally reduce latency on the attitude output?

EKF/estimator_interface.h
```diff
@@ -203,6 +204,9 @@ class EstimatorInterface
 	*/
 	virtual bool reset_imu_bias() = 0;
 
+	// return true if the attitude is usable
+	bool attitude_valid() { return ISFINITE(_output_new.quat_nominal(0)) && _control_status.flags.tilt_align && _control_status.flags.yaw_align; }
```
priseborough: Including yaw align in this could be problematic because if the resetMagHeading function fails to complete due to temporary loss of EV data, then all attitude data would be lost.

priseborough: Attitude data can still be usable for stabilisation even when absolute yaw angle is not known. I will remove that condition.

priseborough:
@mhkabir yes it does if used in combination with https://github.com/PX4/Firmware/pull/9832 to take advantage of the calculate_quaternion() function which can run immediately after setIMUData(...)

PR #472, opened 2018-07-04
[WIP] Add check that the range finder data quality.

jiejiezhang:
In many cases, the ultrasonic data is abnormal, resulting in an abnormal height, so we should add ultrasonic data quality detection. It will be safer.
 @priseborough  Could you see the PR? Thanks!

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -126,6 +126,7 @@ struct baroSample {
 };
 
 struct rangeSample {
+	uint8_t		quality;  ///< range measurement quality
 	float       rng;	///< range (distance to ground) measurement (m)
 	uint64_t    time_us;	///< timestamp of the measurement (uSec)
 };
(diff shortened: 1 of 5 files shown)


xdwgood:
It is important to ensure the credibility of this quality :smiley:

priseborough:
Where is it supposed to get the range finder quality from?

priseborough:
I think we should make following changes to achieve the objective:

1) Modify the setRangeData function so that it rejects range data if distance_sensor.covariance is <=0
2) Modify the EKF to use the maximum of the distance_sensor.covariance or sensor noise parameter
3) Ensure that all range sensor drivers publish an accuracy.
4) Fix the naming convention for distance_sensor.covariance because it is incorrect. Firstly a single measurement can't have a covariance and if it is supposed to be a variance, then it should have units of m^2. My preference is that that message represents the 1-sigma error and be named distance_sensor.accuracy.

jiejiezhang:
> I think we should make following changes to achieve the objective:

Will do now as you suggestion. Thanks!!!!!!!!!!!! :+1:

PR #475, opened 2018-07-09
Ekf opt flow fixes poor flow quality

huangzuliang:
1.Disable optical flow fuse when the flow quality is poor, it will also use poor flow data in altitude mode which will cause velocity to diverge and out of control.

this issue will occurs with EKF2_OF_QMIN value of 1 and poor flow quality, tesing with EKF2_OF_QMIN value of 100 and 40, I found that the optical flow data was used in altitude mode when it is poor quality.

issue log link: https://logs.px4.io/plot_app?log=3d21e384-6bf1-4009-a605-a8d8a3063af2 (EKF2_OF_QMIN = 100)
             https://logs.px4.io/plot_app?log=13445020-c395-480d-aaf7-de21d58d33fd (EKF2_OF_QMIN = 40)
             https://logs.px4.io/plot_app?log=4546e851-5584-4449-8138-79174aebe51b (EKF2_OF_QMIN = 1)
flow-bad3

log with fixs: https://logs.px4.io/plot_app?log=3b1def27-95e5-489f-bc07-36a18df14ac4

2. Add a variable to enable or disable the check of minimum acceptable quality integer from the flow sensor on the ground to start using optical flow, because of the flow sensor may be close to the ground and invalid before takeoff.

@priseborough can you have a look? thanks, Issue:https://github.com/YUNEEC/Firmware/issues/2239

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -291,7 +291,7 @@ struct parameters {
 	float flow_noise_qual_min{0.5f};	///< observation noise for optical flow LOS rate measurements when flow sensor quality is at the minimum useable (rad/sec)
 	int32_t flow_qual_min{1};		///< minimum acceptable quality integer required once fusion of data has commenced
 	int32_t flow_qual_min_init{128};	///< minimum acceptable quality integer required to commence fusion
-	int32_t flow_qual_min_check{0};		///< enable minimum acceptable quality integer check on the ground required to commence fusion
+	int32_t flow_qual_check_on_ground{1};	///< enable minimum acceptable quality integer check on the ground required to commence fusion
 	float flow_innov_gate{3.0f};		///< optical flow fusion innovation consistency gate size (STD)
 	float flow_rate_max{2.5f};		///< maximum valid optical flow rate (rad/sec)
 
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -291,6 +291,7 @@ struct parameters {
 	float flow_noise_qual_min{0.5f};	///< observation noise for optical flow LOS rate measurements when flow sensor quality is at the minimum useable (rad/sec)
 	int32_t flow_qual_min{1};		///< minimum acceptable quality integer required once fusion of data has commenced
 	int32_t flow_qual_min_init{128};	///< minimum acceptable quality integer required to commence fusion
+	int32_t flow_qual_min_check{0};		///< enable minimum acceptable quality integer check on the ground required to commence fusion
 	float flow_innov_gate{3.0f};		///< optical flow fusion innovation consistency gate size (STD)
 	float flow_rate_max{2.5f};		///< maximum valid optical flow rate (rad/sec)
 
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -291,7 +291,7 @@ struct parameters {
 	float flow_noise_qual_min{0.5f};	///< observation noise for optical flow LOS rate measurements when flow sensor quality is at the minimum useable (rad/sec)
 	int32_t flow_qual_min{1};		///< minimum acceptable quality integer required once fusion of data has commenced
 	int32_t flow_qual_min_init{128};	///< minimum acceptable quality integer required to commence fusion
-	int32_t flow_qual_check_on_ground{1};	///< enable minimum acceptable quality integer check on the ground required to commence fusion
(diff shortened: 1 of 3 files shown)


EKF/control.cpp
```diff
@@ -394,8 +394,21 @@ void Ekf::controlOpticalFlowFusion()
...
+
 		if (!flow_quality_ok) {
-			_time_bad_flow_qual = _imu_sample_delayed.time_us;
+			if ((!_control_status.flags.in_air || (_control_status.flags.in_air && _imu_sample_delayed.time_us - _time_takeoff < (uint64_t)5E6))
+				&& _params.flow_qual_min_check) {
+				flow_quality_ok = true;
+				_time_bad_flow_qual = 0;
+			} else {
+				_time_bad_flow_qual = _imu_sample_delayed.time_us;
+			}
```
haiyangyuneec: Could you add some comment for these code for better understanding?

EKF/control.cpp
```diff
@@ -394,8 +394,21 @@ void Ekf::controlOpticalFlowFusion()
...
+		if (!_control_status.flags.in_air) {
+			_time_takeoff = 0;
+		} else if (_time_takeoff == 0) {
+			_time_takeoff = _imu_sample_delayed.time_us;
+		}
+
 		if (!flow_quality_ok) {
-			_time_bad_flow_qual = _imu_sample_delayed.time_us;
+			if ((!_control_status.flags.in_air || (_control_status.flags.in_air && _imu_sample_delayed.time_us - _time_takeoff < (uint64_t)5E6))
+				&& _params.flow_qual_min_check) {
```
haiyangyuneec: The second "_control_status.flags.in_air" is unnecessary.

EKF/control.cpp
```diff
@@ -426,7 +439,7 @@ void Ekf::controlOpticalFlowFusion()
 
 				}
 			}
-		} else if (!(_params.fusion_mode & MASK_USE_OF)) {
+		} else if (!(_params.fusion_mode & MASK_USE_OF) || ((_params.fusion_mode & MASK_USE_OF) && !flow_quality_ok)) {
```
haiyangyuneec: The second "_params.fusion_mode & MASK_USE_OF" is unnecessary.

EKF/control.cpp
huangzuliang: ok, done in commit "EKF:add some comment and simple code"

EKF/common.h
```diff
@@ -291,6 +291,7 @@ struct parameters {
 	float flow_noise_qual_min{0.5f};	///< observation noise for optical flow LOS rate measurements when flow sensor quality is at the minimum useable (rad/sec)
 	int32_t flow_qual_min{1};		///< minimum acceptable quality integer required once fusion of data has commenced
 	int32_t flow_qual_min_init{128};	///< minimum acceptable quality integer required to commence fusion
```
priseborough: I recommend renaming these two parameter variables and update their documentation to make the use clearer here and in the code:

int32_t flow_qual_min_init_air{128};	///< minimum acceptable quality integer required to commence fusion when in air
int32_t flow_qual_min_init_ground{1};	///<minimum acceptable quality integer required to commence fusion when on ground

huangzuliang: ok, but i have a question whether "flow_qual_min_init" is also used in air when _control_status.flags.opt_flow is false, so renaming flow_qual_min_init_ground is not better.

priseborough: I was confused by the pull request explanation above '2. Add a variable to enable or disable the check of minimum acceptable quality integer from the flow sensor on the ground to start using optical flow ...' 

Looking at the code this value can be used both in air and on ground so ignore the comment.

priseborough: I will read the code again to see if the use of the flow_qual_min, flow_qual_min_init and flow_qual_check_on_ground parameters is still consistent with the variables names and descriptions.

priseborough: After reading the documentation and code again, it would make it clearer if the description for the flow_qual_check_on_ground parameter was modified to read:

///< set to 1 to enable, 0 to disable application of the flow_qual_min_init threshhold check when the vehicle is on ground

I have pushed this and some other documentation changes to https://github.com/priseborough/ecl/commit/de206aeff148eebfd9ad272f75fcdef6c5e56f93

huangzuliang: okay. thanks!

huangzuliang:
@priseborough, i have added your improvements to this PR

PR #477, opened 2018-07-11
EKF: Consolidate range finder checking

priseborough:
**DESCRIPTION**

This brings all the range finder data checks (excluding innovation consistency checks) into one place and eliminates the need to perform range checking external to the library. Having checks in multiple places has previously resulted in unintended loss of functionality over time which had to be fixed by https://github.com/PX4/ecl/pull/476. This PR should reduce the likelihood of this type of bug being introduced in the future.

The hard coded optical flow tilt limit has also been  changed to use the same value as the range finder for consistency.

Variable names have been changed changed to make a clear distinction between the max/min values calculated by by the stuck range check and the max/min valid values published by the sensor.

**TESTING**

For full test coverage we would need to check:

- Data below min valid (on ground and in air)
- Data above min valid (on ground and in air)
- Data returning to valid range after > 10 seconds out of range and sticking at a constant value (on ground and in air)
- Vehicle tilted > 45 deg

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -282,7 +282,7 @@ struct parameters {
 	float max_vel_for_range_aid{1.0f};	///< maximum ground velocity for which we allow to use the range finder as height source (if range_aid == 1)
 	int32_t range_aid{0};			///< allow switching primary height source to range finder if certian conditions are met
 	float range_aid_innov_gate{1.0f}; 	///< gate size used for innovation consistency checks for range aid fusion
-	float range_cos_max_tilt{0.7071f};	///< cosine of the maximum tilt angle from the vertical that permits use of range finder data
+	float range_cos_max_tilt{0.7071f};	///< cosine of the maximum tilt angle from the vertical that permits use of range finder and flow data
 
 	// vision position fusion
 	float ev_innov_gate{5.0f};		///< vision estimator fusion innovation consistency gate size (STD)
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -107,21 +107,21 @@ void Ekf::controlFusionModes()
 	}
 
 	// calculate 2,2 element of rotation matrix from sensor frame to earth frame
+	// this is required for use of range finder and flow data
 	_R_rng_to_earth_2_2 = _R_to_earth(2, 0) * _sin_tilt_rng + _R_to_earth(2, 2) * _cos_tilt_rng;
 
-	// Get range data from buffer and check that is within limits and the vehicle is not excessively tilted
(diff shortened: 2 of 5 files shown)


priseborough:
Reversion test using replay data:

https://logs.px4.io/plot_app?log=a56f0800-6944-4f46-8171-4a7fab048923

EKF/common.h
```diff
@@ -282,7 +282,7 @@ struct parameters {
 	float max_vel_for_range_aid{1.0f};	///< maximum ground velocity for which we allow to use the range finder as height source (if range_aid == 1)
 	int32_t range_aid{0};			///< allow switching primary height source to range finder if certian conditions are met
 	float range_aid_innov_gate{1.0f}; 	///< gate size used for innovation consistency checks for range aid fusion
-	float range_cos_max_tilt{0.7071f};	///< cosine of the maximum tilt angle from the vertical that permits use of range finder data
+	float range_cos_max_tilt{0.7071f};	///< cosine of the maximum tilt angle from the vertical that permits use of range finder and flow data
```
mhkabir: It would be good to mention explicitly the value of this angle.

EKF/control.cpp
```diff
@@ -1135,33 +1135,56 @@ void Ekf::rangeAidConditionsMet()
...
-			if (_range_sample_delayed.rng > _rng_check_max_val) {
-				_rng_check_max_val = _range_sample_delayed.rng;
+			if (_range_sample_delayed.rng > _rng_stuck_max_val) {
+				_rng_stuck_max_val = _range_sample_delayed.rng;
 			}
 
-			if (_rng_check_min_val < 0.1f || _range_sample_delayed.rng < _rng_check_min_val) {
-				_rng_check_min_val = _range_sample_delayed.rng;
+			if (_rng_stuck_min_val < 0.1f || _range_sample_delayed.rng < _rng_stuck_min_val) {
+				_rng_stuck_min_val = _range_sample_delayed.rng;
```
mhkabir: Off topic for this PR, but I would prefer if the stuck range checks eventually got pulled out of the ecl EKF and into the specific driver of the device which exhibits this failure mode. Generally I am not aware of any recent Lidar sensor which still exhibits this issue.

This could be a part of the cleanup for https://github.com/PX4/Firmware/issues/9861.

EKF/ekf.h
```diff
@@ -598,7 +598,7 @@ class Ekf : public EstimatorInterface
 	void rangeAidConditionsMet();
 
 	// check for "stuck" range finder measurements when rng was not valid for certain period
```
mhkabir: Comment is outdated.

PR #480, opened 2018-07-13
Remove if statement in checkRangeDataValidity

dakejahl:
Removed `if` statement that would prevent the update of the `_rng_hgt_faulty` flag.


dakejahl:
https://review.px4.io/plot_app?log=1aa6af1e-d571-4998-b111-2afd8dc9b92c

rng_aid_status

This flew great! Because our sensor is pretty noisy, we see rapid switching in and out of range aid mode, this is actually good though because any data that makes in this far should be consider **good** and should be prioritized over baro/gps data.

Issue #478, opened 2018-07-12
EKF continues to use last range data


priseborough:
If range data stops publishing, the EKF continues to use the last data when EKF2_RNG_AID is enabled

https://review.px4.io/plot_app?log=861edb66-663c-46b2-8a60-d1935a7290b6

https://review.px4.io/plot_app?log=89526d09-77bf-4edf-9a02-7c7c34e62681

priseborough:
The practice of indicating invalid or out of range data by not publishing as seen in these logs is not recommended. The code will be updated so that loss of range data causes reversion to use of the default height sensor, but because of the delay associated with a timeout check this mechanism should not be relied on for normal switching.

dakejahl:
Thanks, we will merge #9865 on our side to use the `signal_quality` field.

priseborough:
Prototype fix here: https://github.com/PX4/ecl/tree/pr-ekfRngAidFix

Will raise PR when testing complete.

dakejahl:
Hey Paul, I merged in this commit and went testing **(Note:** I did not change the way our driver is publishing for the sake of simply testing the timeout on the fusing of the data). 
https://review.px4.io/plot_app?log=46f674ad-abbc-41a3-8e62-2a0602417479

I noticed you have a variable `RNG_MAX_INTERVAL` which is set at 200ms that is used to determine if the data has gone stale. 

```
	if ((_imu_sample_delayed.time_us - _range_sample_delayed .time_us) > 2 * RNG_MAX_INTERVAL) {
		_rng_hgt_faulty = true;
		return;
	}
```
After 400ms we should realize that we have stale data. Here are some data snapshots from this log:|
![range_data](https://user-images.githubusercontent.com/37091262/42702132-d10eba2e-8685-11e8-965f-9cf2d62e512d.png)
**Timestamps for the 3 data points**
![timestamps](https://user-images.githubusercontent.com/37091262/42702435-bd001874-8686-11e8-86d1-d9b5687f2525.png)
![range_aid_status](https://user-images.githubusercontent.com/37091262/42701946-2a711bc6-8685-11e8-879c-e63463c61c68.png)

Notice the range_aid does not actually get disabled until it gets the data point at 4meters, which is a time delta of 2.7 seconds, much higher than the expected 400ms. Thanks for the help with this!

dakejahl:
Please see [#480](https://github.com/PX4/ecl/pull/480) 

Thanks!

Issue #424, opened 2018-04-19
checkRangeDataContinuity?


CarlOlsson:
I am wondering about the background regarding the [checkRangeDataContinuity](https://github.com/PX4/ecl/blob/bd72f3c5211176351cbc0596ef69395d359be43d/EKF/terrain_estimator.cpp#L205-L205) function and its use in the [terrain valid flag](https://github.com/PX4/ecl/blob/bd72f3c5211176351cbc0596ef69395d359be43d/EKF/terrain_estimator.cpp#L179-L179).

Why is it here? My concern is the following:
The system is descending for landing with a range sensor measuring the distance to ground. Just before hitting the ground the distance sensor starts giving bad data. The measurements will be rejected by the EKF but the hagl_valid flag will still be true for 5 seconds, all good. 

However, in the exact same scenario but the range sensor breaks completely just before touchdown the hagl_valid flag will instantly go false because the _range_data_continuous flag will be false due to the sensor not giving any data.

It seems unintuitive and non consistent to handle no data so different from bad data.

PR #481, opened 2018-07-13
EKF: Fix handling of distance sensor data timeouts

priseborough:
Fixes https://github.com/PX4/ecl/issues/478

**Tests Required**

These tests could be performed using SITL with modification of the ekf2_main.cpp file to perform the necessary distance_sensor data manipulation.

1) EKF2_RNG_AID = 1 and EKF2_HGT_MODE = 0

Place vehicle in hover at height that activates range finder use. Stop publication of range data for at least 1 second before resuming and verify that baro fusion commences after 0.4 seconds and range finder fusion re-commences when data returns.

2)  EKF2_RNG_AID = 0 and EKF2_HGT_MODE = 0

Set the distance sensor out of range for over 10 seconds and then set to a constant value within range. Verify that the stuck value check prevents the data from being used.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -90,9 +90,6 @@ void Ekf::controlFusionModes()
 	const gpsSample &gps_init = _gps_buffer.get_newest();
 	_gps_hgt_faulty = !((_time_last_imu - gps_init.time_us) < 2 * GPS_MAX_INTERVAL);
 
-	const rangeSample &rng_init = _range_buffer.get_newest();
-	_rng_hgt_faulty = !((_time_last_imu - rng_init.time_us) < 2 * RNG_MAX_INTERVAL);
-
 	// check for arrival of new sensor data at the fusion time horizon
 	_gps_data_ready = _gps_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_gps_sample_delayed);
 	_mag_data_ready = _mag_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_mag_sample_delayed);
@@ -115,6 +112,13 @@ void Ekf::controlFusionModes()
 
 	if (_range_data_ready) {
 		checkRangeDataValidity();
+
+		if (_range_data_ready && !_rng_hgt_faulty) {
+			// correct the range data for position offset relative to the IMU
+			Vector3f pos_offset_body = _params.rng_pos_body - _params.imu_pos_body;
+			Vector3f pos_offset_earth = _R_to_earth * pos_offset_body;
+			_range_sample_delayed.rng += pos_offset_earth(2) / _R_rng_to_earth_2_2;
+		}
 	}
 
 	// We don't fuse flow data immediately because we have to wait for the mid integration point to fall behind the fusion time horizon.
@@ -907,13 +911,6 @@ void Ekf::controlHeightFusion()
 {
 	// set control flags for the desired primary height source
 
-	if (_range_data_ready) {
-		// correct the range data for position offset relative to the IMU
-		Vector3f pos_offset_body = _params.rng_pos_body - _params.imu_pos_body;
-		Vector3f pos_offset_earth = _R_to_earth * pos_offset_body;
-		_range_sample_delayed.rng += pos_offset_earth(2) / _R_rng_to_earth_2_2;
-	}
-
 	rangeAidConditionsMet();
 
 	_range_aid_mode_selected = (_params.range_aid == 1) && _range_aid_mode_enabled;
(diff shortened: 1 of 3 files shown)


dakejahl:
@priseborough In `range_aid` mode shouldn't the ekf be fusing all 3 observations (gps, baro, distance sensor) if they are all available? This would make a lot more sense as gps provides a slow response, distance sensor a medium (depending on sample rate) and barometer a twitch response. I guess I am not understanding why we don't continue to fuse baro data along with the distance sensor data.

priseborough:
Short answer is no. They all give different height measurements, use a different datum and are subject to different and varying bias errors that would each need an extra  state in the estimator.

mcsauder:
Hi @priseborough , 

It seems like the 3 measurements could/should be fused whenever horizontal velocity is small, as the distance sensor data provides an additional observations of (relative) vertical position.  When horizontal velocity is meaningful, this is problematic, but for hovering flight fusing the distance sensor data along with GPS and baro seems like it could improve the state estimate.

Also, presumably, although the distance sensor is providing only height above ground, its rate of change at low horizontal velocity when corrected for body attitude can provide an additional vertical velocity observation that can be fused with GPS vertical velocity observations and accelerometer integration values. Certainly, sensor noise is a significant consideration in this case and the numerical differentiation is problematic, but perhaps no more problematic than the numerical integration in place.

I agree that relative frame reference/datums must be handled with care, but it still seems like the additional sensor observations could/should be fused in the matrix solution to improve the vertical velocity and position solutions.

Is the primary concern an additional state, (states), in the estimator, or is it simply impractical when it comes down the implementation due to sensor noise, etc.?

Thanks for your advice!

-Mark

priseborough:
@mcsauder I've provided responses below:

>
It seems like the 3 measurements could/should be fused whenever horizontal velocity is small, as the distance sensor data provides an additional observations of (relative) vertical position. When horizontal velocity is meaningful, this is problematic, but for hovering flight fusing the distance sensor data along with GPS and baro seems like it could improve the state estimate.
> 

Yes it could, but as explained previously we would need to add a state to handle drift of the baro sensor. 

> 
Also, presumably, although the distance sensor is providing only height above ground, its rate of change at low horizontal velocity when corrected for body attitude can provide an additional vertical velocity observation that can be fused with GPS vertical velocity observations and accelerometer integration values.
> 

That is making an assumption about  the terrain under the vehicle.

> 
Certainly, sensor noise is a significant consideration in this case and the numerical differentiation is problematic, but perhaps no more problematic than the numerical integration in place.
> 

Direct differentiation of sensor dat is not the solution, particularly given a sensor (eg range) that is prone to data dropouts and spikes. I don't understand the point you are making about numerical integration.

> 
I agree that relative frame reference/datums must be handled with care, but it still seems like the additional sensor observations could/should be fused in the matrix solution to improve the vertical velocity and position solutions.

Is the primary concern an additional state, (states), in the estimator, or is it simply impractical when it comes down the implementation due to sensor noise, etc.?
> 

The primary concern is the amount of work required to add states to the estimator vs the potential improvement and limited use case., It is a major work package that touches at the matlab derivation, autocode fragments, reset functions, etc. 

Sensor noise is not an issue, but when fusing range finder into the main estimator, sensor failure mechanisms and terrain variation are. 

I'm not saying we shouldn't pursue it in the future, but right now it is low on the priority list.

CarlOlsson:
@priseborough Is there a specific reason you assigned me to review this PR? We don't use the range aid feature so I don't think I am the best suited to review and test this. Or are you also fixing [this](https://github.com/PX4/ecl/issues/424) here?

priseborough:
@CarlOlsson  It was because it substantially changes the function that does the 'stuck' reading test which you wrote. I have attempted to preserve the stuck reading functionality but wanted you to be OK with the change.

Thank you for reminding me about https://github.com/PX4/ecl/issues/424 I will add a commit to fix it.

priseborough:
@CarlOlsson Due to the stuck range test interfering with the resumption of optical flow use in SITL I have made the 1.0m difference test threshold a parameter with a smaller default.

CarlOlsson:
> It was because it substantially changes the function that does the 'stuck' reading test which you wrote.

Okey I see, actually it was @ChristophTobler who [wrote it](https://github.com/PX4/ecl/pull/314), we just discovered it at approximately at the same time.

> I have attempted to preserve the stuck reading functionality but wanted you to be OK with the change.

I am ok with the changes. However, we should expose the ```range_stuck_threshold``` to the user since the new default value is very low and effectively disables the check for a lot of cases. Note that the problem with the sensor being stuck after going out of range for an extended period of time affects up to 3 % of the LIDAR-Lite v3 sensors.

dakejahl:
@CarlOlsson Would it make more sense to handle device specific failures like this within their respective drivers?

CarlOlsson:
> @CarlOlsson Would it make more sense to handle device specific failures like this within their respective drivers?

Yes I think so

PR #485, opened 2018-07-23
l1 position control: added roll angle setpoint slew rate limiting

RomanBapst:
This adds slew rate limiting of the roll angle setpoint to the l1 controller.
It also implements a private method to calculate the roll setpoint (update_roll_setpoint) and a public method to get the roll setpoint (get_roll_setpoint). Previously the roll setpoint was calculated more than one time during one control cycle.

The aim of this is to avoid sudden steps in roll angle setpoint which can occur if there is a switch in waypoints.

### DIFF
--- a/l1/ecl_l1_pos_controller.cpp
+++ b/l1/ecl_l1_pos_controller.cpp
@@ -58,11 +58,7 @@ void ECL_L1_Pos_Controller::update_roll_setpoint()
 	}
 
 	// slew rate limiting active
-	if ((roll_new - _roll_setpoint) / _dt > _roll_slew_rate) {
-		roll_new = _roll_setpoint + _roll_slew_rate * _dt;
-	} else if ((roll_new - _roll_setpoint) / _dt < -_roll_slew_rate) {
-		roll_new = _roll_setpoint - _roll_slew_rate * _dt;
-	}
+	roll_new = math::constrain(roll_new, _roll_setpoint - _roll_slew_rate * _dt, _roll_setpoint + _roll_slew_rate * _dt);
 
 	_roll_setpoint = roll_new;
 
--- a/l1/ecl_l1_pos_controller.cpp
+++ b/l1/ecl_l1_pos_controller.cpp
@@ -45,11 +45,27 @@
 using matrix::Vector2f;
 using matrix::wrap_pi;
 
-float ECL_L1_Pos_Controller::nav_roll()
+
+void ECL_L1_Pos_Controller::update_roll_setpoint()
 {
-	float ret = atanf(_lateral_accel * 1.0f / CONSTANTS_ONE_G);
-	ret = math::constrain(ret, -_roll_lim_rad, _roll_lim_rad);
-	return ret;
+	float roll_new = atanf(_lateral_accel * 1.0f / CONSTANTS_ONE_G);
+	roll_new = math::constrain(roll_new, -_roll_lim_rad, _roll_lim_rad);
+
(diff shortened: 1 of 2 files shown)


l1/ecl_l1_pos_controller.cpp
```diff
@@ -45,11 +45,27 @@
...
+	float roll_new = atanf(_lateral_accel * 1.0f / CONSTANTS_ONE_G);
+	roll_new = math::constrain(roll_new, -_roll_lim_rad, _roll_lim_rad);
+
+	// no slew rate limiting active
+	if (_dt <= 0.0f || _roll_slew_rate <= 0.0f) {
+		_roll_setpoint = roll_new;
+		return;
+	}
+
+	// slew rate limiting active
```
dagar: Constrain to _roll_setpoint +- _roll_slew_rate * _dt?

l1/ecl_l1_pos_controller.cpp
```diff
@@ -45,11 +45,27 @@
...
+void ECL_L1_Pos_Controller::update_roll_setpoint()
 {
-	float ret = atanf(_lateral_accel * 1.0f / CONSTANTS_ONE_G);
-	ret = math::constrain(ret, -_roll_lim_rad, _roll_lim_rad);
-	return ret;
+	float roll_new = atanf(_lateral_accel * 1.0f / CONSTANTS_ONE_G);
+	roll_new = math::constrain(roll_new, -_roll_lim_rad, _roll_lim_rad);
+
+	// no slew rate limiting active
+	if (_dt <= 0.0f || _roll_slew_rate <= 0.0f) {
```
dagar: What about storing only the last timestamp within L1 and have it calculate its own dt as needed?

RomanBapst: @dagar Yeah I was wondering about that too. I checked TECS and that's apparently the way you do it.

dagar: It's not necessarily "the right way", but avoiding the need for multiple calls into something in the same iteration certainly makes the API harder to use incorrectly.

l1/ecl_l1_pos_controller.cpp
RomanBapst: @dagar Done!

tstastny:
@RomanBapst Ah, my bad -- I looked at the equations wrong -- The time constant there is acting on the attitude error, not the setpoint itself.

> I think as long as you don't overdo it, e.g. setting the limit very low you will not see any degradation in control performance. Keep in mind that the controller will most probably only hit the slew rate when the switch between waypoints happens, for the rest the roll setpoint is very smooth already.

Sure - should be fine.

RomanBapst:
I've tested this in SITL once without slew-rate active and once with a slew rate of 90 deg/s.
You don't really notice any difference in the tracking performance but the steps in roll actuation are gone.

**Without slew active**
no_slew

Body x rate setpoint and roll actuator control.
log_no_slew

**Slew rate active**
slew

Body x rate setpoint and roll actuator control.
log_slew

l1/ecl_l1_pos_controller.cpp
dagar: That or grab its own timestamp internally when called. L1 should either track its own state to do this independently or stick to the simple return. It's nitpicky at this point, but I think we should try to avoid letting the API become awkward (eg tecs) if anyone is ever possibly going to use this outside of PX4.

PR #488, opened 2018-07-30
cmake ignore doxygen if not a standalone build

dagar:
Small change to enable PX4 top level generation.


dagar:
This change is harmless.

PR #491, opened 2018-08-07
Fix invisible ceiling when using range aiding

dakejahl:
This PR adds a check for **range aiding** to determine if we are `relying_on_rangefinder`. This fixes the bug we are experiencing where we are limited to the range finders max height above ground when using it in aiding mode. If range aiding is not enabled, and thus the range finder _is_ intended to be the sole source of height observation, then this check will evaluate true and the controller will still properly limit the height above ground.

Let me know if you have any questions or concerns, thanks!
Jake

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -1091,7 +1091,7 @@ void Ekf::get_ekf_ctrl_limits(float *vxy_max, float *vz_max, float *hagl_min, fl
 
 	// TODO : calculate visual odometry limits
 
-	bool relying_on_rangefinder = _control_status.flags.rng_hgt;
+	bool relying_on_rangefinder = _control_status.flags.rng_hgt && !_params.range_aid;
 
 	bool relying_on_optical_flow = _control_status.flags.opt_flow && !(_control_status.flags.gps || _control_status.flags.ev_pos);
 


mhkabir:
Nice fix! Looks good to me.

priseborough:
If enabling range aiding when on ground, we need to make sure if the vehicle is tilted excessively that the switch to baro happens immediately and we don't wait for a timeout. The decision to not  perform range aiding when on ground was due to problems with vehicle being carried and tilted. I worry that changing that behaviour creates unnecessary risk given that baro height is more reliable than range height prior to motor start.

EKF/control.cpp
```diff
@@ -1083,12 +1083,11 @@ void Ekf::rangeAidConditionsMet()
 {
 	// if the parameter for range aid is enabled we allow to switch from using the primary height source to using range finder as height source
 	// under the following conditions
-	// 1) Vehicle is in-air
-	// 2) Range data is valid
-	// 3) Vehicle is no further than max_hagl_for_range_aid away from the ground
-	// 4) Ground speed is not higher than max_vel_for_range_aid
-	// 5) Terrain estimate is stable (needs better checks)
-	if (_control_status.flags.in_air && !_rng_hgt_faulty) {
+	// 1) Range data is valid
```
priseborough: This change is not required to fix the 'invisible ceiling; defect. It increase vulnerability to range finder obscuration and tilt when the vehicle is being handled prior to takeoff.

EKF/ekf_helper.cpp
```diff
@@ -1091,7 +1091,7 @@ void Ekf::get_ekf_ctrl_limits(float *vxy_max, float *vz_max, float *hagl_min, fl
 
 	// TODO : calculate visual odometry limits
 
-	bool relying_on_rangefinder = _control_status.flags.rng_hgt;
+	bool relying_on_rangefinder = _control_status.flags.rng_hgt && !_params.range_aid;
```
priseborough: Agree with this change.

dakejahl:
@priseborough Removed.

dakejahl:
@priseborough

PR #493, opened 2018-08-08
Reset baro datum to current state when switching out of range aiding

dakejahl:
This PR introduces 2 changes with respect to **range aiding**.
1. New function `Ekf::resetBaroDatum`. This function simply resets the datum so that the current baro measurements match the predicted state. I replaced two instances of this behavior and added one new one. The new instance of this resetting of baro datum occurs when switching back to the barometer from rangefinder in **range aid** mode. I have flight tested this and collected some data at the estimator level with some custom topic logging. 
baro_hgt_reset
rng_aiding

2. Allow range sensor measurements to be used while `landed` as long as the measurements are within +/-10cm of the expected `rng_gnd_clearance` which is defined by **EKF2_MIN_RNG**

https://review.px4.io/plot_app?log=95753bfa-f112-4f9b-9d01-54c9ba92152e

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -934,7 +934,8 @@ void Ekf::controlHeightFusion()
 			setControlBaroHeight();
 			_fuse_height = true;
 
-			// we have just switched to using baro height, we don't need to set a height sensor offset
+			// we have just switched to using baro height, use current estimate to recalculate the height offset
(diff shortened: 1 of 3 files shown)


dakejahl:
I am not sure why **range aiding** turned off there at the end. I think I missed something in the `terrain_estimator.cpp`

PR #494, opened 2018-08-12
L1 enhancements: handling small loiter radii and high winds

tstastny:
See description in https://github.com/PX4/Firmware/pull/10219

### DIFF
--- a/l1/ecl_l1_pos_controller.cpp
+++ b/l1/ecl_l1_pos_controller.cpp
@@ -79,8 +79,12 @@ ECL_L1_Pos_Controller::navigate_waypoints(const Vector2f &vector_A, const Vector
 	float xtrack_vel = 0.0f;
 	float ltrack_vel = 0.0f;
 
+	/* reset _L1_ratio */
+	set_l1_period(_L1_period);
+
 	/* get the direction between the last (visited) and next waypoint */
-	_target_bearing = get_bearing_to_next_waypoint((double)vector_curr_position(0), (double)vector_curr_position(1), (double)vector_B(0), (double)vector_B(1));
+	_target_bearing = get_bearing_to_next_waypoint((double)vector_curr_position(0), (double)vector_curr_position(1),
+			  (double)vector_B(0), (double)vector_B(1));
 
 	/* enforce a minimum ground speed of 0.1 m/s to avoid singularities */
 	float ground_speed = math::max(ground_speed_vector.length(), 0.1f);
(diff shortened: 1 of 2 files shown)


chamois94:
I could try the new algorithm in SITL with such a powerful wind (half the maximum allowed airpeed) and I encoutered a problem when the wind estimate is not correct. UAV is in loiter mode around a point and I applied a very quick 180° rotation to the wind vector (I know it's not physically possible but it was to test the robutsness of the control loop) : in this case the UAV ran away in a 90° direction from the wind. The problem is that in this case the EKF will never be able to converge to a good estimation but does not know it. I know it seems logical but maybe we should add a safety in this case ?

l1/ecl_l1_pos_controller.cpp
```diff
@@ -70,20 +70,40 @@ float ECL_L1_Pos_Controller::switch_distance(float wp_radius)
...
+{
+	/* following [3] */
+	// NOTE: uses current "lambda" angle, this function should only be called *after any other navigation function
+
+	const float airsp_incr_max = math::max(airsp_max - airsp_ref, 0.0f);
+
+	airspeed = math::max(airspeed, 0.1f);
+	min_ground_speed = math::max(min_ground_speed, 0.0f);
+
+	const float wind_ratio = (wind_speed + min_ground_speed) / airspeed;
```
priseborough: We will need to do a sensitivity analysis to determine the effect of wind speed estimation errors on this method. We may have to add some additional protection using the dot product of the X body axis projected onto earth axes and the velocity vector to determine when forwards speed has dropped below a threshold and increment the airspeed demand accordingly.

l1/ecl_l1_pos_controller.cpp
```diff
@@ -200,107 +222,95 @@ ECL_L1_Pos_Controller::navigate_waypoints(const Vector2f &vector_A, const Vector
...
-	float xtrack_vel_circle = -ltrack_vel_center;
 	/* radial distance from the loiter circle (not center) */
-	float xtrack_err_circle = vector_A_to_airplane.length() - radius;
+	const float xtrack_err_circle = norm_vector_A_to_airplane - radius;
+
+	/* check circle tracking feasibility */
+	if (_L1_distance > radius && fabsf(xtrack_err_circle) <= _L1_distance) {
+		// re-calculate L1 ratio & distance (effectively reduce period to allow tracking small radii)
+		// NOTE: when far from the loiter radius, operator defined period and damping are unaffected
+		_L1_ratio = math::max(fabsf(xtrack_err_circle), radius) / ground_speed;
```
priseborough: Reducing L1 distance does increase the guidance loop gain which can produce limit cycling in roll if the roll angle AP has not been tuned correctly and the vehicle has excessive lag from demanded to achieved lateral acceleration. This can be caused by lack of tuning or by adverse yaw. That is the main disadvantage to this approach and the main reason why the original method  of tracking circles was implemented.

We need to update user documentation so that they will know what to do if this happens.

l1/ecl_l1_pos_controller.cpp
tstastny: Totally. I've done some minor playing around with the effects of wrong wind estimates (mainly cases where the airspeed sensor is not properly calibrated and either over or under estimates) -- small errors (order of 1-2m/s) don't effect things much -- but it has been on my list to do a proper analysis (just haven't gotten around to it). However I guess I probably should have also made it clear that this functionality **relies** on a proper wind estimate. So if e.g. the airspeed sensor is not calibrated, or sideslip fusion has some significant uncertainty, it will not work as nicely -- however, if the wind estimates are good enough, it is much more efficient than the ground speed undershoot solution. Nevertheless, I agree with you that it is probably a good idea to put back in a fallback solution to some form of safe guard with the x-body - gsp vector dot product especially when the wind estimate e.g. goes invalid or has high uncertainty.

Something to note: the logic is currently not used when the wind estimate is invalid (defaults to ground speed and "zero" wind speed as input args) -- this was one of the "open questions" checkboxes I left on the PR, that we would need to think what to do when this case happens -- so I guess again this points towards a defaulted ground speed undershoot. I'll look into how to put it back in and make it play nice with the other logic in the next day or so.

chamois94:
any news on this pull request ?

priseborough:
My concern about the guidance loop stability implications of reducing L1 distance to enable use in circular loiter has not been addressed. The problem this caused in the past was why we use a difference guidance strategy for circular loiter. Introducing this change could destabilise exisiting setups, so until we have a way of ensuring the guidance loop gain in circular loiter does not increase above what the user has tuned previously, this change is not ready for general use.

priseborough:
@tstastny its been 4 months so I will close this. Feel free to re-open when you have a solution to maintaining gain from lateral velocity/position error for legacy setups.

PR #496, opened 2018-08-20
l1 position controller: protect against NAN

RomanBapst:
@eyeam3 mentioned that under certain circumstances the the l1 position controller can produce a NAN roll setpoint. If we don't protect the slew rate limit logic against this the state will not be able to recover.

@eyeam3 Thanks for your input! Do you want to review? :-)


RomanBapst:
@dagar Done

PR #498, opened 2018-08-26
EKF: Add missing optical flow ground motion protection

priseborough:
Motion compensated optical flow rates are supposed to be zeroed if flow data is failing validity checks when on ground. This was not being done in the code. The comments and logic have been amended to be consistent and make the design intent clearer.

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -376,9 +376,10 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
 		// check quality metric
 		bool flow_quality_good = (flow->quality >= _params.flow_qual_min);
 
-		// Always use data when on ground to allow for bad quality due to unfocussed sensors and operator handling
-		// If flow quality fails checks on ground, assume zero flow rate after body rate compensation
-		if ((delta_time_good && flow_quality_good && (flow_magnitude_good || relying_on_flow)) || !_control_status.flags.in_air) {
+		// Check data validity and write to buffers
+		// Use a zero velocity assumption to constrain drift when on-ground if necessary
+		float use_flow_data_to_navigate = delta_time_good && flow_quality_good && (flow_magnitude_good || relying_on_flow);
+		if (use_flow_data_to_navigate || (!_control_status.flags.in_air && relying_on_flow)) {
 			flowSample optflow_sample_new;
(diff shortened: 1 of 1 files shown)


mhkabir:
This broke optical flow. The EKF now diverges if optical flow aiding is enabled. Haven't investigated further yet.

Replay log : 
https://review.px4.io/plot_app?log=09958de4-6654-4585-a0b0-eab8a1e4ca66

There is a bad timestamp in the log, which causes the weird plot visualisation, but it should  be fine otherwise.

priseborough:
@mhkabir I reviewed the code and found a potential issue which has been fixed by 
https://review.px4.io/plot_app?log=09958de4-6654-4585-a0b0-eab8a1e4ca66 however the reported flow quality on your log is always above the minimum so this should not have affected your setup.

With this fix, optical flow is working well on SITL 'make posix gazebo_iris_opt_flow'

I was unable to get replay working on your log using upstream master with or without the fix. The range data and flow data timestamps cover different ranges and there are lots of differences in message formats reported by replay. You need to get your timestamp issue fixed and provide data recorded with up to date message definitions.

priseborough:
The flow data doesn't start until 34 seconds.

mhkabir:
@priseborough Here is another log where is EKF diverges on the ground after some moving around: https://logs.px4.io/plot_app?log=ea18fc88-8ce5-4382-ab62-35a9a0d015f3. It's pretty reproducible and happens sometimes even without any motion while on the ground.

For replay, you will need to be on v1.8.

mhkabir:
I am able to replay the above log successfully, and the EKF diverges with optical flow enabled, but does not with optical flow disabled.

priseborough:
There is a period just after motors start where there are repeated rapid accel transients and high angular rates. It looks like the vehicle is rapidly bouncing between ground contact points. This motion is causing the estimator velocities to diverge. Setting EKF2_OF_QMIN to a larger value to force use of zero motion compensated flow rates did not help, so another strategy is needed to handle the characteristics of this hardware.

screen shot 2018-09-03 at 7 57 24 am

Will investigate further.

mhkabir:
I found the fix actually. From the log, flow use is successfully inhibited by the "motion excessive" tests, however the check for no fusion over a time period is incorrect. 

Will push to a branch and ask for review.

priseborough:
That makes sense. The angular rates and accelerations seen during takeoff were high. I've reviewed, approved and merged the PR.

PR #502, opened 2018-09-02
EKF: Fix bug that prevents use of flow sensors without gyros

priseborough:
The handling of invalid flow data when on ground is performed in controlOpticalFlowFusion() where it is able to handle the case of flow sensors that don't publish gyro data. By handling the on-ground with invalid flow data case in the interface before data is pushed into the buffer, the change introduced by commit a53ad9c removed the ability to handle flow sensors that publish NAN for gyro data. This was not picked up in testing because the SITL model does nor simulate that type of sensor.

**Testing**

Testing performed was on SITL with 'make posix gazebo_iris_opt_flow' target

Default test log with flow sensor gyro data: https://logs.px4.io/plot_app?log=ca849660-3d78-4a60-833e-84d1ad9dd63e

The simulator_mavlink.cpp file was then modified to set the published flow gyro integral data to NAN

	//rotate_3f(flow_rot, flow.gyro_x_rate_integral, flow.gyro_y_rate_integral, flow.gyro_z_rate_integral);
	flow.gyro_x_rate_integral = NAN;
	flow.gyro_y_rate_integral = NAN;
	flow.gyro_z_rate_integral = NAN;

 https://logs.px4.io/plot_app?log=0dbf3ee3-cf69-4094-b305-1862f5c145ba

screen shot 2018-09-03 at 9 00 28 am

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -375,7 +375,7 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
 		bool flow_quality_good = (flow->quality >= _params.flow_qual_min);
 
 		// Check data validity and write to buffers
-		// Use a zero velocity assumption to constrain drift when on-ground if necessary
+		// Invalid flow data is allowed when on ground and is handled as a special case in controlOpticalFlowFusion()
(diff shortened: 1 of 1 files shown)


EliaTarasov:
@mhkabir any news on this testing?

mhkabir:
@EliaTarasov this works fine, and is a part of #503. Please test that PR.

PR #505, opened 2018-09-04
Optical flow fusion control improvements

mhkabir:
Brings in some fixes and improvements in optical flow fusion control. Includes https://github.com/PX4/ecl/pull/502 as it was needed for flight testing. Most of these corner cases are more visible on platforms running optical flow algorithms offboard. (e.g Aero,  Snapdragon Flight)

Functional changes : 
1. Add protection against large dts, which can put the fusion control into a bad state (if fusion stops, it never starts again) due to float overflows and unsigned integer checks.
 (https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L496)
2. Remove unnecessary check for tilt angle when popping new data from flow buffer. There is a similar check downstream, and this one was interfering with correct fusion inhibition detection.
3. Handle flow data timeouts correctly (fusion timeouts != data stopped)
4. Remove state predictor alignment which doesn't seem to be used for any of the other aiding sources.  
5. Add 2 minor `ECL_INFO` statements which go a long way in figuring out what went wrong in logs.

Log :
https://logs.px4.io/plot_app?log=a13404d3-078e-4f04-94e0-19a1f287bc72

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -588,7 +588,6 @@ void Ekf::controlGpsFusion()
 
 					if (_control_status.flags.gps) {
 						ECL_INFO("EKF commencing GPS fusion");
-						_time_last_gps = _time_last_imu;
 					}
 				}
 			}
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -119,8 +119,7 @@ void Ekf::controlFusionModes()
 	// We don't fuse flow data immediately because we have to wait for the mid integration point to fall behind the fusion time horizon.
 	// This means we stop looking for new data until the old data has been fused.
 	if (!_flow_data_ready) {
-		_flow_data_ready = _flow_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_flow_sample_delayed)
-				   && (_R_to_earth(2, 2) > _params.range_cos_max_tilt);
+		_flow_data_ready = _flow_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_flow_sample_delayed);
 	}
 
 	_ev_data_ready = _ext_vision_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_ev_sample_delayed);
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -446,10 +446,6 @@ void Ekf::controlOpticalFlowFusion()
 				if (!_control_status.flags.gps || !_control_status.flags.ev_pos) {
 					resetVelocity();
 					resetPosition();
-
-					// align the output observer to the EKF states
-					alignOutputFilter();
-
 				}
 			}
 
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -418,38 +418,31 @@ void Ekf::controlOpticalFlowFusion()
 			}
 		}
 
-		// Handle cases where we are using optical flow but are no longer able to because data is old
-		// or its use has been inhibited.
-		if (_control_status.flags.opt_flow) {
-		       if (_inhibit_flow_use) {
-			       _control_status.flags.opt_flow = false;
-			       _time_last_of_fuse = 0;
-
-			} else if ((_time_last_imu - _time_last_of_fuse) > (uint64_t)_params.reset_timeout_max) {
-				_control_status.flags.opt_flow = false;
-
-			}
+		// Handle cases where we are using optical flow but are no longer able to because
+		// its use has been inhibited.
+		if (_control_status.flags.opt_flow && _inhibit_flow_use) {
+			ECL_WARN("Optical flow data quality poor - stopping use");
+			_control_status.flags.opt_flow = false;
 		}
 
 		// optical flow fusion mode selection logic
 		if ((_params.fusion_mode & MASK_USE_OF) // optical flow has been selected by the user
-			&& !_control_status.flags.opt_flow // we are not yet using flow data
+			&& !_control_status.flags.opt_flow 	// we are not yet using flow data
 			&& _control_status.flags.tilt_align // we know our tilt attitude
-			&& !_inhibit_flow_use
-			&& get_terrain_valid()) // we have a valid distance to ground estimate
+			&& !_inhibit_flow_use				// use is not inhibited
+			&& get_terrain_valid()) 			// we have a valid distance to ground estimate
 		{
 			// If the heading is not aligned, reset the yaw and magnetic field states
 			if (!_control_status.flags.yaw_align) {
 				_control_status.flags.yaw_align = resetMagHeading(_mag_sample_delayed.mag);
 			}
 
-			// If the heading is valid and use is not inhibited , start using optical flow aiding
+			// If the heading is valid, start using optical flow aiding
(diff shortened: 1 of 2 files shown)


mhkabir:
More testing revealed an additional corner case which can cause divergence. Let me investigate further.

xdwgood:
screenshot from 2018-09-06 10-54-14

mhkabir:
@xdwgood Additional context?

xdwgood:
@mhkabir  I use the log you provided in this PR.

mhkabir:
@priseborough Can you please have a look here? The fixes here are crucial and necessary, and some help figuring out the last 5% would be appreciated, thanks!

priseborough:
OK will do

mhkabir:
@priseborough Any luck?

mhkabir:
@priseborough Ping.

priseborough:
@mhkabir I assume the last 5% you are referring to is the position divergence at 375 seconds in the log https://review.px4.io/plot_app?log=488ad808-7886-4372-a0da-7f51a04be2fe .

Unfortunately there is nothing that stands out in the data itself to explain what happened so I tried running a replay using upstream master but was unable to due to the large number of data dropouts and or the mismatched topics. Where did you fork form upstream for PX4/Firmware?

Resolving this is going to require a data log suitable for replay. If replacing the SD card with a better quality one or reformatting the one you doesn't work, you may have to set SDLOG_PROFILE = 2

priseborough:
@mhkabir Are you able to provide a log suitable for replay? Without that, diagnosing the divergence seen in your testing is impractical.

mhkabir:
@RomanBapst @DanielePettenuzzo Could you guys give this a test with your setup after rebasing on master?

priseborough:
@mhkabir The reported position divergence that may have been fixed by the covariance reset fixes that have gone in recently. It would be worth rebasing and retesting.

If the divergence issue is still present, then a replay log is required to perform the required diagnostics.

mhkabir:
I'll rebase.

RomanBapst:
@mhkabir Can you please rebase this on master? I have a vehicle here ready for optical flow and I can provide a replay log so that we can continue with this.

mhkabir:
@RomanBapst @DanielePettenuzzo Rebased, please test!

priseborough:
@mhkabir Have you done your own testing on this PR ?

EKF/estimator_interface.cpp
```diff
@@ -366,31 +366,28 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
 		// of min arrival interval because too much data is being lost
 		float delta_time = 1e-6f * (float)flow->dt;
 		float delta_time_min = 5e-7f * (float)_min_obs_interval_us;
-		bool delta_time_good = delta_time >= delta_time_min;
+		float delta_time_max = 0.1f;
```
priseborough: Is this value based on testing or set above what current supported sensors output. It would prevent use of a 10Hz sensor.

priseborough:
The change to reset logic in https://github.com/PX4/ecl/pull/505/commits/377829b434a7c8b03f0f5d43ac1996f115dcd405 requires careful testing.

EKF/estimator_interface.cpp
dagar: > Is this value based on testing or set above what current supported sensors output. It would prevent use of a 10Hz sensor.

@mhkabir ?

mhkabir: I set 10Hz fairly arbitrarily. I can increase to 5Hz minimum. Will this be fine? 

The PMW3901 runs at 40Hz. I'm not sure about PX4Flow.

mhkabir:
@jkflying @RomanBapst Would you be able to do some more structured testing to the reset logic. These should be carried out with logging from boot, and optical flow as the only enabled sensor in `EKF2_AID_MASK`: 

1. Shake the vehicle fast, which should trigger the bad data quality checks (https://github.com/PX4/ecl/pull/505/files#diff-4cac31c333867233bb8a08de1e812a9aR424). Fusion should recover once the shaking is stopped.

2. Disconnect the sensor physically or stop the driver from console while fusion is active. This should trigger the data loss checks (https://github.com/PX4/ecl/pull/505/files#diff-4cac31c333867233bb8a08de1e812a9aR490). 

In both of these tests, local position should go invalid with appropriate messages in the console, and then recover once initial conditions are restored.

priseborough:
Any update on completing testing? I do not support merging this until the testing is complete.

mhkabir:
@jkflying @RomanBapst

EKF/control.cpp
```diff
@@ -419,45 +418,34 @@ void Ekf::controlOpticalFlowFusion()
...
-			       _time_last_of_fuse = 0;
-
-			} else if ((_time_last_imu - _time_last_of_fuse) > (uint64_t)_params.reset_timeout_max) {
-				_control_status.flags.opt_flow = false;
-
-			}
+		// Handle cases where we are using optical flow but are no longer able to because
+		// its use has been inhibited.
+		if (_control_status.flags.opt_flow && _inhibit_flow_use) {
+			ECL_WARN("Optical flow data quality poor - stopping use");
```
priseborough: _inhibit_flow_use can toggle on and off as the vehicle goes between hover and moving if it is also using GPS. Because of this this message may flood the interface.

priseborough: Also this ability to switch flow use on and off depending on vehicle movement (when it is also using GPS) is a feature and does not necessarily indicate a fault or bad flow data, just an operating condition where it is better to not use flow.

EKF/control.cpp
```diff
@@ -119,8 +119,7 @@ void Ekf::controlFusionModes()
 	// We don't fuse flow data immediately because we have to wait for the mid integration point to fall behind the fusion time horizon.
 	// This means we stop looking for new data until the old data has been fused.
 	if (!_flow_data_ready) {
-		_flow_data_ready = _flow_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_flow_sample_delayed)
-				   && (_R_to_earth(2, 2) > _params.range_cos_max_tilt);
+		_flow_data_ready = _flow_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_flow_sample_delayed);
```
priseborough: I don't see downstream where the use of flow data when in air is stopped if the vehicle is tilted excessively other than indirectly through loss of range fusion which them after several seconds causes terrain to go invalid.

I think this change will allow use of flow data in flight when excessively tilted.

priseborough:
@RomanBapst and I walked through the changes and found  two issues that need to be fixed - see comments.

priseborough:
@mhkabir Any progress?

mhkabir:
Working on it. I have a flow setup going again with this new sensor: https://github.com/PX4/Firmware/pull/12148/. Will keep you posted.

EKF/control.cpp
mhkabir: This is here : https://github.com/PX4/ecl/pull/505/files#diff-4cac31c333867233bb8a08de1e812a9aR370

PR #507, opened 2018-09-06
Add vision height covariance passthrough

dgonano:
For the vision input to the ekf there is only single variance figure used (`posErr` in the structs) for position representing a spherical error around said position. Some companion computer sensor systems have different errors on horizontal and vertical positions. I guess this is one of the reasons you can set the height mode independently via the `EKF2_HGT_MODE` flag. 
I came across the issue where I wanted to influence the horizontal position accurately, but influence the vertical position only a little (or varying variances). 

I've expanded the struct in this repo and included a separate `hgtErr` which is used on this line instead. I've also modified `Firmware` to extract it from the mavlink message and pass it through. It's all tested and working with the variance correctly influencing the EKF. If this is wanted I can issue a PR in Firmware also with the change.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -81,8 +81,8 @@ struct ext_vision_message {
 	Vector3f posNED;	///< measured NED position relative to the local origin (m)
 	Quatf quat;		///< measured quaternion orientation defining rotation from NED to body frame
 	float posErr;		///< 1-Sigma horizontal position accuracy (m)
+	float hgtErr;		///< 1-Sigma height accuracy (m)
 	float angErr;		///< 1-Sigma angular error (rad)
-	float hgtErr;		///< 1-Sigma height error (rad)
 };
 
 struct outputSample {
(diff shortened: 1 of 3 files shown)


EKF/common.h
```diff
@@ -80,8 +80,9 @@ struct flow_message {
 struct ext_vision_message {
 	Vector3f posNED;	///< measured NED position relative to the local origin (m)
 	Quatf quat;		///< measured quaternion orientation defining rotation from NED to body frame
-	float posErr;		///< 1-Sigma spherical position accuracy (m)
+	float posErr;		///< 1-Sigma horizontal position accuracy (m)
 	float angErr;		///< 1-Sigma angular error (rad)
+	float hgtErr;		///< 1-Sigma height error (rad)
```
mhkabir: Should be "1-Sigma vertical position accuracy (m)"

mhkabir: Also move this up after posError

mhkabir:
Generally looks great except the nitpick!

mhkabir:
@priseborough Please merge when ready.

priseborough:
@dgonano Can you please follow up with the changes required in ekf2_main.cpp so that the ev_data.hgt_err gets set correctly. We can't update PX4/Firmware to use this until that happens

dgonano:
@priseborough Already PR'd and ecl submodule updated here: https://github.com/PX4/Firmware/pull/10432
Just waiting on approval / merge by those who can

PR #510, opened 2018-09-11
EKF: Protect against bad yaw immediately after a yaw reset.

priseborough:
**Bug Description**

A loss of navigation following takeoff from a surface with high levels of magnetic interference was reported. This can be verified by plotting the magnetic heading innovation (the EKF uses this up to 5m where it switches to using 3-axis mag fusion), integrated z gyro, baro alt and measured body axis yaw of xy mag data.

screen shot 2018-09-11 at 8 13 21 am

The magnetic yaw change with height was extremely large in this instance and about 2 radians or 115 degrees. What is also noticeable this instance is that there is very little change in magnetic field strength with initial height gain.

The EKF is supposed to perform a reset of the heading at 5 metres altitude before it switches to using 3-axis fusion. This reset was performed, but the NE earth field estimates collapse immediately. This is correlated to a yaw transient that occurs at the same time. This transient appears to to be the combination of residual uncompensated yaw reset where the yaw setpoint does not precisely match the change in yaw estimate and a closed loop interaction between the EKF and the yaw controller. When the EKF has to a yaw reset of > 15 degrees it also resets the quaternion covariances in addition to those for the magnetic field states.

screen shot 2018-09-11 at 8 32 31 am

The code prevents the 3-axis fusion from modifying states other than the NED and XYZ mag field estimates during the first 5 seconds after a reset to prevent small variations in roll/pitch that were perceptible to operators. The unintended consequence of this is that if the yaw reset is large and covariances are modified, GPS and inertial sensing errors can result in a bad yaw estimate being formed before the quaternion covariance values stabilise.

**Fix Description**

This PR introduces logic that when a quaternion covariance reset has been performed, causes the heading fusion to be performed in addition to 3-axis fusion until the 3-axis fusion starts constraining heading drift.

**Testing**

Used the following procedure:  

1. Place the vehicle on the ground and locate a magnet near the body that causes the compass heading to rotate by over 90 degrees.
2. Ensure logging of replay data from boot is enabled
3. Ensure ALTCTL mode can be selected
4. Repower the vehicle and wait for checks to pass
5. Takeoff in POSCTL mode and continue the climb to above 3m. Be ready to switch to ALTCTL if the position diverges.

Step 5. was varied by including yaw rotations to coincide with the yaw reset.

See comment below for list of log files and analysis.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1440,6 +1440,10 @@ void Ekf::controlMagFusion()
 			// before they are used to constrain heading drift
 			_flt_mag_align_converging = ((_imu_sample_delayed.time_us - _flt_mag_align_start_time) < (uint64_t)5e6);
 
+			// If the yaw reset has been large enough to require rewt of the quaternion covariance sub-matrix
+			// we continue to use heading fusion to constrain heading drift to give the covariance time to stabilise
+			_flt_yaw_align_converging = ((_imu_sample_delayed.time_us - _flt_yaw_align_start_time) < (uint64_t)5e6);
+
 			if (!_control_status.flags.update_mag_states_only && _control_status_prev.flags.update_mag_states_only) {
 				// When re-commencing use of magnetometer to correct vehicle states
 				// set the field state variance to the observation variance and zero
@@ -1503,6 +1507,12 @@ void Ekf::controlMagFusion()
 
 		// fuse magnetometer data using the selected methods
 		if (_control_status.flags.mag_3D && _control_status.flags.yaw_align) {
+			// Always heading fusion to assit with initial yaw convergence if
+			// quaternion state covariances have been reset.
+			if (_flt_yaw_align_converging) {
+				fuseHeading();
+			}
+
 			fuseMag();
 
 			if (_control_status.flags.mag_dec) {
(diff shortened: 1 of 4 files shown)


priseborough:
The method of starting the vehicle near a magnet was able to produce a magnetic heading error at ground level of over 2.4 radians:

screen shot 2018-09-12 at 3 29 43 pm

However heading fusion did not overlap as expected:

screen shot 2018-09-12 at 3 35 31 pm

This behaviour was replicated on replay, so further investigation is required.

priseborough:
The summary from todays testing is that after 10 takeoffs with ground level magnetic yaw errors varying between 0.5 and 2.4 radians, loss of navigation resulting from the yaw reset was encountered and therefore at this stage it cannot be demonstrated that this PR is having a beneficial effect.

For the three replay data only logs these are the innovations before and after application of the PR.

Without the PR:

before 1

before 2

before 3

With the PR:

after 1

after 2

after 3

Further testing with replay logging enabled is required to try and replicate the failure.

chamois94:
did you have time to work on that ?

priseborough:
@chamois94 No time available at the moment.

priseborough:
Replaced by https://github.com/PX4/ecl/pull/549

PR #512, opened 2018-09-14
EKF add new simple setIMUData() interface

dagar:
Once PX4 has an IMU message this will make it slightly easier to insert an IMU message directly.

Small piece of https://github.com/PX4/Firmware/pull/9756.

@dakejahl


mhkabir:
We should look at having separate timestamps for gyro and accel data while we're on it.

priseborough:
This was merged without a Review approval being recorded.

PR #514, opened 2018-10-17
Reinstate lost commits from PR 497

priseborough:
This branch was approved and merged and merged previously - see https://github.com/PX4/ecl/pull/497

However those commits have now disappeared which is worrying.


priseborough:
Resolved by removing the upstream remote from my git client and re-adding it.

Issue #519, opened 2018-11-09
Issues with heading estimation


CarlOlsson:
@priseborough 
There seems to be serious issues present regarding heading estimation at places with high magnetic declination.

The problem is easy to reproduce in SITL. Just [change the location to e.g. Seattle](https://github.com/PX4/sitl_gazebo/blob/3d80f63562c24c1537e1f8423ce649c99ebc15ea/include/gazebo_gps_plugin.h#L97-L100) with a magnetic declination of ~15 degrees, ```make clean```, run ```make posix_sitl_default gazebo``` and fly a mission.

Here is a [logfile](https://logs.px4.io/plot_app?log=ffe9fb87-6a4b-44ae-b781-d58ad959e969) from such a case. I also set EKF2_MAG_DECL to -5 degrees for illustration purpose but it is not needed to expose the bug.

#### The following happens:

t = t0
The system boots up. The true heading is 25 degrees and the true declination is 15 degrees.
The EKF starts to initialize the filter by collecting and low pass filtering the magnetometer and barometer sensor data.

t = t1
When the EKF has collected 2 times the observation buffer length it proceeds with the initialization by running the [resetMagHeading](https://github.com/PX4/ecl/blob/master/EKF/ekf_helper.cpp#L546) function. Here it initialize the heading by taking the last magnetometer measurement, rotate it into the inertial frame using the current attitude estimate ignoring the yaw, calculate the angle between the N and E component and add the magnetic declination. However, the declination is taken from the class variable _mag_declination and since the function [calcMagDeclination()](https://github.com/PX4/ecl/blob/master/EKF/ekf_helper.cpp#L753) has not yet run a single time it will use the [default value of zero](https://github.com/PX4/ecl/blob/master/EKF/ekf.h#L375).

Now the function [initialiseFilter()](https://github.com/PX4/ecl/blob/master/EKF/ekf.cpp#L135) returns true and the function [controlFusionModes()](https://github.com/PX4/ecl/blob/10a0fef7b7db749f5baf3384e9c900d3aa0e8140/EKF/control.cpp#L46-L46) is called. First thing that happens here is that we call [calcMagDeclination()](https://github.com/PX4/ecl/blob/10a0fef7b7db749f5baf3384e9c900d3aa0e8140/EKF/ekf_helper.cpp#L753-L753). In here since _NED_origin_initialised is false we set the _mag_declination variable to the EKF2_MAG_DECL parameter, in our case -5.
[...]

PR #520, opened 2018-11-12
EKF: Ensure yaw gets reset when declination is set

priseborough:
Fixes issue reported here: https://github.com/PX4/ecl/issues/519

**TESTING**

Testing performed in SITL with gazebo_gps_plugin.h modified to use following WGS-84 position

  // Seattle downtown (15 deg declination): 47.592182, -122.316031
   double lat_home = 47.592182 * M_PI / 180;    // rad
   double lon_home = -122.316031 * M_PI / 180;  // rad
   double alt_home = 86.0;                      // meters

and SDLOG_MODE = 1 to start logging from boot.

Before Changes Log: https://logs.px4.io/plot_app?log=479aba4f-d2f1-4fb3-b31d-8da8b4422f36

before

After Changes Log: https://logs.px4.io/plot_app?log=24a35542-4512-4342-b8cb-a49c3f6a3607

after

Because this does enforce reset the quaternion states for small angle changes less than 15 degrees that were previously ignored, flight testing is required to ensure that the filter stability and attitude controller handling of the resets is still OK. The alternative to doing this is to limit the scope of the small angle (<15 deg yaw change) quaternion state resets to on ground only, however this will lose the benefit of the more stable yaw gyro bias estimate following small yaw change in-air resets.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1356,6 +1356,13 @@ void Ekf::controlMagFusion()
 	// check for new magnetometer data that has fallen behind the fusion time horizon
 	// If we are using external vision data for heading then no magnetometer fusion is used
 	if (!_control_status.flags.ev_yaw && _mag_data_ready) {
+		// perform a yaw reset if requested by other functions
+		if (_mag_yaw_reset_req) {
+			if (!_mag_use_inhibit ) {
+				resetMagHeading(_mag_sample_delayed.mag);
+			}
+			_mag_yaw_reset_req = false;
+		}
 
 		// Determine if we should use simple magnetic heading fusion which works better when there are large external disturbances
 		// or the more accurate 3-axis fusion
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -545,6 +545,11 @@ bool Ekf::realignYawGPS()
 // Reset heading and magnetic field states
 bool Ekf::resetMagHeading(Vector3f &mag_init)
 {
+	// don't reseet twice on the same frame
+	if (_imu_sample_delayed.time_us == _flt_mag_align_start_time) {
+		return true;
+	}
+
 	if (_params.mag_fusion_type >= MAG_FUSE_TYPE_NONE) {
 		// do not use the magnetomer and deactivate magnetic field states
 		zeroRows(P, 16, 21);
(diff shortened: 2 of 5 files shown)


priseborough:
@dagar I've reworked as per your suggestion and retested. This time I changed the location to

  // Port-aux-Francais (-55 deg declination): -49.350330,  70.219162
  double lat_home = -49.350330 * M_PI / 180;  // rad
  double lon_home = 70.219162 * M_PI / 180;  // rad
  double alt_home = 10.0;                    // meters
 
to give a large -55 degree declination.

The EKF2_MAG_DECL parameter had to be zeroed after each test because the value was being saved for the next startup. 

To make the test more realistic I also modified gazebo_mavlink_interface.cpp to delay GPS 3D lock to 15 seconds from simulator start.

Log Before Changes: https://logs.px4.io/plot_app?log=61bb049a-c3d2-4519-b6df-2e23d6b90cba
before


Log After Changes: https://logs.px4.io/plot_app?log=d4586a72-42d4-4e94-9fc1-82762e70b71d
after

thomasgubler:
@PX4/testflights  @priseborough will create a branch and instruction how to test it

@PX4/testflights I could not add you as a reviewer, put please test this according to @priseborough's instructions

CarlOlsson:
I don't think this works as intended.

1) SITL testing in Seattle with EKF2_MAG_DECL = 0 deg and EKF2_MAG_DECL = 15 deg
image
As can be seen with EKF2_MAG_DECL = 0 deg the heading converges after ~40 sec, however, with EKF2_MAG_DECL = 15 deg (the true declination at the place of the test) the heading has not converged even after 2 minutes. 
This happens because of the build up of delta angle bias before GPS lock
image

2) The problem that the first call to resetMagHeading() don't use the EKF2_MAG_DECL parameter but the class variable default is still present.

3) For the logfiles uploaded above that was simulated at Port-aux-Francais something strange is going on with the groundtruth of vehicle attitude @dagar maybe you know why this happened?
image

4) I think it would be beneficial to agree on a startup sequence of the EKF before jumping into implementation, maybe in a call or here on github. E.g. what is your take on the actual benefit of initializing the EKF before GPS lock? (Given of course that GPS data is being pushed into the filter). Here is the same test as above but only initializing the EKF when lock and GPS checks pass
image
image

priseborough:
@CarlOlsson Initialising the EKF before GPS lock saves time by enabling the gyro bias states and attitude states to converge. It also enables users to takeoff without GPSand climb in ALTCTRL above obstructions to get a GPs lock.

I will go through your logs in more detail to see what is different to the tests I ran using the Port-aux-Francais location.

dagar:
What about calling `calcMagDeclination()` in `resetMagHeading()` to ensure the saved declination is used initially if available?

priseborough:
To avoid duplication I would add it to InitialiseFilter() above where resetMagHeading(...) is called.

priseborough:
@PX4/testflights I've created the https://github.com/PX4/Firmware/tree/test-ecl-pr520 branch for you to test this PR.

The test  procedure is to manually set the EKF2_MAG_DECL parameter to -15.0 and SDLOG_MODE = 1 just prior to power off. With the vehicle stationary outside, re-apply power and wait for EKF checks to pass, then perform a short flight, climbing above 5m with some position and yaw change manoeuvres in POSCTRL before landing, disarming and removing power. 

Repeat the process, but do not change the stored EKF2_MAG_DECL value because we want to verify the value obtained for that lat/lng was saved from the previous flight.

For reference, the NOAA website specifies the magnetic declination in Tijuana as 11.35° E  ± 0.33°

priseborough:
Bench testing using pixhawk 4 hardware and the initial value of EKF2_MAG_DECL offset by ~ -25 degrees:

screen shot 2018-11-14 at 9 10 13 am

CarlOlsson:
> Initialising the EKF before GPS lock saves time by enabling the gyro bias states and attitude states to converge.

Yes, I was questioning it because the plots I posted above showed the opposite since the gyro z bias took almost two minutes to converge. However, this was because of the bug that was fixed in your [last commit](389f1420576e3693f0d65fc70ca7398050497372). Now it seems to work, nice!

> For the logfiles uploaded above that was simulated at Port-aux-Francais something strange is going on with the groundtruth of vehicle attitude

When looking at your log files I thought it was only ground tests, there is actually no issue, sorry about that

EKF/ekf_helper.cpp
```diff
@@ -545,6 +545,11 @@ bool Ekf::realignYawGPS()
 // Reset heading and magnetic field states
 bool Ekf::resetMagHeading(Vector3f &mag_init)
 {
+	// don't reseet twice on the same frame
```
CarlOlsson:
```suggestion
	// don't reset twice on the same frame
```

EKF/control.cpp
```diff
@@ -1356,6 +1356,13 @@ void Ekf::controlMagFusion()
 	// check for new magnetometer data that has fallen behind the fusion time horizon
 	// If we are using external vision data for heading then no magnetometer fusion is used
 	if (!_control_status.flags.ev_yaw && _mag_data_ready) {
+		// perform a yaw reset if requested by other functions
+		if (_mag_yaw_reset_req) {
+			if (!_mag_use_inhibit ) {
```
dagar: I'm just curious, why is _mag_use_inhibit only checked in this reset case? There are quite a few other calls to reset (as I'm sure you know).

priseborough: That variable was introduced as part of the protection against indoor magnetic field errors and can only be true if _params.mag_fusion_type == MAG_FUSE_TYPE_INDOOR.

_mag_use_inhibit  is in used inside fuseHeading() to prevent a magnetic heading being used and causing either a static heading of 0 or a zero innovation to be used during heading fusion.

In theory a mag heading reset should not be called if _mag_use_inhibit  is true, however by introducing that constraint to the other resets that are not part of this fix we could be unintentionally introducing a bug due to a possible race condition when GPS or EV use is gained/lost. I wanted to keep the scope of this change minimal to reduce possibility of unintended consequences.

PR #524, opened 2018-11-21
EKF: use low pass filtered mag measurements for heading initialization

CarlOlsson:
in initialiseFilter() we low pass filter the measurements from the magnetometer, however when initializing the heading we took the last value from the mag instead of the filtered values. At least in SITL this sometimes results in up to 5 degrees heading error at bootup

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -591,7 +591,7 @@ bool Ekf::resetMagHeading(Vector3f &mag_init)
 
 		} else if (_params.mag_fusion_type <= MAG_FUSE_TYPE_AUTOFW) {
 			// rotate the magnetometer measurements into earth frame using a zero yaw angle
-			Vector3f mag_earth_pred = R_to_earth * _mag_sample_delayed.mag;
+			Vector3f mag_earth_pred = R_to_earth * mag_init;
 			// the angle of the projection onto the horizontal gives the yaw angle
 			euler321(2) = -atan2f(mag_earth_pred(1), mag_earth_pred(0)) + _mag_declination;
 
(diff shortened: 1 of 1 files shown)


CarlOlsson:
mag_init in this case is what is passed into the function so if I don't miss something it is indeed updated
https://github.com/PX4/ecl/blob/c49ab760c78fc6dc4cb174d6f9b7b39b01e19a7b/EKF/ekf_helper.cpp#L546-L546

In ```initialiseFilter()``` we call ```resetMagHeading``` with the low pass filtered magnetometer measurements, but in all other cases we call it with the last magnetometer measurements.

So this PR would solve the first issue mentioned [here](https://github.com/PX4/ecl/issues/525) that we should use the low pass filtered mag measurements for heading initialization, but not the fact that as soon as we get GPS lock we again reinitialize the heading without a low pass filter on the mag measurements.

priseborough:
OK, I think I now understand the problem and what you are proposing. I checked out this PR and following the code through it looks like there are no adverse effects and it solves the issue of noise affecting the initial alignment so I will approve and merge it once there is a test log. 

With respect to yawing and lag due to filtering, because we run the fusion on a delayed time horizon, by filtering the data going onto the buffer and using a filter that has a group delay equivalent to the buffer length, the effects of filter lag will be cancelled.

CarlOlsson:
> it looks like there are no adverse effects and it solves the issue of noise affecting the initial alignment so I will approve and merge it once there is a test log.

I don't have access to a system running standard PX4 at the moment, Is a sitl log enough?

> With respect to yawing and lag due to filtering, because we run the fusion on a delayed time horizon, by filtering the data going onto the buffer and using a filter that has a group delay equivalent to the buffer length, the effects of filter lag will be cancelled.

True, nice!

priseborough:
SITL will be OK

CarlOlsson:
[Here](https://logs.px4.io/plot_app?log=fa9d1f51-5530-483e-babc-22034372e8ea) is a sitl logfile with current master PX4 and this PR
image

Issue #527, opened 2018-11-28
airDataAiding always declared false without synthethic sideslip fusion


CarlOlsson:
If the EKF stops fusing GPS it will report that it is dead reckoning if it is [not fusing airspeed and synthetic sideslip](https://github.com/PX4/ecl/blob/acde4ebcc38692213ee475c392f68ecb0b853818/EKF/ekf_helper.cpp#L1321-L1321). This seems wrong since synthetic sideslip fusion can e.g. be turned off by setting EKF2_FUSE_BETA to 0.

One solution would be to only check for synthetic sideslip fusion if it is enabled. However, I am questioning the usefulness of synthetic sideslip fusion to constrain drift in the scenario where GPS is failing the innovation checks. So another solution would be to simply remove the sideslip check in the calculation of airDataAiding

CarlOlsson:
See https://github.com/PX4/ecl/pull/530 for Paul's explanation

PR #530, opened 2018-11-29
EKF: do not consider synthetic sideslip fusion as air data aiding

CarlOlsson:
Fixes https://github.com/PX4/ecl/issues/527

Before merging it would be good to get @priseborough opinion on the usefulness of synthetic sideslip fusion to constrain drift in the case when the velocity and position measurements are being rejected

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -1326,7 +1326,7 @@ void Ekf::update_deadreckoning_status()
 				|| (_time_last_imu - _time_last_vel_fuse <= _params.no_aid_timeout_max)
 				|| (_time_last_imu - _time_last_delpos_fuse <= _params.no_aid_timeout_max));
 	bool optFlowAiding = _control_status.flags.opt_flow && (_time_last_imu - _time_last_of_fuse <= _params.no_aid_timeout_max);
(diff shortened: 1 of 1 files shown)


priseborough:
The combination of airspeed and synthetic sideslip are effective in constraining velocity drift and enable fixed wing platforms to maintain a valid horizon reference indefinitely. The position does drift off at a fairly constant rate, but it buys time to perform an RTL back for a manual recovery.

priseborough:
If only airspeed is used, then the inertial nav will diverge in the cross track direction when flying straight and level.

CarlOlsson:
Okey thanks for the explanation!
Our issue is that after forward transition we don't commence synthetic sideslip fusion directly since during this time the sideslip can be extensive. If a heading error is present that can lead to the velocity and position states diverging and since sideslip fusion is not yet active the EKF reports the global position and velocity uncertainty based on the innovations, leading to activation of failsafe in commander. By removing the need for sideslip to be fused in this logic this would not happened, but will figure out something else (Y)

priseborough:
The fusion can tolerate a reasonable amount of sideslip transient up to 20 degrees because the default observation noise is set fairly high so I would try enabling it as soon as you have completed forward transition.  Inspecting Y axis 'g' in the flight log combined with airspeed should tell you how much sideslip angle.

PR #531, opened 2018-12-02
Optical flow: fixed sign in calculation of optical flow sensor velocity

RomanBapst:
The velocity of the optical flow sensor relative to the IMU expressed in
body frame is the cross product of the angular velocity with the vector
from IMU to the sensor. If we use the angular velocity stored in the flow
sample struct we need to use a negative sign as that angular velocity
follows the opposite sign convention.

I have not tested this yet on a vehicle! This in an attempt to fix estimation problems when using optical flow during motions in yaw.

### DIFF
--- a/EKF/optflow_fusion.cpp
+++ b/EKF/optflow_fusion.cpp
@@ -75,7 +75,8 @@ void Ekf::fuseOptFlow()
 	Vector3f pos_offset_body = _params.flow_pos_body - _params.imu_pos_body;
 
 	// calculate the velocity of the sensor reelative to the imu in body frame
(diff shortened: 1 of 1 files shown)


RomanBapst:
More context:
For flow sensor which do not provide a body angular velocity measurement _flow_sample_delayed.gyroXYZ
is calculated [here](https://github.com/PX4/ecl/blob/master/EKF/optflow_fusion.cpp#L578).

chuangniu:
@RomanBapst 
the place confused me. can you give some advice? need the following modifications?
(of_body_rate(0) - reference_body_rate(0)) -->
-(-of_body_rate(0) - reference_body_rate(0))-->
(of_body_rate(0) + reference_body_rate(0)) 
https://github.com/PX4/ecl/blob/4822bf83861513cf330ab1482439ed1c72e99232/EKF/optflow_fusion.cpp#L567-L573

PR #533, opened 2018-12-04
ekf: force 3D fusion for a short time right after takeoff

CarlOlsson:
Currently we reset the heading the first time we are above 1.5 meter altitude and declare the mag biases or the yaw angle observable. For a quad this is fine, but for a VTOL this can cause problems. If the vehicle ascends without changing orientation and then the forward transition finishes before the horizontal acceleration is above ```EKF2_MAG_ACCLIM```, then the heading and earth magnetic field states will be reset based on the assumption that the measured horizontal velocity is aligned with the vehicle heading. If this happens right after forward transition in a lot of sidewind it can cause the earth magnetic field states to be initialized to values far from reality.

E.g. here is the estimated magnetic declination from a flight in Zurich
image

This PR forces the use of 3D axis fusion for 2 seconds the first time the vehicle reaches 1.5 meter altitude. This makes sure that the earth magnetic field states are initialized based on the magnetometer as soon as the vehicle is away from magnetic interference close to the ground.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1350,6 +1350,7 @@ void Ekf::controlMagFusion()
 	if (!_control_status.flags.in_air) {
 		_last_on_ground_posD = _state.pos(2);
 		_flt_mag_align_complete = false;
+		_height_achieved_once = false;
 		_num_bad_flight_yaw_events = 0;
 	}
 
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1414,6 +1414,12 @@ void Ekf::controlMagFusion()
 				_time_last_movement = _imu_sample_delayed.time_us;
 			}
 
+			// Force 3D fusion after takeoff for a short time to make sure we reset the heading
+			if (height_achieved && !_height_achieved_once) {
+				_height_achieved_once = true;
+				_time_last_movement = _imu_sample_delayed.time_us;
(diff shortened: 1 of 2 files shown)


EKF/control.cpp
```diff
@@ -1414,6 +1414,12 @@ void Ekf::controlMagFusion()
 				_time_last_movement = _imu_sample_delayed.time_us;
 			}
 
+			// Force 3D fusion after takeoff for a short time to make sure we reset the heading
+			if (height_achieved && !_height_achieved_once) {
+				_height_achieved_once = true;
```
priseborough: I'm thinking about the use case where the vehicle lands, disarms, arms and takes off again. We should follow the same 3D fusion reset on subsequent takeoffs, so this flag will need to be reset when the EKF detects the vehicle is on ground.

CarlOlsson:
sitl logfile with two takeoffs is [here](https://logs.px4.io/plot_app?log=a9ab476d-d6ba-4737-8ad0-cb370f309e0b)

mhkabir:
@CarlOlsson Is this still relevant/an issue after the slew of magnetometer fixes on master?

CarlOlsson:
nope, this is now fixed with forcing a heading reset and earth mag state initialization at 1.5 m alt.

PR #535, opened 2018-12-09
Pr tecs front trans

RomanBapst:
This PR adds a mode to TECS for assisting in VTOL (standard & tiltrotor) front transitions.
The idea is that TECS can already during the front transition start to control altitude using pitch.
The transition happens in non-airspeed control mode and throttle is ramped up in an open loop manner.
This concept has been tested (on an earlier branch) both in simulation and on a vehicle (delta quad).
It seemed to improve the altitude hold performance during transitions without any previous hacks and also allows a seamless transition to pure fixed wing flight (no jumps in pitch or throttle setpoint).
Obviously this is unusable for tailsitters.

I will provide logs of simulation and real flights which demonstrate the functionality of this PR.
Although this has not been necessary until now we may need to add special constraints on the pitch setpoint in positive direction as this could lead to unwanted drag created by the multirotor lift system.


LorenzMeier:
Nice!

RomanBapst:
Active control of altitude via the multirotor controller has shown to give good altitude control during the transition.
We might come back to this at some point if we feel it's still needed.

PR #538, opened 2018-12-17
Underspeed protection fix

almaaro:
Changed the criteria for underspeed detection: 
- Ignored the throttle setting. Underspeed condition can occur even at low throttle values. Why? If the set airspeed is close to the minimum speed, it is possible that the P term in the throttle PID doesn't increase the throttle enough for it to reach 0.95 * max throttle.

Also changed the throttle setpoint in underspeed conditions from 100% to the user defined max value. No more burnt ESCs.

### DIFF
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -221,6 +221,11 @@ void TECS::_update_speed_setpoint()
 
 void TECS::_update_height_setpoint(float desired, float state)
 {
+        // Don't try to change the altitude if underspeed.
+        if (_underspeed_detected) {
+                desired = state;
+        }
+
 	// Detect first time through and initialize previous value to demand
 	if (ISFINITE(desired) && fabsf(_hgt_setpoint_in_prev) < 0.1f) {
 		_hgt_setpoint_in_prev = desired;
@@ -316,7 +321,7 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 	// Calculate the throttle demand
 	if (_underspeed_detected) {
 		// always use full throttle to recover from an underspeed condition
-		_throttle_setpoint = 1.0f;
+                _throttle_setpoint = _throttle_setpoint_max;
 
 	} else {
 		// Adjust the demanded total energy rate to compensate for induced drag rise in turns.
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -272,7 +272,7 @@ void TECS::_detect_underspeed()
 		return;
 	}
 
-        if (_tas_state < _TAS_min || ((_STE_error > 0) && _underspeed_detected)) {
+        if (_tas_state < _TAS_min * 0.9f || ((_vert_pos_state < _hgt_setpoint_adj) && _underspeed_detected)) {
 
 		_underspeed_detected = true;
 
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -314,7 +314,7 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 
 	// Calculate the throttle demand
 	if (_underspeed_detected) {
-		// always use full throttle to recover from an underspeed condition
+		// always use max throttle to recover from an underspeed condition
 		_throttle_setpoint = _throttle_setpoint_max;
 
 	} else {
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -272,7 +272,7 @@ void TECS::_detect_underspeed()
 		return;
 	}
 
-	if (_tas_state < _TAS_min * 0.9f || ((_vert_pos_state < _hgt_setpoint_adj) && _underspeed_detected)) {
+	if (_tas_state < _TAS_min || ((_vert_pos_state < _hgt_setpoint_adj) && _underspeed_detected)) {
 
 		_underspeed_detected = true;
 
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -201,7 +201,7 @@ void TECS::_update_speed_setpoint()
 	// Set the airspeed demand to the minimum value if an underspeed or
 	// or a uncontrolled descent condition exists to maximise climb rate
 	if ((_uncommanded_descent_recovery) || (_underspeed_detected)) {
-		_TAS_setpoint = _TAS_min;
+		_TAS_setpoint = 1.1f * _TAS_min;
 	}
 
 	_TAS_setpoint = constrain(_TAS_setpoint, _TAS_min, _TAS_max);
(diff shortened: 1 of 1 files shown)


Antiheavy:
This PR introduces the idea of stall speed to PX4 (COM_ASPD_STALL): https://github.com/PX4/Firmware/pull/10733/files#diff-1b2f00d02614bed0dbff3dca792a1f1cR119   I wonder if using a stall speed setting would be more appropriate for underspeed protection vs a minimum airspeed (derived from FW_AIRSPD_MIN)?

tecs/tecs.cpp
```diff
@@ -316,7 +315,7 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 	// Calculate the throttle demand
 	if (_underspeed_detected) {
 		// always use full throttle to recover from an underspeed condition
```
Antiheavy: if this change is accepted, you might want to update the comment to read something like "//always use max throttle to recover from an underspeed condition"

Antiheavy:
> Changed the criteria for underspeed detection:
> 
>     * Ignored the throttle setting. Underspeed condition can occur even at low throttle values. Why? If the set airspeed is close to the minimum speed, it is possible that the P term in the throttle PID doesn't increase the throttle enough for it to reach 0.95 * max throttle.
> 
> 
> Also changed the throttle setpoint in underspeed conditions from 100% to the user defined max value. No more burnt ESCs.

You'd have to ask the original author, but one explanation for why use 100% throttle (vs throttle max) is that the " _throttle_setpoint_max * 0.95f" term already the system is already trying very hard to recover from the underspeed condition and it might need an extra boost to prevent a dangerous stall....  maybe?

Antiheavy:
> Also changed the throttle setpoint in underspeed conditions from 100% to the user defined max value. No more burnt ESCs.

...now you have me thinking about the role of the PX4 Mixer file vs the FW_THR_MAX parameter.  If an ESC would burn out from a short duration full throttle burst, should it be limited in the Mixer?  There may not be a "correct" answer to this question.  This also has me thinking about if FW_THR_MAX should scale with battery voltage, but that is a totally different topic...  I'm not sure if these comments are directly relevant to this PR, but maybe speak to the bigger picture.

tecs/tecs.cpp
almaaro: changed this

almaaro:
> This PR introduces the idea of stall speed to PX4 (COM_ASPD_STALL): https://github.com/PX4/Firmware/pull/10733/files#diff-1b2f00d02614bed0dbff3dca792a1f1cR119 I wonder if using a stall speed setting would be more appropriate for underspeed protection vs a minimum airspeed (derived from FW_AIRSPD_MIN)? e.g. instead of 10% below minimum speed, maybe trigger with 10% above stall speed?

I agree that this would be a better solution as it would be more explicitly defined.

> You'd have to ask the original author, but one explanation for why use 100% throttle (vs throttle max) is that the " _throttle_setpoint_max * 0.95f" term implies the system is already trying very hard to recover from the underspeed condition and it might need an extra boost to prevent a dangerous stall.... maybe?

There is a point. Maybe there could be 2 underspeed cases: one where the throttle will exceed the throttle_max value (the current master) and one where it will set the throttle immediately to throttle_max when the airspeed drops dangerously low. 

> ...now you have me thinking about the role of the PX4 Mixer file vs the FW_THR_MAX parameter. If an ESC would burn out from a short duration full throttle burst, should it be limited in the Mixer? There may not be a "correct" answer to this question. This also has me thinking about if FW_THR_MAX should scale with battery voltage, but that is a totally different topic... I'm not sure if these comments are directly relevant to this PR, but maybe speak to the bigger picture.

The mixer file is also a valid point (just a bit less user-friendly). And by the way, FW_THR_MAX scales with the battery voltage. 

Here is my suggestion: 
- Detecting the underspeed condition when the airspeed drops dangerously low (defined by COM_ASPD_STALL as soon as it gets implemented). 
- Not taking the throttle setting into account. 
- The absolute maximum throttle will be limited by the Mixer, tecs.cpp will read "_throttle_setpoint = 1.0f;". 
- FW_THR_MAX will just be the throttle limit under normal operating conditions.

RomanBapst:
@almaaro Regarding underspeed detection: Do you have a log where we can see the underspeed detection not triggering when it should have? I do not see a problem removing the throttle condition as long as we don't increase the number of false positives on underspeed detection drastically.
A couple of test flights on real vehicles are required.

almaaro:
Here are 2 cases where it didn't seem to act. Minimum airspeed was 7.5 m/s -> threshold 6.75 m/s. Maximum throttle was 54%.
underspeed1
us2

Here is a screenshot with the underspeed protection in action (max throttle is 54%). However, the code alters from the pull request a bit in this case. In this case the if clause is 

`if (_tas_state < _TAS_min || ((_STE_error > 0) && _underspeed_detected))`

not like this as in the pull request:
`if (_tas_state < _TAS_min * 0.9f || ((_vert_pos_state < _hgt_setpoint_adj) && _underspeed_detected)`

...Which actually gets us to another topic; whether we should exit the underspeed condition already when the STE is enough to get us to the desired altitude at minimum airspeed (reducing oveshoots)?.
us3-act

Antiheavy:
@dagar how do you think this lines up with the "review groundspeed undershoot" issue we discussed a year ago?

Antiheavy:
@dagar how do you think the changes in https://github.com/PX4/Firmware/pull/11142 impact this old issue?  Is there still some area needing improvement?  I'm happy to create an upstream issue if we think there is something specific we can do.

tecs/tecs.cpp
```diff
@@ -315,8 +314,8 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 
 	// Calculate the throttle demand
 	if (_underspeed_detected) {
-		// always use full throttle to recover from an underspeed condition
-		_throttle_setpoint = 1.0f;
+		// always use max throttle to recover from an underspeed condition
+		_throttle_setpoint = _throttle_setpoint_max;
```
dagar: @almaaro maybe we should break this part off and merge immediately?

almaaro: Sure. However, I'd like to keep this PR still open in order to address the throttle requirement for an underspeed condition (that currently the thr has to be over 95% of the max value regardless of how low the airspeed is).

LorenzMeier:
Thank you so much for taking the time to make this contribution! To make the review process efficient and fast, could you please rebase or cherry-pick these commits on top of master? It is very hard to read and not mergable in this state. Thanks!

priseborough:
Closing because request for testing to determine impact on users with conservative minimum airspeed value remained unanswered.

PR #540, opened 2018-12-18
TECS to take airspeed into account in throttle calculation

almaaro:
Edit. Updated the section describing the relationship between throttle and V2 (not assuming that V2 is proportional to throttle but rather that V2-V0 is.)
Hello,

**The current situation**
Throttle is calculated from STE rate setpoint and it's minimum and maximum limits at the default airspeed.

**The new situation**
Making these assumptions:
- We know the vehicle's mass, wingspan and propeller/fan diameter
- FW_AIRSPD_MAX is the airspeed that is achieved at level flight with the maximum throttle
- Thrust = 0.5 * A * rho*(V2² - V0²), where V2 is the air velocity after the propeller and V0 is the vehicle's airspeed. Source: https://www.grc.nasa.gov/www/k-12/airplane/propth.html
- If the throttle setting is kept constant, the thrust will reduce linearly as the airspeed increases (above a certain advance ratio, I _believe_ it is because the propeller will stall at very low advance ratios, but I am not sure). Source: https://www.researchgate.net/figure/Typical-propeller-thrust-curves-as-a-function-of-advance-ratio-J_fig4_281946347. We can calculate the slope if we know the thrust required at 2 different airspeeds ( from the maximum airspeed at level flight and from the maximum climb rate at trim airspeed).
- The induced and parasitic drag can be calculated at any airspeed if we know STE_rate_min flown at FW_AIRSPD_TRIM, because STE_rate_min = (induced drag + parasitic drag) * airspeed.
- The induced drag coefficient can be calculated from airspeed, lift and wingspan. The parasitic drag coefficient can be calculated at a known airspeed if we know the total drag and induced drag. Source: https://en.wikipedia.org/wiki/Lift-induced_drag#Calculation_of_induced_drag and  https://en.wikipedia.org/wiki/Parasitic_drag#Description
- The value for delta V (=V2-V0) at maximum throttle can be calculated for any airspeed using the linear relationship between thrust and airspeed (by solving V2 from the thrust equation)
- The throttle setting controls the delta V rather than V2. **This is not proven by any means, it just seems to give good results. It won't work if the ESC has a governor.**
- The shape of the (throttle, delta V) curve remains the same at all airspeeds. It will just be scaled taller or shorter according to the maximum delta V at the current airspeed. The shape can be approximated by 3 rules:
    - When throttle is zero, delta V is zero. A rough approximation, but seems to give ok results. In reality, the propeller probably slows the air to give a _negative_ delta V.
    - At throttle_cruise, delta V can be solved from the thrust equation as we know the required thrust to sustain level flight.
    - At maximum throttle, delta V can be solved as we know the thrust required to sustain level flight and increase the potential energy at the maximum rate.
    - The required throttle will be solved from the required delta V  by interpolating the delta V between minimum, cruise and maximum throttle.

Throttle calculation:
- Calculating the induced and parasitic drag
- Adjusting the (throttle, delta V) curve to the current airspeed
- As we know the required thrust to overcome the aerodynamic drag and to climb/descend and the airspeed, we can calculate the required thrust and V2 and delta V.
- Then we just solve the throttle setting to achieve the required delta V.

So that's about it. Here is an excel sheet where you can try how the formulas work.
[thr-airspeed calc 2-2.xlsx](https://github.com/PX4/ecl/files/2729398/thr-airspeed.calc.2-2.xlsx)
(rev. 5 Jan 2018)

**Other solutions**
Another solution would be to just get some additional throttle settings for climbs/descends at various airspeeds as parameters. This would be very easy and robust.

The Firmware-side: (to set AUW, wingspan, propeller diameter and trimmed airspeed) https://github.com/almaaro/Firmware/tree/advanced-ste-limits

I hope this gets some conversation going.
Regards, 
almaaro


tecs/tecs.cpp
```diff
@@ -208,8 +208,8 @@ void TECS::_update_speed_setpoint()
 
 	// Calculate limits for the demanded rate of change of speed based on physical performance limits
 	// with a 50% margin to allow the total energy controller to correct for errors.
-	float velRateMax = 0.5f * _STE_rate_max / _tas_state;
-	float velRateMin = 0.5f * _STE_rate_min / _tas_state;
+        float velRateMax = 0.5f * _STE_rate_max / _tas_state;
```
dagar: Could you revert these whitespace changes?

almaaro: Done, thanks for pointing that out.

tecs/tecs.cpp
```diff
@@ -573,10 +573,34 @@ void TECS::_initialize_states(float pitch, float throttle_cruise, float baro_alt
...
 	// Calculate the specific total energy lower rate limits from the min throttle sink rate
-	_STE_rate_min = - _min_sink_rate * CONSTANTS_ONE_G;
+	const float rate_min = - _min_sink_rate * CONSTANTS_ONE_G;
+
+	// Cd_i_specific = ... assuming planar wing with elliptical lift distribution
+	_Cd_i_specific = _auw * CONSTANTS_ONE_G; //lift
+	_Cd_i_specific = _Cd_i_specific * _Cd_i_specific / (0.5f * M_PI_F * CONSTANTS_AIR_DENSITY_SEA_LEVEL_15C * _wingspan * _wingspan);
+
+	// Cd_o_specific: subtracting induced drag from total drag at a known airspeed to calculate parasitic drag
+	_Cd_o_specific = (-rate_min - _Cd_i_specific / _indicated_airspeed_trim) / (_indicated_airspeed_trim * _indicated_airspeed_trim * _indicated_airspeed_trim);
```
dagar: Can these variables (_Cd_i_specific,  _Cd_o_specific, etc) be kept out of the class? I only see them being calculated and used in _update_STE_rate_lim().

What I'd really like to do in TECS is cleanup the unnecessary state between calls.

almaaro: Thanks for pointing that out.

tecs/tecs.cpp
```diff
@@ -573,10 +573,38 @@ void TECS::_initialize_states(float pitch, float throttle_cruise, float baro_alt
...
+
+	const float lift = _auw * CONSTANTS_ONE_G;
+
+	// _Cd_i_specific: Vehicle specific induced drag coefficient, which equals to 1/2*S*rho*Cd_i
+	// Cd_i_specific = ... assuming planar wing with elliptical lift distribution
+	const float _Cd_i_specific = lift * lift / (0.5f * M_PI_F * CONSTANTS_AIR_DENSITY_SEA_LEVEL_15C * _wingspan * _wingspan);
+
+	// _Cd_o_specific: Vehicle specific parasitic drag coefficient, which equals to 1/2*A*rho*Cd_o
+	// Cd_o_specific: subtracting induced drag from total drag at a known airspeed to calculate parasitic drag
+	const float _Cd_o_specific = (-rate_min - _Cd_i_specific / _indicated_airspeed_trim) / (_indicated_airspeed_trim * _indicated_airspeed_trim * _indicated_airspeed_trim);
```
RomanBapst: @almaaro Isn't there a unit mismatch here? rate_min is an energy rate (or power) and has the unit m/s * m/s^2
_Cd_i_specific / _indicated_airspeed_trim according to your previous calculations has the unit kg*m^3 / s^4

almaaro: Definitely a point there. Here's what I came up with (bold for units that cancel out):
[_Cd_i_specific] = kg * m/s² * kg * m/s² * 1/(kg/m³) * 1/m * 1/m = kg * **m**/s² * kg * **m**/s² * 1/(kg/m³) * 1/**m** * 1/**m** = **kg**/s² * kg/s² * 1/(**kg**/m³) = kg * m³/s⁴

that divided by [_indicated_airspeed_trim] (m/s) gives kg*m²/s³.

So there is the extra kilogram. Let's see.

The starting equation:
`_min_sink_rate * CONSTANTS_ONE_G * _auw = _Cd_i_specific / EAS + _Cd_o_specific * EAS^3`
since
`rate_min = _min_sink_rate * CONSTANTS_ONE_G`
the `_Cd_i_specific` and `_Cd_o_specific` should be divided by _auw. But since `_Cd_o_specific` is calculated inderectly from `rate_min` and `_Cd_i_specific`, there is nothing to be divided while calculating it. 

So it should be
`const float _Cd_i_specific = lift * lift / (0.5f * M_PI_F * CONSTANTS_AIR_DENSITY_SEA_LEVEL_15C * _wingspan * _wingspan * _auw);`

@RomanBapst do you think this is valid?

I'll also change the underscoring to match local variable conventions.

almaaro:
I have some improvements for this pull request to come. I'll upload them when I get some test data. It will include taking the effect of airspeed to thrust into account. Until these changes, this PR is kind of outdated.

priseborough:
Can you explain the rationale for doing this? Besides the challenges involved with estimating drag polars, It seems to be too propulsion system specific. Although the majority of users use fixed pitch props and electric motors, this is not universal. Some of the biggest variations from a tuning perspective are due to density altitude and battery voltage.

almaaro:
So it would just be for a better control of the airspeed, just because I would like that. (I also added new throttle control algorithms to this branch after closing this pr). But I think that this is a big improvement compared to the original one, where the throttle open loop control was valid for only the trim airspeed. Now, it works at least a little bit better for other airspeeds, too. 

Here is some test data from X-Plane for starters. I tested this with the C172, Cessna 400 and Baron 58. I think that it tells that the required throttle setting is different at different airspeeds and that this new model for estimating the predicted throttle is actually pretty decent:

The first 2 colums of numbers contain the test data, the 3rd is for the estimation. The rows with the green description are the ones needed to calculate the estimates. The rows with no color in description are for validation.
image

priseborough:
To be representative of the most common use case, the equations should be based on calculations done assuming electric propulsion with fixed motor voltage and a fixed pitch prop. With these systems,  shaft and thrust power reduces with airspeed because the propeller RPM increases and back EMF reduces motor current and therefore torque. 

Also the modelling of drag is problematic. A better user facing parameterisation would be wing span, wing area, weight and measured sink rate at the trim airspeed as these are things a user can directly measure. Cdi and Cdo are not. 

The weight and wing area give the CL value. The span and area are used to calculate the aspect ratio. Aspect ratio, and span efficiency factor are then used to calculate the induced drag factor. A span efficiency factor of 0.85 would be  reasonable default. The drag polar equation can then be used to solve for CDo.

For the propulsion model, a propeller pitch and diameter parameter would suffice and the thrust available calculation should take the reduction in motor power with RPM (at a constant input voltage) into consideration and also altitude density effect.

almaaro:
I'll reopen this PR since the content is now updated. I also added the intro text from https://github.com/PX4/ecl/issues/546.

> To be representative of the most common use case, the equations should be based on calculations done assuming electric propulsion with fixed motor voltage and a fixed pitch prop. With these systems, shaft and thrust power reduces with airspeed because the propeller RPM increases and back EMF reduces motor current and therefore torque.

Yeah, so the current model isn't taking this into account. I couldn't come up with a solution for this and thought that the effect is not very large. I just assumed a linear relationship with throttle and airspeed. Source: https://www.electricrcaircraftguy.com/2014/04/propeller-static-dynamic-thrust-equation-background.html (part "Application & Conjecturing")

> Also the modelling of drag is problematic. A better user facing parameterisation would be wing span, wing area, weight and measured sink rate at the trim airspeed as these are things a user can directly measure. Cdi and Cdo are not.

The current model uses the wingspan, weight and the sink rate to calculate Cdi and Cdo -like coefficients (actually they are induced cd_i_specific = Di*V² and parasitic cd_o_specific =  Do/V² -> total drag = cd_i_specific/V² + cd_o_specific*V²) Source for induced drag calculation: https://en.wikipedia.org/wiki/Lift-induced_drag#Calculation_of_induced_drag

> The weight and wing area give the CL value. The span and area are used to calculate the aspect ratio. Aspect ratio, and span efficiency factor are then used to calculate the induced drag factor. A span efficiency factor of 0.85 would be reasonable default. The drag polar equation can then be used to solve for CDo.
The efficiency factor is missing. Adding to TODO list.

> For the propulsion model, a propeller pitch and diameter parameter would suffice and the thrust available calculation should take the reduction in motor power with RPM (at a constant input voltage) into consideration and also altitude density effect.

The propulsion model is slightly different in https://github.com/PX4/ecl/issues/546. It tries to estimate the trust by approximating how much the propeller will change the air's momentum. It does take the diameter as parameter, but calculates the pitch implicitly. It is not taking the density of air into account, but It could (TODO).

LorenzMeier:
It sounds like this will be very hard to maintain as these quantities are generally not known for small drones. We will likely not take this change for the sake of generality.

priseborough:
Also this is too specific to a particular type of propulsion method. Moving that part of the calculation into a separate module that converts thrust to throttle and enables different propulsion methods is required. The use of drag polars as Lorenz points out will not be useful to most users, so such a feature would have to be made optional.

almaaro:
So does this mean that PX4 only supprorts the default (trim) airspeed?

sfuhrer:
FYI variable airspeed is now supported through https://github.com/PX4/PX4-Autopilot/pull/21345. It assumes a linear dependency of throttle to airspeed setpoint, which showed to be accurate enough on the planes we tested with. It's thus very simple to tune (operator can define a airspeed to throttle mapping with different slope angles below/above trim airspeed by setting the airspeed/throttle pairs at MIN, TRIM, MAX airspeed.

PR #542, opened 2018-12-21
Updated pitch and throttle control algorithms

almaaro:
**Major changes to:**

1. Throttle integrator: only adding to the integrator if the absolute STE error is increasing. This is to prevent overshoots when the airspeed setpoint changes. 
In the log below this happens at T=41s: The integrator would normally start decreasing the throttle, but as we can see at T=45s, the real need is to actually increase the integrator value. 

     Also, limiting the integrator term from becoming very large in eg. quick descends where the airspeed 
     will build up even if the throttle is set to 0 (similar decaying system as with the pitch integrator). 

2. Replacing FW_T_INTEG_GAIN with FW_T_INTEG_GAIN_P (for pitch control) and FW_T_INTEG_GAIN_T for throttle control. This is for more accurate control of the gains.

3. Limiting SKE_weighting and SPE_weighting to the maximum of 1 instead of 2. The calculated control variables (for example the _SPE_rate_setpoint from _hgt_rate_setpoint) are directly proportional to these values and I find it more intuitive to increase the PID parameters instead of hiddenly scaling up the errors. This is also why the _hgt_rate_setpoint will be limited to 2 times the maximum/minimum climb and sink rates.

4. Changes to the throttle setpoint calculation: Assuming that the required throttle is proportional to the square root of the desired  STE rate. Source: https://www.grc.nasa.gov/www/k-12/airplane/propanl.html . Edit: There might be better ways to estimate the required `throttle_cruise` setting at different airspeeds, need to have a look at them.  

5. Not using STE_rate_setpoint as the starting point of throttle setpoint calculation, but instead using the SKE and SKE rate errors for proportional and derivative control and the _SPE_rate_setpoint_pitch derived from the part of the pitch setpoint that is controlling the SPE rate (as defined by the values of SKE_weighting and SPE_weighting). This gives a much sharper throttle response eg. when starting a climb as the throttle will increase with the pitch setpoint.
This can be seen in the log below at T=33s when the throttle is increased radically as the pitch setpoint is increased.  

Log with the above changes (has been tested >10 flights):
image

**Other Considerations**
1. Replacing/removing FW_T_INTEG_GAIN means no backwards compatibility.
2. The parameter FW_THR_CRUISE will be obsolete. This can have effects elsewhere, such as in FW autoland where the flare throttle is limited by reducing the cruise throttle.
3. The way of calculating the throttle setpoint with the square root of demanded STE rate is not any scientific method, but seems to work ok with various airspeeds. In my test setup.
4. This has not been tested without an airspeed sensor.
5. This branch has a lot of change, should this be divided?

**Other context**
The commit having the Firmware changes: https://github.com/PX4/Firmware/commit/e597fb453ef80da82c6fcc31e6e3ae6236e47304

### DIFF
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -346,7 +346,7 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 
 		// Add proportional and derivative control feedback to kinetic energy error.
 		// Question: Should this be compensated for the pitch-driven kinetic energy control?
-		float SKE_feedback = (SKE_error + SKE_rate_error * _throttle_damping_gain) * STE_to_throttle;
+		float SKE_feedback = (SKE_error + _SKE_rate_error * _throttle_damping_gain) * STE_to_throttle;
 
 		// Calculate throttle and constrain to throttle limits.
 		_throttle_setpoint = sqrtf(constrain(turn_compensation + _SPE_rate_setpoint_pitch + SKE_feedback - _STE_rate_min, 0.01f, _STE_rate_max - _STE_rate_min)
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -257,12 +257,8 @@ void TECS::_update_height_setpoint(float desired, float state)
 	_hgt_setpoint_adj_prev = _hgt_setpoint_adj;
 
 	// Limit the rate of change of height demand to respect vehicle performance limits
-	if (_hgt_rate_setpoint > _max_climb_rate) {
-		_hgt_rate_setpoint = _max_climb_rate;
-
-	} else if (_hgt_rate_setpoint < -_max_sink_rate) {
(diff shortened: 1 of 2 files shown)


Antiheavy:
Some great ideas here!

RomanBapst:
@almaaro Thanks for your contributions! I have some general comments:
1) It would be good if you attach the link to a log file which demonstrates the effect of your changes. Preferably, you would show the before and after scenario.
2) You have a couple of changes in this pull request, I think it's better if you split this up into multiple pull request and it's quite hard otherwise to judge which of your changes had what effect.
3) 

> Assuming that the required throttle is proportional to the square root of the desired STE rate
If you make assumptions you should state why you have made those assumptions? Are they based on measured data or did you just try this by chance?

almaaro:
Ok, I could make this into like 3 different pull requests. It may take some time until I get all the test data gathered, but I'd love to hear some general comments about the code before that also, if you come up with any.

PR #543, opened 2018-12-23
ekf_helper: add more useful methods to interface with the covariances

TSC21:
This PR aims to improve the interface with the covariances obtained by EKF. It extends far beyond the the existent method `get_covariances()`, by also adding ways of obtaining dimensions covariances.
Allows:
- obtain the full state covariances matrix;
- obtain the position, the quaternion and the velocity covariance;
- by implementing the covariances law, one is now able to obtain the 3x3 covariance from the euler anges, which is useful for the current odometry pipeline implemented on PX4;
- merges the position and orientation (euler) covariances into a single pose covariance, which not being a cross-covariance matrix, it helps at least to parse it to the PX4 internals.

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -122,8 +122,28 @@ public:
 	// get the true airspeed in m/s
 	void get_true_airspeed(float *tas);
 
-	// get the diagonal elements of the covariance matrix
+	void propagate_covariances_from_quat_to_euler(matrix::SquareMatrix<float, 3> &euler_cov);
+
+	// get the full covariance matrix
 	void get_covariances(float *covariances);
+
+	// get the diagonal elements of the covariance matrix
+	void get_covariances_diagonal(float *covariances);
+
+	// get the position covariances
+	void get_position_covariances(float *covariances);
+
+	// get the quaternion covariances
+	void get_quaternion_covariances(float *covariances);
+
+	// get the euler angles covariances
+	void get_euler_covariances(float *covariances);
+
+	// get the pose covariances (position + orientation in euler angles)
+	void get_pose_covariances(float *covariances);
+
+	// get the linear velocity covariances
+	void get_velocity_covariances(float *covariances);
 
 	// ask estimator for sensor data collection decision and do any preprocessing if required, returns true if not defined
 	bool collect_gps(const gps_message &gps);
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -123,19 +123,19 @@ public:
 	void get_true_airspeed(float *tas);
 
 	// get the full covariance matrix
-	const matrix::SquareMatrix<float, 24> covariances() const;
+	matrix::SquareMatrix<float, 24> covariances() const;
 
 	// get the diagonal elements of the covariance matrix
-	const matrix::Vector<float, 24> covariances_diagonal() const;
+	matrix::Vector<float, 24> covariances_diagonal() const;
 
 	// get the position covariances
-	const matrix::SquareMatrix<float, 3> position_covariances() const;
+	matrix::SquareMatrix<float, 3> position_covariances() const;
 
 	// get the orientation (quaternion) covariances
-	const matrix::SquareMatrix<float, 4> orientation_covariances() const;
+	matrix::SquareMatrix<float, 4> orientation_covariances() const;
 
 	// get the linear velocity covariances
-	const matrix::SquareMatrix<float, 3> velocity_covariances() const;
+	matrix::SquareMatrix<float, 3> velocity_covariances() const;
 
 	// ask estimator for sensor data collection decision and do any preprocessing if required, returns true if not defined
 	bool collect_gps(const gps_message &gps);
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -123,19 +123,19 @@ public:
 	void get_true_airspeed(float *tas);
 
 	// get the full covariance matrix
-	matrix::SquareMatrix<float, 24> covariances() const;
+	matrix::SquareMatrix<float, 24> covariances() const { return matrix::SquareMatrix<float, _k_num_states>(P); }
 
 	// get the diagonal elements of the covariance matrix
-	matrix::Vector<float, 24> covariances_diagonal() const;
+	matrix::Vector<float, 24> covariances_diagonal() const { return covariances().diag(); }
 
-	// get the position covariances
-	matrix::SquareMatrix<float, 3> position_covariances() const;
-
-	// get the orientation (quaternion) covariances
-	matrix::SquareMatrix<float, 4> orientation_covariances() const;
+	// get the orientation (quaterion) covariances
+	matrix::SquareMatrix<float, 4> orientation_covariances() const { return covariances().slice<4, 4>(0, 0); }
 
 	// get the linear velocity covariances
-	matrix::SquareMatrix<float, 3> velocity_covariances() const;
+	matrix::SquareMatrix<float, 3> velocity_covariances() const { return covariances().slice<3, 3>(4, 4); }
+
+	// get the position covariances
+	matrix::SquareMatrix<float, 3> position_covariances() const { return covariances().slice<3, 3>(7, 7); }
 
 	// ask estimator for sensor data collection decision and do any preprocessing if required, returns true if not defined
 	bool collect_gps(const gps_message &gps);
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -122,9 +122,6 @@ public:
 	// get the true airspeed in m/s
 	void get_true_airspeed(float *tas);
 
-	// covariance propagation from quaternions to euler angles using the covariance law
-	const matrix::SquareMatrix<float, 3> propagate_covariances_from_quat_to_euler() const;
-
 	// get the full covariance matrix
 	const matrix::SquareMatrix<float, 24> covariances() const;
 
@@ -136,12 +133,6 @@ public:
 
 	// get the quaternion covariances
 	const matrix::SquareMatrix<float, 4> quaternion_covariances() const;
-
-	// get the euler angles covariances
-	const matrix::SquareMatrix<float, 3> euler_covariances() const;
-
-	// get the pose covariances (position + orientation in euler angles)
-	const matrix::SquareMatrix<float, 6> pose_covariances() const;
 
 	// get the linear velocity covariances
 	const matrix::SquareMatrix<float, 3> velocity_covariances() const;
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -49,7 +49,7 @@ class Ekf : public EstimatorInterface
 public:
 
 	Ekf() = default;
-	~Ekf() = default;
+	virtual ~Ekf() = default;
 
 	// initialise variables to sane values (also interface class)
 	bool init(uint64_t timestamp);
@@ -122,28 +122,29 @@ public:
 	// get the true airspeed in m/s
 	void get_true_airspeed(float *tas);
 
-	void propagate_covariances_from_quat_to_euler(matrix::SquareMatrix<float, 3> &euler_cov);
+	// covariance propagation from quaternions to euler angles using the covariance law
+	const matrix::SquareMatrix<float, 3> propagate_covariances_from_quat_to_euler() const;
 
 	// get the full covariance matrix
-	void get_covariances(float *covariances);
+	const matrix::SquareMatrix<float, 24> covariances() const;
 
 	// get the diagonal elements of the covariance matrix
-	void get_covariances_diagonal(float *covariances);
+	const matrix::Vector<float, 24> covariances_diagonal() const;
 
 	// get the position covariances
-	void get_position_covariances(float *covariances);
+	const matrix::SquareMatrix<float, 3> position_covariances() const;
 
 	// get the quaternion covariances
-	void get_quaternion_covariances(float *covariances);
+	const matrix::SquareMatrix<float, 4> quaternion_covariances() const;
 
 	// get the euler angles covariances
-	void get_euler_covariances(float *covariances);
+	const matrix::SquareMatrix<float, 3> euler_covariances() const;
 
 	// get the pose covariances (position + orientation in euler angles)
-	void get_pose_covariances(float *covariances);
+	const matrix::SquareMatrix<float, 6> pose_covariances() const;
 
 	// get the linear velocity covariances
-	void get_velocity_covariances(float *covariances);
+	const matrix::SquareMatrix<float, 3> velocity_covariances() const;
 
 	// ask estimator for sensor data collection decision and do any preprocessing if required, returns true if not defined
 	bool collect_gps(const gps_message &gps);
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -131,8 +131,8 @@ public:
 	// get the position covariances
 	const matrix::SquareMatrix<float, 3> position_covariances() const;
 
-	// get the quaternion covariances
-	const matrix::SquareMatrix<float, 4> quaternion_covariances() const;
+	// get the orientation (quaternion) covariances
+	const matrix::SquareMatrix<float, 4> orientation_covariances() const;
 
 	// get the linear velocity covariances
 	const matrix::SquareMatrix<float, 3> velocity_covariances() const;
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -982,12 +982,125 @@ void Ekf::get_gyro_bias(float bias[3])
 	memcpy(bias, temp, 3 * sizeof(float));
 }
 
-// get the diagonal elements of the covariance matrix
+// covariance propagation from quaternions to euler angles using the covariance law
+// source: "Development of a Real-Time Attitude System Using a Quaternion
+// Parameterization and Non-Dedicated GPS Receivers", John B. Schleppe 1996
+void Ekf::propagate_covariances_from_quat_to_euler(matrix::SquareMatrix<float, 3> &euler_cov)
+{
+	// Jacobian matrix (3x4) containing the partial derivatives of the
+	// Euler angle equations with respect to the quaternions
+	matrix::Matrix<float, 3, 4> G;
+
+	float q1 = _output_new.quat_nominal(0);
+	float q2 = _output_new.quat_nominal(1);
+	float q3 = _output_new.quat_nominal(2);
+	float q4 = _output_new.quat_nominal(3);
+
+	G(0,0) = -(q3+q2) / ((q3+q2)*(q3+q2) + (q4+q2)*(q4+q1)) + (q3-q2)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1));
+	G(0,1) =  (q4+q1) / ((q3+q2)*(q3+q2) + (q4+q2)*(q4+q1)) - (q4-q1)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1));
+	G(0,2) =  (q4+q1) / ((q3+q2)*(q3+q2) + (q4+q2)*(q4+q1)) + (q4-q1)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1));
+	G(0,3) = -(q3+q2) / ((q3+q2)*(q3+q2) + (q4+q2)*(q4+q1)) - (q3-q2)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1));
+	G(1,0) = 2 * q4 / sqrt(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4));
+	G(1,1) = 2 * q3 / sqrt(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4));
+	G(1,2) = 2 * q2 / sqrt(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4));
+	G(1,3) = 2 * q1 / sqrt(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4));
+	G(2,0) = G(0,3);
+	G(2,1) = G(0,2);
+	G(2,2) = G(0,1);
+	G(2,3) = G(0,0);
+
+	float quat_cov[16];
+	get_quaternion_covariances(quat_cov);
+	matrix::SquareMatrix<float, 4> Cq(quat_cov);
+	euler_cov = G * Cq * G.transpose();
+}
+
+// get the full covariance matrix
 void Ekf::get_covariances(float *covariances)
+{
+	for (unsigned n = 0; n < _k_num_states; n++) {
+		for (unsigned m = 0; m < _k_num_states; m++) {
+			covariances[n * _k_num_states + m] = P[n][m];
+		}
+	}
+}
+
+// get the diagonal elements of the covariance matrix
+void Ekf::get_covariances_diagonal(float *covariances)
 {
 	for (unsigned i = 0; i < _k_num_states; i++) {
 		covariances[i] = P[i][i];
 	}
+}
+
+// get the position covariances
+void Ekf::get_position_covariances(float *covariances)
+{
+	covariances[0] = P[7][7];	// P(X,X)
+	covariances[1] = P[7][8];
+	covariances[2] = P[7][9];
+	covariances[3] = P[8][7];
+	covariances[4] = P[8][8];	// P(Y,Y)
+	covariances[5] = P[8][9];
+	covariances[6] = P[9][7];
+	covariances[7] = P[9][8];
+	covariances[8] = P[9][9];	// P(Z,Z)
+}
+
+// get the quaternion covariances
+void Ekf::get_quaternion_covariances(float *covariances)
+{
+	for (unsigned n = 0; n < 4; n++) {
+		for (unsigned m = 0; m < 4; m++) {
+			covariances[n * 4 + m] = P[n][m];
+		}
+	}
+}
+
+// get the euler angles covariances
+void Ekf::get_euler_covariances(float *covariances)
+{
+	matrix::SquareMatrix<float, 3> euler_cov = matrix::eye<float, 3>();
+	propagate_covariances_from_quat_to_euler(euler_cov);
+
+	(*covariances) = *euler_cov.data();
+}
+
+// get the pose covariances (position + orientation in euler angles)
+void Ekf::get_pose_covariances(float *covariances)
+{
+	matrix::SquareMatrix<float, 6> pose_cov = matrix::eye<float, 6>();
+
+	// get position covariances
+	float pos_cov[9];
+	get_position_covariances(pos_cov);
+	matrix::SquareMatrix<float, 3> position_cov(pos_cov);
+
+	// get orientation covariances
+	float euler_cov[9];
+	get_euler_covariances(euler_cov);
+	matrix::SquareMatrix<float, 3> orientation_cov(euler_cov);
+
+	// Fill the pose covariance matrix
+	// It's not a cross-covariance matrix but simplifies propagating th data
+	pose_cov.set(position_cov, 0, 0);
+	pose_cov.set(orientation_cov, 3, 3);
+
+	(*covariances) = *pose_cov.data();
+}
+
+// get the linear velocity covariances
+void Ekf::get_velocity_covariances(float *covariances)
+{
+	covariances[0] = P[4][4];	// P(VX,VX)
+	covariances[1] = P[4][5];
+	covariances[2] = P[4][6];
+	covariances[3] = P[5][4];
+	covariances[4] = P[5][5];	// P(VY,VY)
+	covariances[5] = P[5][6];
+	covariances[6] = P[6][4];
+	covariances[7] = P[6][5];
+	covariances[8] = P[6][6];	// P(VZ,VZ)
 }
 
 // get the position and height of the ekf origin in WGS-84 coordinates and time the origin was set
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -982,36 +982,6 @@ void Ekf::get_gyro_bias(float bias[3])
 	memcpy(bias, temp, 3 * sizeof(float));
 }
 
-// covariance propagation from quaternions to euler angles using the covariance law
-// source: "Development of a Real-Time Attitude System Using a Quaternion
-// Parameterization and Non-Dedicated GPS Receivers", John B. Schleppe 1996
-const matrix::SquareMatrix<float, 3> Ekf::propagate_covariances_from_quat_to_euler() const
-{
-	// Jacobian matrix (3x4) containing the partial derivatives of the
-	// Euler angle equations with respect to the quaternions
-	matrix::Matrix<float, 3, 4> G;
-
-	float q1 = _output_new.quat_nominal(0);
-	float q2 = _output_new.quat_nominal(1);
-	float q3 = _output_new.quat_nominal(2);
-	float q4 = _output_new.quat_nominal(3);
-
-	G(0,0) = -(q3+q2) / ((q3+q2)*(q3+q2) + (q4+q2)*(q4+q1)) + (q3-q2)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1));
-	G(0,1) =  (q4+q1) / ((q3+q2)*(q3+q2) + (q4+q2)*(q4+q1)) - (q4-q1)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1));
-	G(0,2) =  (q4+q1) / ((q3+q2)*(q3+q2) + (q4+q2)*(q4+q1)) + (q4-q1)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1));
-	G(0,3) = -(q3+q2) / ((q3+q2)*(q3+q2) + (q4+q2)*(q4+q1)) - (q3-q2)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1));
-	G(1,0) = 2 * q4 / sqrtf(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4));
-	G(1,1) = 2 * q3 / sqrtf(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4));
-	G(1,2) = 2 * q2 / sqrtf(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4));
-	G(1,3) = 2 * q1 / sqrtf(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4));
-	G(2,0) = G(0,3);
-	G(2,1) = G(0,2);
-	G(2,2) = G(0,1);
-	G(2,3) = G(0,0);
-
-	return G * quaternion_covariances() * G.transpose();
-}
-
 // get the full covariance matrix
 const matrix::SquareMatrix<float, 24> Ekf::covariances() const
 {
@@ -1055,23 +1025,6 @@ const matrix::SquareMatrix<float, 4> Ekf::quaternion_covariances() const
 			cov(n, m) = P[n][m];
 		}
 	}
-	return cov;
-}
-
-// get the euler angles covariances
-const matrix::SquareMatrix<float, 3> Ekf::euler_covariances() const
-{
-	return propagate_covariances_from_quat_to_euler();
-}
-
-// get the pose covariances (position + orientation in euler angles)
-const matrix::SquareMatrix<float, 6> Ekf::pose_covariances() const
-{
-	// Fill the pose covariance matrix
-	// It's not a cross-covariance matrix but simplifies propagating the data
-	matrix::SquareMatrix<float, 6> cov = matrix::eye<float, 6>();
-	cov.set(position_covariances(), 0, 0);
-	cov.set(euler_covariances(), 3, 3);
 	return cov;
 }
 
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -982,21 +982,6 @@ void Ekf::get_gyro_bias(float bias[3])
 	memcpy(bias, temp, 3 * sizeof(float));
 }
 
-// get the full covariance matrix
-matrix::SquareMatrix<float, 24> Ekf::covariances() const { return matrix::SquareMatrix<float, _k_num_states>(P); }
-
-// get the diagonal elements of the covariance matrix
-matrix::Vector<float, 24> Ekf::covariances_diagonal() const { return covariances().diag(); }
-
-// get the position covariances
-matrix::SquareMatrix<float, 3> Ekf::position_covariances() const { return covariances().slice<3, 3>(7, 7); }
-
-// get the orientation (quaterion) covariances
-matrix::SquareMatrix<float, 4> Ekf::orientation_covariances() const { return covariances().slice<4, 4>(0, 0); }
-
-// get the linear velocity covariances
-matrix::SquareMatrix<float, 3> Ekf::velocity_covariances() const { return covariances().slice<3, 3>(4, 4); }
-
 // get the position and height of the ekf origin in WGS-84 coordinates and time the origin was set
 // return true if the origin is valid
 bool Ekf::get_ekf_origin(uint64_t *origin_time, map_projection_reference_s *origin_pos, float *origin_alt)
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -983,62 +983,19 @@ void Ekf::get_gyro_bias(float bias[3])
 }
 
 // get the full covariance matrix
-matrix::SquareMatrix<float, 24> Ekf::covariances() const
-{
-	matrix::SquareMatrix<float, 24> cov;
-	for (unsigned n = 0; n < _k_num_states; n++) {
-		for (unsigned m = 0; m < _k_num_states; m++) {
-			cov(n, m) = P[n][m];
-		}
-	}
-	return cov;
-}
+matrix::SquareMatrix<float, 24> Ekf::covariances() const { return matrix::SquareMatrix<float, _k_num_states>(P); }
 
 // get the diagonal elements of the covariance matrix
-matrix::Vector<float, 24> Ekf::covariances_diagonal() const
-{
-	matrix::Vector<float, 24> diag;
-	for (unsigned i = 0; i < _k_num_states; i++) {
-		diag(i) = P[i][i];
-	}
-	return diag;
-}
+matrix::Vector<float, 24> Ekf::covariances_diagonal() const { return covariances().diag(); }
 
 // get the position covariances
-matrix::SquareMatrix<float, 3> Ekf::position_covariances() const
-{
-	matrix::SquareMatrix<float, 3> cov;
-	for (unsigned n = 0; n < 3; n++) {
-		for (unsigned m = 0; m < 3; m++) {
-			cov(n, m) = P[n + 7][m + 7];
-		}
-	}
-	return cov;
-}
+matrix::SquareMatrix<float, 3> Ekf::position_covariances() const { return covariances().slice<3, 3>(7, 7); }
 
 // get the orientation (quaterion) covariances
-matrix::SquareMatrix<float, 4> Ekf::orientation_covariances() const
-{
-	matrix::SquareMatrix<float, 4> cov;
-	for (unsigned n = 0; n < 4; n++) {
-		for (unsigned m = 0; m < 4; m++) {
-			cov(n, m) = P[n][m];
-		}
-	}
-	return cov;
-}
+matrix::SquareMatrix<float, 4> Ekf::orientation_covariances() const { return covariances().slice<4, 4>(0, 0); }
 
 // get the linear velocity covariances
-matrix::SquareMatrix<float, 3> Ekf::velocity_covariances() const
-{
-	matrix::SquareMatrix<float, 3> cov;
-	for (unsigned n = 0; n < 3; n++) {
-		for (unsigned m = 0; m < 3; m++) {
-			cov(n, m) = P[n + 4][m + 4];
-		}
-	}
-	return cov;
-}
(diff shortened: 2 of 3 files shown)


EKF/ekf_helper.cpp
```diff
@@ -964,14 +964,127 @@ void Ekf::get_gyro_bias(float bias[3])
 	memcpy(bias, temp, 3 * sizeof(float));
 }
 
-// get the diagonal elements of the covariance matrix
+// covariance propagation from quaternions to euler angles using the covariance law
+// source: "Development of a Real-Time Attitude System Using a Quaternion
+// Parameterization and Non-Dedicated GPS Receivers", John B. Schleppe 1996
+void Ekf::propagate_covariances_from_quat_to_euler(matrix::SquareMatrix<float, 3> &euler_cov)
```
dagar: You can return this by value.

TSC21: Yeah I wanted to keep the consistency here but since this is only consumed internally, I guess we can return by value.

dagar: Yes, but you can still have all the new methods here returning values and marked const.

TSC21:
> How have you verified the Euler covariance output?

At this moment I am just relying on the math, as it's not an obvious verification. Though, it would make sense to add a unittest for it.

EKF/ekf_helper.cpp
TSC21: @dagar done!

EKF/ekf_helper.cpp
```diff
@@ -964,12 +964,109 @@ void Ekf::get_gyro_bias(float bias[3])
...
+	G(2,0) = G(0,3);
+	G(2,1) = G(0,2);
+	G(2,2) = G(0,1);
+	G(2,3) = G(0,0);
+
+	return G * quaternion_covariances() * G.transpose();
+}
+
+// get the full covariance matrix
+const matrix::SquareMatrix<float, 24> Ekf::covariances() const
```
dagar: Did you try using `_k_num_states` instead of 24 here?

TSC21: Yes and got a compile error stating it is undefined.

dagar:
@priseborough could you comment briefly on the intended ecl/EKF architecture? Is `EstimatorInterface` supposed to be the only interface accessible to the frontend (eg ekf2_main). I ask because there's a small additional cost to all of these virtual calls that might not need to exist at all.

TSC21:
@dagar @priseborough is there anything else I should address so we can bring this in?

LorenzMeier:
@priseborough Quick feedback would be appreciated. Thanks!

EKF/ekf_helper.cpp
```diff
@@ -964,12 +964,109 @@ void Ekf::get_gyro_bias(float bias[3])
 	memcpy(bias, temp, 3 * sizeof(float));
 }
 
+// covariance propagation from quaternions to euler angles using the covariance law
+// source: "Development of a Real-Time Attitude System Using a Quaternion
+// Parameterization and Non-Dedicated GPS Receivers", John B. Schleppe 1996
+const matrix::SquareMatrix<float, 3> Ekf::propagate_covariances_from_quat_to_euler() const
```
CarlOlsson: what is the order of rotation used here to get the Euler angles?

EKF/ekf_helper.cpp
priseborough: A 321 Tait-Bryan rotation sequence will be badly conditioned around +-90 degrees pitch. The EKF is designed to work for arbitrary rotations, so where we use Euler angles inside the EKF, we also provide an alternative 312 sequence to use when it is better conditioned.

Also I haven't gone through the maths but there is no explicit protection against negative sqrt and /0 floating point exceptions. When I tried using these equations in a Matlab script (attached), the G matrix blew up when RPY was set to zero (quaternion 1,0,0,0)

[test of tait bryan jacobian .m.txt](https://github.com/PX4/ecl/files/2732257/test.of.tait.bryan.jacobian.m.txt)

TSC21: > Also I haven't gone through the maths but there is no explicit protection against negative sqrt and /0 floating point exceptions.

@priseborough how do you suggest to add this protection?

> When I tried using these equations in a Matlab script (attached), the G matrix blew up when RPY was set to zero (quaternion 1,0,0,0)

I will go through the math again to check all's good and there's no typo. If it matches the source, then something else is blowing up the G matrix.

TSC21:
@priseborough @CarlOlsson I actually found a typo on the G matrix filling. Sorry for it. Can you please retry on your side and see if it works as it should?

CarlOlsson:
I tested it and it looks better in the sense that one angle uncertainty increase much more than the others
image

However, it is still the middle diagonal element in the Euler angles covariance matrix

TSC21:
> However, it is still the middle diagonal element in the Euler angles covariance matrix

@CarlOlsson what do you mean with this?

CarlOlsson:
Yes, I plot the diagonal elements of the Euler angles covariance matrix C.
Where I say C(0,0) is roll, C(1,1) is pitch and C(2,2) is yaw.
When the heading is reset the yaw variance should increase, not the pitch. However, it is likely that I make a mistake somewhere, don't know where yet

TSC21:
> The EKF is designed to work for arbitrary rotations, so where we use Euler angles inside the EKF, we also provide an alternative 312 sequence to use when it is better conditioned.

@CarlOlsson it may be given to this. And, in that case, we need a way of identifying when that happens (if there's a way at all). @priseborough can you give your input here?

TSC21:
> The sitl test above is from a quad so should not be related

Have you tried changing the quaternion order in https://github.com/PX4/ecl/pull/543/files#diff-6ff089a846fa4bd452bf076e28a0bef7R976 to https://github.com/PX4/ecl/pull/543/files#diff-6ff089a846fa4bd452bf076e28a0bef7R979?

CarlOlsson:
Sorry I don't understand what you mean, changing
```
float q1 = _output_new.quat_nominal(0);
float q2 = _output_new.quat_nominal(1);
float q3 = _output_new.quat_nominal(2);
float q4 = _output_new.quat_nominal(3);
```
to what?

TSC21:
Maybe,

```c++
float q1 = _output_new.quat_nominal(1);
float q2 = _output_new.quat_nominal(2);
float q3 = _output_new.quat_nominal(3);
float q4 = _output_new.quat_nominal(0);
```
I am not sure that this point.

CarlOlsson:
This gives a much smaller increase in "pitch" variance
image

EKF/ekf_helper.cpp
TSC21: @priseborough I have put the source on the comment. Maybe you can take a look at it and assert yourself about the source. I cannot tell myself that the source is correct, it was just the only source I have found that gave me a solution for this aspect and which seems trustful.

TSC21: Also, architecturaly speeking, we do need to propagate the covariances for euler angles on the odometry pipeline, since the message definitions and conventions we currently use on PX4 (and ROS) for odometry oblige to so. We may question about the source but we do need a solution for this in the end.

TSC21:
@CarlOlsson let's just be sure about the quaternion convention and then be sure that we handle also both Tait-Bryan conventions so we know we are at leas feeding the correct information to the matrix.

EKF/ekf_helper.cpp
TSC21: So you don't have to look for it: www.ucalgary.ca/engo_webdocs/GL/96.20096.JSchleppe.pdf

CarlOlsson:
I verified the Jacobian by calculating it using sympy based on the Euler angle definitions in the paper:
```
yaw = atan2(q3+q2,q4+q1) + atan2(q3-q2,q4-q1) # psi
pitch = asin(2*(q2*q3+q1*q4)) # theta
roll = atan2(q3+q2,q4+q1) - atan2(q3-q2,q4-q1) # phi
```

And it gave the same result
```
yaw_q1 = (-q2 - q3)/((q1 + q4)**2 + (q2 + q3)**2) - (q2 - q3)/((-q1 + q4)**2 + (-q2 + q3)**2)
yaw_q2 = -(-q1 + q4)/((-q1 + q4)**2 + (-q2 + q3)**2) + (q1 + q4)/((q1 + q4)**2 + (q2 + q3)**2)
yaw_q3 = (-q1 + q4)/((-q1 + q4)**2 + (-q2 + q3)**2) + (q1 + q4)/((q1 + q4)**2 + (q2 + q3)**2)
yaw_q4 = (-q2 - q3)/((q1 + q4)**2 + (q2 + q3)**2) + (q2 - q3)/((-q1 + q4)**2 + (-q2 + q3)**2)
pitch_q1 = 2*q4/sqrt(-(2*q1*q4 + 2*q2*q3)**2 + 1)
pitch_q2 = 2*q3/sqrt(-(2*q1*q4 + 2*q2*q3)**2 + 1)
pitch_q3 = 2*q2/sqrt(-(2*q1*q4 + 2*q2*q3)**2 + 1)
pitch_q4 = 2*q1/sqrt(-(2*q1*q4 + 2*q2*q3)**2 + 1)
roll_q1 = (-q2 - q3)/((q1 + q4)**2 + (q2 + q3)**2) + (q2 - q3)/((-q1 + q4)**2 + (-q2 + q3)**2)
roll_q2 = (-q1 + q4)/((-q1 + q4)**2 + (-q2 + q3)**2) + (q1 + q4)/((q1 + q4)**2 + (q2 + q3)**2)
roll_q3 = -(-q1 + q4)/((-q1 + q4)**2 + (-q2 + q3)**2) + (q1 + q4)/((q1 + q4)**2 + (q2 + q3)**2)
roll_q4 = (-q2 - q3)/((q1 + q4)**2 + (q2 + q3)**2) - (q2 - q3)/((-q1 + q4)**2 + (-q2 + q3)**2)
```

However, if I instead use the definition of the Euler angles on page 39 in [this paper](http://www.sedris.org/wg8home/Documents/WG80485.pdf) I get the following result:

```
yaw = atan2(q2*q3+q1*q4,0.5-(q3**2+q4**2)) # psi
pitch = asin(-2*(q2*q4-q1*q3)) # theta
roll = atan2(q3*q4+q1*q2,0.5-(q2**2+q3**2)) # phi
yaw_q1 = q4*(-q3**2 - q4**2 + 0.5)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2)
yaw_q2 = q3*(-q3**2 - q4**2 + 0.5)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2)
yaw_q3 = q2*(-q3**2 - q4**2 + 0.5)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2) - 2*q3*(-q1*q4 - q2*q3)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2)
yaw_q4 = q1*(-q3**2 - q4**2 + 0.5)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2) - 2*q4*(-q1*q4 - q2*q3)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2)
pitch_q1 = 2*q3/sqrt(-(2*q1*q3 - 2*q2*q4)**2 + 1)
pitch_q2 = -2*q4/sqrt(-(2*q1*q3 - 2*q2*q4)**2 + 1)
pitch_q3 = 2*q1/sqrt(-(2*q1*q3 - 2*q2*q4)**2 + 1)
pitch_q4 = -2*q2/sqrt(-(2*q1*q3 - 2*q2*q4)**2 + 1)
roll_q1 = q2*(-q2**2 - q3**2 + 0.5)/((q1*q2 + q3*q4)**2 + (-q2**2 - q3**2 + 0.5)**2)
roll_q2 = q1*(-q2**2 - q3**2 + 0.5)/((q1*q2 + q3*q4)**2 + (-q2**2 - q3**2 + 0.5)**2) - 2*q2*(-q1*q2 - q3*q4)/((q1*q2 + q3*q4)**2 + (-q2**2 - q3**2 + 0.5)**2)
roll_q3 = -2*q3*(-q1*q2 - q3*q4)/((q1*q2 + q3*q4)**2 + (-q2**2 - q3**2 + 0.5)**2) + q4*(-q2**2 - q3**2 + 0.5)/((q1*q2 + q3*q4)**2 + (-q2**2 - q3**2 + 0.5)**2)
roll_q4 = q3*(-q2**2 - q3**2 + 0.5)/((q1*q2 + q3*q4)**2 + (-q2**2 - q3**2 + 0.5)**2)
```

If I then use this to calculate the Euler angle uncertainties from the same logfile as [here](https://github.com/PX4/ecl/issues#issuecomment-452227301) I get this
image

And this seems correct! The Jacobian is
image

So the first element should be yaw. I guess the difference comes from the order of rotations in the definition of the Euler angles

TSC21:
Great findings! So you suggest we change the jacobian or that we adapt the input according the the euler convention?

CarlOlsson:
I just did another test and something is still not right.
Here is the Eular angles variances from a square mission, in every corner the quad was loitering for 30 seconds. During this time I increased and decreased the heading noise
image

As can be seen which element that experience an increase in uncertainty is depending on the heading

jkflying:
@TSC21 I'd suggest making an alternative implementation (in a separate file/project) using Eigen parameterized with their AutoDiff types, so that you can test for what the rotation order is they are using. The Eigen code would look something like this:

```
double x,y,z,w; // inputs

using namespace Eigen;
using AD = AutoDiffScalar<Vector4d>;
using QuaterionAD = Quaterion<AD>;
AD xd(x, Vector4d(1,0,0,0));
AD yd(y, Vector4d(0,1,0,0));
AD zd(z, Vector4d(0,0,1,0));
AD wd(w, Vector4d(0,0,0,1));

QuaternionAD qad(xd, yd, zd, wd);

Matrix<AD, 3, 1> values = qad.toRotationMatrix().eulerAngles(2,1,0);

Matrix<double, 3, 4> G;
for (int i = 0; i < 3; i++)
    G.row(i) = values(i).derivatives;
(apologies for any compiler errors, I'm typing this straight in the comment box)
```
This would then let you compare your G matrix with different input values and different euler angle ordering.


Regarding the protection against /0 and negative sqrt that @priseborough mentioned, this would be something along the lines of factoring out those denominators `d1 = ((q3+q2)*(q3+q2) + (q4+q1)*(q4+q1))` , `d2 = (q3-q2)*(q3-q2) + (q4-q1)*(q4-q1)` and checking `abs(d1) > 1e-5f && abs(d2) > 1e-5f`. For the sqrt checking that `1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4) > 0.f`. This is to prevent NaN propagation or, platform dependent, crashes.
However, it probably makes sense to figure out the euler angle ordering first.

TSC21:
@CarlOlsson the source outputs a covariance matrix with the diagonal being the variances of Yaw, Pitch, Roll, by that order, which means that the input should match the same convention. So we need to make a check for the input quaternion, check the convention, and apply the correct convention when possible.

CarlOlsson:
@TSC21 As discussed yesterday, below is the python code for the tests I did above:

**Calculating the Jacobian**
```
from sympy import *

q1, q2, q3, q4 = symbols('q1 q2 q3 q4')

# From https://www.ucalgary.ca/engo_webdocs/GL/96.20096.JSchleppe.pdf
# yaw = atan2(q3+q2,q4+q1) + atan2(q3-q2,q4-q1) # psi
# pitch = asin(2*(q2*q3+q1*q4)) # theta
# roll = atan2(q3+q2,q4+q1) - atan2(q3-q2,q4-q1) # phi

# From https://math.stackexchange.com/questions/687964/getting-euler-tait-bryan-angles-from-quaternion-representation
yaw = atan2(q2*q3+q1*q4,0.5-(q3**2+q4**2)) # psi
pitch = asin(-2*(q2*q4-q1*q3)) # theta
roll = atan2(q3*q4+q1*q2,0.5-(q2**2+q3**2)) # phi

yaw_q1 = diff(yaw, q1)
yaw_q2 = diff(yaw, q2)
yaw_q3 = diff(yaw, q3)
yaw_q4 = diff(yaw, q4)

pitch_q1 = diff(pitch, q1)
pitch_q2 = diff(pitch, q2)
pitch_q3 = diff(pitch, q3)
pitch_q4 = diff(pitch, q4)

roll_q1 = diff(roll, q1)
roll_q2 = diff(roll, q2)
roll_q3 = diff(roll, q3)
roll_q4 = diff(roll, q4)

print('yaw_q1 = ' + str(yaw_q1))
print('yaw_q2 = ' + str(yaw_q2))
print('yaw_q3 = ' + str(yaw_q3))
print('yaw_q4 = ' + str(yaw_q4))

print('pitch_q1 = ' + str(pitch_q1))
print('pitch_q2 = ' + str(pitch_q2))
print('pitch_q3 = ' + str(pitch_q3))
print('pitch_q4 = ' + str(pitch_q4))

print('roll_q1 = ' + str(roll_q1))
print('roll_q2 = ' + str(roll_q2))
print('roll_q3 = ' + str(roll_q3))
print('roll_q4 = ' + str(roll_q4))
```

**Converting the quaternion covariance matrix to an Euler angle covariance matrix**
```
from pyulog import *
import matplotlib.pyplot as plt
import numpy as np
import pandas as pd

# Convert a pyulog.core.ULog object to a dictionary of dataframes
def ulog_to_df(logfile_str):
    ulog_dict = {}
    for elem in sorted(ULog(logfile_str).data_list, key=lambda d: d.name + str(d.multi_id)):
        topic_name = elem.name + "_" + str(elem.multi_id)
        column_names = set(elem.data.keys())
        df = pd.DataFrame(index=elem.data['timestamp'] / 1e6)
        for name in column_names - {'timestamp'}:
            df[name] = elem.data[name]

        ulog_dict[topic_name] = df
    return ulog_dict


df_dict = ulog_to_df('path_to_logfile_as_string')

euler_cov = []
yaw = []
pitch = []
roll = []
for index, row in df_dict['estimator_status_0'].iterrows():
    q1 = row['states[0]']
    q2 = row['states[1]']
    q3 = row['states[2]']
    q4 = row['states[3]']
    
    # From https://www.ucalgary.ca/engo_webdocs/GL/96.20096.JSchleppe.pdf
    # yaw.append(np.arctan2(q3+q2,q4+q1) + np.arctan2(q3-q2,q4-q1))
    # pitch.append(np.arcsin(2*(q2*q3+q1*q4)))
    # roll.append(np.arctan2(q3+q2,q4+q1) - np.arctan2(q3-q2,q4-q1))

    # From https://math.stackexchange.com/questions/687964/getting-euler-tait-bryan-angles-from-quaternion-representation
    yaw.append(np.arctan2(q2*q3+q1*q4,0.5-(q3**2+q4**2)))
    pitch.append(np.arcsin(-2*(q2*q4-q1*q3)))
    roll.append(np.arctan2(q3*q4+q1*q2,0.5-(q2**2+q3**2)))

    C = np.zeros((4,4))
    C[0][0] = row['quat_covariances[0]']
    C[0][1] = row['quat_covariances[1]']
    C[0][2] = row['quat_covariances[2]']
    C[0][3] = row['quat_covariances[3]']

    C[1][0] = row['quat_covariances[4]']
    C[1][1] = row['quat_covariances[5]']
    C[1][2] = row['quat_covariances[6]']
    C[1][3] = row['quat_covariances[7]']

    C[2][0] = row['quat_covariances[8]']
    C[2][1] = row['quat_covariances[9]']
    C[2][2] = row['quat_covariances[10]']
    C[2][3] = row['quat_covariances[11]']

    C[3][0] = row['quat_covariances[12]']
    C[3][1] = row['quat_covariances[13]']
    C[3][2] = row['quat_covariances[14]']
    C[3][3] = row['quat_covariances[15]']
    
    G = np.zeros((3,4))
    # From https://www.ucalgary.ca/engo_webdocs/GL/96.20096.JSchleppe.pdf
    if False:
        G[0][0] = -(q3+q2) / ((q3+q2)*(q3+q2) + (q4+q1)*(q4+q1)) + (q3-q2)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1))
        G[0][1] =  (q4+q1) / ((q3+q2)*(q3+q2) + (q4+q1)*(q4+q1)) - (q4-q1)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1))
        G[0][2] =  (q4+q1) / ((q3+q2)*(q3+q2) + (q4+q1)*(q4+q1)) + (q4-q1)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1))
        G[0][3] = -(q3+q2) / ((q3+q2)*(q3+q2) + (q4+q1)*(q4+q1)) - (q3-q2)/((q3-q2)*(q3-q2) + (q4-q1)*(q4-q1))
        G[1][0] = 2 * q4 / np.sqrt(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4))
        G[1][1] = 2 * q3 / np.sqrt(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4))
        G[1][2] = 2 * q2 / np.sqrt(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4))
        G[1][3] = 2 * q1 / np.sqrt(1 - 4 * (q2*q3 + q1*q4)*(q2*q3 + q1*q4))
        G[2][0] = G[0][3]
        G[2][1] = G[0][2]
        G[2][2] = G[0][1]
        G[2][3] = G[0][0]
    
    # From https://math.stackexchange.com/questions/687964/getting-euler-tait-bryan-angles-from-quaternion-representation
    else:
        G[0][0] = q4*(-q3**2 - q4**2 + 0.5)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2)
        G[0][1] = q3*(-q3**2 - q4**2 + 0.5)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2)
        G[0][2] = q2*(-q3**2 - q4**2 + 0.5)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2) - 2*q3*(-q1*q4 - q2*q3)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2)
        G[0][3] = q1*(-q3**2 - q4**2 + 0.5)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2) - 2*q4*(-q1*q4 - q2*q3)/((q1*q4 + q2*q3)**2 + (-q3**2 - q4**2 + 0.5)**2)
        G[1][0] = 2*q3/np.sqrt(-(2*q1*q3 - 2*q2*q4)**2 + 1)
        G[1][1] = -2*q4/np.sqrt(-(2*q1*q3 - 2*q2*q4)**2 + 1)
        G[1][2] = 2*q1/np.sqrt(-(2*q1*q3 - 2*q2*q4)**2 + 1)
        G[1][3] = -2*q2/np.sqrt(-(2*q1*q3 - 2*q2*q4)**2 + 1)
        G[2][0] = G[0][3]
        G[2][1] = G[0][2]
        G[2][2] = G[0][1]
        G[2][3] = G[0][0]
        
    euler_cov.append(G @ C @ G.transpose())
```

**Plot the result**
```
E0 = []
E1 = []
E2 = []
for elem in euler_cov:
    E0.append(elem[0][0])
    E1.append(elem[1][1])
    E2.append(elem[2][2])

plt.figure(0)
plt.plot(E0)
plt.plot(E1)
plt.plot(E2)
plt.grid(True)
plt.legend(['diag elem 0','diag elem 1','diag elem 2'])

plt.figure(1)
plt.plot(roll)
plt.plot(pitch)
plt.plot(yaw)
plt.grid(True)
plt.legend(['roll', 'pitch', 'yaw'])
```

This code of course assumes that the quaternion covariance matrix is logged in estimator status as quat_covariances, such as in [this](https://logs.px4.io/plot_app?log=decfa528-c32b-4049-9d47-6205a514ce3e) logfile. If needed I can also link that branch

priseborough:
> Maybe,
> 
> ```c++
> float q1 = _output_new.quat_nominal(1);
> float q2 = _output_new.quat_nominal(2);
> float q3 = _output_new.quat_nominal(3);
> float q4 = _output_new.quat_nominal(0);
> ```
> 
> I am not sure at this point.

That was what I meant

TSC21:
> > Maybe,
> > 
> > ```c++
> > float q1 = _output_new.quat_nominal(1);
> > float q2 = _output_new.quat_nominal(2);
> > float q3 = _output_new.quat_nominal(3);
> > float q4 = _output_new.quat_nominal(0);
> > ```
> > 
> > I am not sure at this point.
> >
> That was what I meant

So that's the definite order on this case? What if I one wants the 123 order? The idea is to actually get the covariances in the Roll, Pitch, Yaw order.

TSC21:
The resulting Jacobian for Euler angles in the XYZ notation is:
```python
roll = atan2(2*(q1*q2+q2*q3),1-2*(q2**2+q3**2)) #phi
pitch = asin(2*(q1*q3+q4*q2)) #theta
yaw = atan2(2*(q1*q4+q2*q3),1-2*(q3**2+q4**2)) #psi

yaw_q1 = 2*q4*(-2*q3**2 - 2*q4**2 + 1)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2)
yaw_q2 = 2*q3*(-2*q3**2 - 2*q4**2 + 1)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2)
yaw_q3 = 2*q2*(-2*q3**2 - 2*q4**2 + 1)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2) - 4*q3*(-2*q1*q4 - 2*q2*q3)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2)
yaw_q4 = 2*q1*(-2*q3**2 - 2*q4**2 + 1)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2) - 4*q4*(-2*q1*q4 - 2*q2*q3)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2)
pitch_q1 = 2*q3/sqrt(-(2*q1*q3 + 2*q2*q4)**2 + 1)
pitch_q2 = 2*q4/sqrt(-(2*q1*q3 + 2*q2*q4)**2 + 1)
pitch_q3 = 2*q1/sqrt(-(2*q1*q3 + 2*q2*q4)**2 + 1)
pitch_q4 = 2*q2/sqrt(-(2*q1*q3 + 2*q2*q4)**2 + 1)
roll_q1 = 2*q2*(-2*q2**2 - 2*q3**2 + 1)/((2*q1*q2 + 2*q2*q3)**2 + (-2*q2**2 - 2*q3**2 + 1)**2)
roll_q2 = -4*q2*(-2*q1*q2 - 2*q2*q3)/((2*q1*q2 + 2*q2*q3)**2 + (-2*q2**2 - 2*q3**2 + 1)**2) + (2*q1 + 2*q3)*(-2*q2**2 - 2*q3**2 + 1)/((2*q1*q2 + 2*q2*q3)**2 + (-2*q2**2 - 2*q3**2 + 1)**2)
roll_q3 = 2*q2*(-2*q2**2 - 2*q3**2 + 1)/((2*q1*q2 + 2*q2*q3)**2 + (-2*q2**2 - 2*q3**2 + 1)**2) - 4*q3*(-2*q1*q2 - 2*q2*q3)/((2*q1*q2 + 2*q2*q3)**2 + (-2*q2**2 - 2*q3**2 + 1)**2)
roll_q4 = 0
```
@CarlOlsson

TSC21:
@CarlOlsson here's the resulting plot for the log you shared:
variances_plot
I think now this matches with the result you expect right?
Here's the G matrix:
```python
G[0][0] = 2*q2*(-2*q2**2 - 2*q3**2 + 1)/((2*q1*q2 + 2*q2*q4)**2 + (-2*q2**2 - 2*q3**2 + 1)**2)
        G[0][1] = -4*q2*(-2*q1*q2 - 2*q2*q4)/((2*q1*q2 + 2*q2*q4)**2 + (-2*q2**2 - 2*q3**2 + 1)**2) + (2*q1 + 2*q4)*(-2*q2**2 - 2*q3**2 + 1)/((2*q1*q2 + 2*q2*q4)**2 + (-2*q2**2 - 2*q3**2 + 1)**2)
        G[0][2] = -4*q3*(-2*q1*q2 - 2*q2*q4)/((2*q1*q2 + 2*q2*q4)**2 + (-2*q2**2 - 2*q3**2 + 1)**2)
        G[0][3] = 2*q2*(-2*q2**2 - 2*q3**2 + 1)/((2*q1*q2 + 2*q2*q4)**2 + (-2*q2**2 - 2*q3**2 + 1)**2)
        G[1][0] = 2*q3/np.sqrt(-(2*q1*q3 + 2*q2*q4)**2 + 1)
        G[1][1] = 2*q4/np.sqrt(-(2*q1*q3 + 2*q2*q4)**2 + 1)
        G[1][2] = 2*q1/np.sqrt(-(2*q1*q3 + 2*q2*q4)**2 + 1)
        G[1][3] = 2*q2/np.sqrt(-(2*q1*q3 + 2*q2*q4)**2 + 1)
        G[2][0] = 2*q4*(-2*q3**2 - 2*q4**2 + 1)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2)
        G[2][1] = 2*q3*(-2*q3**2 - 2*q4**2 + 1)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2)
        G[2][2] = 2*q2*(-2*q3**2 - 2*q4**2 + 1)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2) - 4*q3*(-2*q1*q4 - 2*q2*q3)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2)
        G[2][3] = 2*q1*(-2*q3**2 - 2*q4**2 + 1)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2) - 4*q4*(-2*q1*q4 - 2*q2*q3)/((2*q1*q4 + 2*q2*q3)**2 + (-2*q3**2 - 2*q4**2 + 1)**2)
```

TSC21:
Also, using,
```python
yaw.append(np.arctan2(q2*q3+q1*q4,0.5-(q3**2+q4**2)))
pitch.append(np.arcsin(-2*(q2*q4-q1*q3)))
roll.append(np.arctan2(q3*q4+q1*q2,0.5-(q2**2+q3**2)))
```
or
```python
roll.append(np.arctan2(2*(q1*q2+q2*q4),1-2*(q2**2+q3**2)))
pitch.append(np.arcsin(-2*(q2*q4-q1*q3)))
yaw.append(np.arctan2(2*(q1*q4+q2*q3),1-2*(q3**2+q4**2)))
```
Gives exactly the same result as above, with the same G matrix.

CarlOlsson:
Yes awesome!
For both logfiles this looks exactly how we would expect the yaw variance to increase.
The problem seems to have been that in the [first paper](https://www.ucalgary.ca/engo_webdocs/GL/96.20096.JSchleppe.pdf) (page 12) that was used q4 was the element of the quaternion vector defining the amount of rotation, and in the last equations used (from [here](https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/Conversion_between_quaternions_and_Euler_angles.html#cite_note-nasa-rotation-1)) the first element of the quaternion vector is the rotation which corresponds to the convention used in the EKF.

TSC21:
>Yes awesome!
>For both logfiles this looks exactly how we would expect the yaw variance to increase.
>The problem seems to have been that in the [first paper](https://www.ucalgary.ca/engo_webdocs/GL/96.20096.JSchleppe.pdf) (page 12) that was used q4 was the element of the quaternion vector defining the amount of rotation, and in the last equations used (from [here](https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/Conversion_between_quaternions_and_Euler_angles.html#cite_note-nasa-rotation-1)) the first element of the quaternion vector is the rotation which corresponds to the convention used in the EKF.

Yeah that basically matches what I have here: https://github.com/PX4/ecl/pull/543#issuecomment-452099335.

I will fix the G matrix then and add protections against divisions by zero and and sqrt of negative. My question is: @priseborough @jkflying if a division by zero or a sqrt of negative is detected, what should be done? We just don't send the covariances?

jkflying:
@TSC21 The normal solution is to change to a different Euler angle convention when approaching gimbal lock, or applying a rotation to take it into a safe area and transforming back one the problem is solved. The other solution of course is to stick with quaterions, is there a reason you need the Euler angles specifically and can't work with the quaterion values?

TSC21:
> @TSC21 The normal solution is to change to a different Euler angle convention when approaching gimbal lock, or applying a rotation to take it into a safe area and transforming back one the problem is solved.

So you mean that when approaching a gimbal lock on the `_output_new` quaternion values, we should change the quaterion convention to 312 for example? But that would result on a different orientation covariance, and the message spec expects a 3x3 XYZ euler angle covariance.

> The other solution of course is to stick with quaterions, is there a reason you need the Euler angles specifically and can't work with the quaterion values?

The odometry message spec, both in PX4 and on ROS, expects the orientation covariance to be in the XYZ euler convention. Not in quaternions. The current pipeline aims to respect [ROS REP103](http://www.ros.org/reps/rep-0103.html) and [ROS REP147](http://www.ros.org/reps/rep-0147.html).

TSC21:
I guess that if the spec expects a rotation order of XYZ, if we reach a gimbal lock, we probably don't want to propagate the covariances, right?

TSC21:
@CarlOlsson I have been thinking about the rotation convention and I think that what we have here is enough. What you need to be worried is the way that the quaternion is initialised, which in this case is using the XYZW order. From there, what you actually have to worry about is the output: if you want a covariance matrix with the Roll, Pitch, Yaw rotation order, then you need to compute the Jacobian for that same order. Since the convention that you want is defined, then you just need to use one Jacobian. You can though provide options for the rotation orders, depending on the convention you want, by passing the rotation order as an argument to the function and apply the correct Jacobian according to it.
For now, and considering that this function it's only useful to fill the covariance matrices on the `vehicle_odometry` message, which expects a RPY rotation order, I think the current method is enough.

@priseborough, when you are able, I would like your quick feedback here so we can consider this good and merge it. Thanks!

CarlOlsson:
I don't fully understand your comment above

>  What you need to be worried is the way that the quaternion is initialised, which in this case is using the XYZW order.

Yes the EKF is using the Hamilton convention. There was some confusion regarding this above but now as you found that problem it should be sorted out

> Since the convention that you want is defined, then you just need to use one Jacobian.

Do you mean you as in me? In that case I don't want any convention, I personally don't plan to use this code in the near future :)

> For now, and considering that this function it's only useful to fill the covariance matrices on the vehicle_odometry message, which expects a RPY rotation order, I think the current method is enough.

The problem is that ECL is a library used in more projects than standard PX4. So if you add a function which calculates the Euler angle covariances it should work for any attitude and the current implementation in this PR does not work for pitch = 90 degrees. 

My suggestion would be to either
1) Only output the 4x4 quaternion covariance matrix from ECL and calculate the Euler uncertainty at the place you need it, e.g. in ekf2_main.cpp in PX4 firmware.
2) Switch Jacobian matrix depending on the attitude of the vehicle

TSC21:
> Do you mean you as in me? In that case I don't want any convention, I personally don't plan to use this code in the near future :)

I meant someone in general :)

> Only output the 4x4 quaternion covariance matrix from ECL and calculate the Euler uncertainty at the place you need it, e.g. in ekf2_main.cpp in PX4 firmware.

Probably then that's what I do. I don't want to support stuff that in the end it's not used on the protocols/pipelines. So if everyone agrees, I will push the Euler covariance computation into the PX4 side and let ECL just output the quaternion covariances (which is a method I also added in this PR).

@jkflying @dagar @priseborough thoughts?

jkflying:
Makes sense to me. The calculation of the G matrix and the G * q_cov * Gt could even be done right near the sending of the message, since it is only the protocol which is causing the limitation of the covariance needing to be in Euler angles, not quaternions. Better not to mix a specific protocol's requirements into the EKF.

TSC21:
We can do that but the message sending (uORB) part is still on the ekf2_main side.

priseborough:
I agree. Lets keep the ecl attitude representation as a quaternion and move any translation to euler angles into the PX4/Firmware. The logical place to output the topic would be the ekf2_main module, but added in such a way that it is after publication of any states used by control loops so we are not affecting control latency with.

TSC21:
@priseborough @jkflying brought the covariance propagation to EKF2 in PX4 now: https://github.com/PX4/Firmware/pull/11092/commits/9294f3a983d7022b135f8b938847f0a85e95e438. Can you do a last check and approve it if all ok so we can bring it in?

EKF/ekf_helper.cpp
```diff
@@ -972,12 +972,62 @@ void Ekf::get_gyro_bias(float bias[3])
...
+	for (unsigned n = 0; n < 3; n++) {
+		for (unsigned m = 0; m < 3; m++) {
+			cov(n, m) = P[n + 7][m + 7];
+		}
+	}
+	return cov;
+}
+
+// get the quaternion covariances
+const matrix::SquareMatrix<float, 4> Ekf::quaternion_covariances() const
```
jkflying: I'd rather call this `orientation_covariances` to be consistent with the other naming.

TSC21: alright. done in 9961b59

TSC21:
@dagar anything else you want to add here? if not, can you merge it?

TSC21:
@dagar how much work is left before we can bring this in? #554 and #556 are already solved.

EKF/estimator_interface.h
```diff
@@ -104,7 +104,15 @@ class EstimatorInterface
 
 	virtual void get_true_airspeed(float *tas) = 0;
 
-	virtual void get_covariances(float *covariances) = 0;
+	virtual const matrix::SquareMatrix<float, 24> covariances() const = 0;
```
dagar: These don't need to be part of EstimatorInterface, the ekf2 frontend instantiates an Ekf instance. 
https://github.com/PX4/Firmware/blob/master/src/modules/ekf2/ekf2_main.cpp#L288

TSC21: So I just remove them?

dagar: Yes

TSC21: Done

TSC21:
@dagar rebased and addressed your comments. Can we finally bring this in?

EKF/ekf.h
```diff
@@ -122,8 +122,20 @@ class Ekf : public EstimatorInterface
 	// get the true airspeed in m/s
 	void get_true_airspeed(float *tas);
 
+	// get the full covariance matrix
+	const matrix::SquareMatrix<float, 24> covariances() const;
```
dagar:
```suggestion
	matrix::SquareMatrix<float, 24> covariances() const { return matrix::SquareMatrix<float, 24>{P}; }
```

EKF/ekf.h
```diff
@@ -122,8 +122,20 @@ class Ekf : public EstimatorInterface
...
+
 	// get the diagonal elements of the covariance matrix
-	void get_covariances(float *covariances);
+	const matrix::Vector<float, 24> covariances_diagonal() const;
```
dagar:
```suggestion
	matrix::Vector<float, 24> covariances_diagonal() const;
```

EKF/ekf.h
```diff
@@ -122,8 +122,20 @@ class Ekf : public EstimatorInterface
...
+
+	// get the position covariances
+	const matrix::SquareMatrix<float, 3> position_covariances() const;
```
dagar:
```suggestion
	matrix::SquareMatrix<float, 3> position_covariances() const;
```

EKF/ekf.h
```diff
@@ -122,8 +122,20 @@ class Ekf : public EstimatorInterface
...
+
+	// get the orientation (quaternion) covariances
+	const matrix::SquareMatrix<float, 4> orientation_covariances() const;
```
dagar:
```suggestion
	matrix::SquareMatrix<float, 4> orientation_covariances() const;
```

EKF/ekf.h
```diff
@@ -122,8 +122,20 @@ class Ekf : public EstimatorInterface
...
+
+	// get the linear velocity covariances
+	const matrix::SquareMatrix<float, 3> velocity_covariances() const;
```
dagar:
```suggestion
	matrix::SquareMatrix<float, 3> velocity_covariances() const;
```

EKF/ekf_helper.cpp
```diff
@@ -982,12 +982,62 @@ void Ekf::get_gyro_bias(float bias[3])
 	memcpy(bias, temp, 3 * sizeof(float));
 }
 
+// get the full covariance matrix
+const matrix::SquareMatrix<float, 24> Ekf::covariances() const
```
dagar: Can be removed (single line in header).

EKF/ekf_helper.cpp
```diff
@@ -982,12 +982,62 @@ void Ekf::get_gyro_bias(float bias[3])
...
+		for (unsigned m = 0; m < _k_num_states; m++) {
+			cov(n, m) = P[n][m];
+		}
+	}
+	return cov;
+}
+
 // get the diagonal elements of the covariance matrix
-void Ekf::get_covariances(float *covariances)
+const matrix::Vector<float, 24> Ekf::covariances_diagonal() const
```
dagar:
```suggestion
matrix::Vector<float, 24> Ekf::covariances_diagonal() const
```

EKF/ekf_helper.cpp
```diff
@@ -982,12 +982,62 @@ void Ekf::get_gyro_bias(float bias[3])
...
+	matrix::Vector<float, 24> diag;
 	for (unsigned i = 0; i < _k_num_states; i++) {
-		covariances[i] = P[i][i];
+		diag(i) = P[i][i];
+	}
+	return diag;
+}
+
+// get the position covariances
+const matrix::SquareMatrix<float, 3> Ekf::position_covariances() const
```
dagar:
```suggestion
matrix::SquareMatrix<float, 3> Ekf::position_covariances() const
```

EKF/ekf_helper.cpp
```diff
@@ -982,12 +982,62 @@ void Ekf::get_gyro_bias(float bias[3])
...
+	for (unsigned n = 0; n < 3; n++) {
+		for (unsigned m = 0; m < 3; m++) {
+			cov(n, m) = P[n + 7][m + 7];
+		}
+	}
+	return cov;
+}
+
+// get the orientation (quaterion) covariances
+const matrix::SquareMatrix<float, 4> Ekf::orientation_covariances() const
```
dagar:
```suggestion
matrix::SquareMatrix<float, 4> Ekf::orientation_covariances() const
```

EKF/ekf_helper.cpp
```diff
@@ -982,12 +982,62 @@ void Ekf::get_gyro_bias(float bias[3])
...
+	for (unsigned n = 0; n < 4; n++) {
+		for (unsigned m = 0; m < 4; m++) {
+			cov(n, m) = P[n][m];
+		}
+	}
+	return cov;
+}
+
+// get the linear velocity covariances
+const matrix::SquareMatrix<float, 3> Ekf::velocity_covariances() const
```
dagar:
```suggestion
matrix::SquareMatrix<float, 3> Ekf::velocity_covariances() const
```

EKF/ekf_helper.cpp
TSC21: I was actually trying `matrix::Vector<float, 24> Ekf::covariances_diagonal() const { return matrix::Vector<float, 24>{P}; }` by your suggestion but that doesn't work.

dagar: Matrix or SquareMatrix, not Vector.

TSC21:
@dagar anything else here?

TSC21:
I was trying to make the implementation take advantage of what Matrix has to offer and tried to make all the function one-liners:
```c++
// get the full covariance matrix
matrix::SquareMatrix<float, 24> Ekf::covariances() const { return matrix::SquareMatrix<float, _k_num_states>(P); }

// get the diagonal elements of the covariance matrix
matrix::Vector<float, 24> Ekf::covariances_diagonal() const { return covariances().diag(); }

// get the position covariances
matrix::SquareMatrix<float, 3> Ekf::position_covariances() const { return covariances().slice<3, 3>(7, 7); }

// get the orientation (quaterion) covariances
matrix::SquareMatrix<float, 4> Ekf::orientation_covariances() const { return covariances().slice<4, 4>(0, 0); }

// get the linear velocity covariances
matrix::SquareMatrix<float, 3> Ekf::velocity_covariances() const { return covariances().slice<3, 3>(4, 4); }
```

Even though they build correctly, there seems to be some sort of an issue with the `slice` function, since the returned values are different from the ones currently being returned from the for loops.

@jgoppert any thoughts on this? I suppose we are always considering row-major representations on this, right?

@dagar any suggestions?

dagar:
Related: https://github.com/PX4/Matrix/pull/76

EKF/ekf_helper.cpp
```diff
@@ -982,12 +982,62 @@ void Ekf::get_gyro_bias(float bias[3])
 	memcpy(bias, temp, 3 * sizeof(float));
 }
 
+// get the full covariance matrix
+matrix::SquareMatrix<float, 24> Ekf::covariances() const
```
bkueng: Can you use `_k_num_states`, or does that not work?

TSC21: @bkueng as said in https://github.com/PX4/ecl/pull/543#pullrequestreview-188416379, it doesn't work.

TSC21:
@bkueng maybe we can try to bring https://github.com/PX4/Matrix/pull/76 first and then I will update this with one liner functions?

TSC21:
@dagar @bkueng I just converted the functions into one-liners, since we now have the PX4 Matrix `slice` method working as it should. I have tested it on https://github.com/PX4/Firmware/pull/11092. Can we finally bring this in? Thank you.

TSC21:
Thanks Daniel!

PR #544, opened 2018-12-23
EKF: Critical yaw reset bug fix

priseborough:
Due to a combination of previous changes, the reset of the yaw when climbing above 1.5m was not performed until 3-axis fusion was enabled. This could result in loss of navigation if the ground level magnetic declination error was large enough and the value of  EKF2_MAG_TYPE and the flight profile did not cause the EKF to commence 3-axis fusion.

SITL test log before fix: https://logs.px4.io/plot_app?log=c9f62df4-3898-4e61-b6e0-791d984ea964
screen shot 2018-12-24 at 10 15 03 am

SITL test log after fix: https://logs.px4.io/plot_app?log=66c9b4a1-1d67-4b66-a424-fc7e5b165075
screen shot 2018-12-24 at 10 15 25 am

These were tested using the following modification to sitl_gazebo: https://github.com/priseborough/sitl_gazebo/commit/16a86ee1352e4cbf6c1f18f9047a1bc8d0613f7c

We need a way to incorporate this type of magnetic field error into the reversion testing.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1356,10 +1356,19 @@ void Ekf::controlMagFusion()
 	// check for new magnetometer data that has fallen behind the fusion time horizon
 	// If we are using external vision data for heading then no magnetometer fusion is used
 	if (!_control_status.flags.ev_yaw && _mag_data_ready) {
+
+		// We need to reset the yaw angle after climbing away from the ground to enable
+		// recovery from ground level magnetic interference.
+		if (!_flt_mag_align_complete) {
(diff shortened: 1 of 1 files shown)


RomanBapst:
@LorenzMeier Can you please merge?

hdiethelm:
Hello,
Can you tell me if this bug is already an issue in PX4 Stable using ECL  1a110687
I don't want to update to master to avoid new problems. But it seams there is a problem with the mag sensor.
Thanks!

PR #547, opened 2019-01-02
EKF: added absolute range altimeter

tuloski:
Added support for absolute altimeter or for range finders mounted on a gimbal such that they are alway vertical, so there is no need to rotate from body to earth frame.
A parameter is used to enable this feature.

WARNING: requires also the pull request in PX4/Firmware with the appropriate modifications to EKF2 module.


tuloski:
It's a good question :) but actually I don't know if that rotation matrix (or only the coscos in R[2][2]) can be used in other scenario.

priseborough:
Please provide a test log. For a simple change like this, it could be using SITL, with the range finder modified to provide true HAGL.

priseborough:
I'm closing this as it has become stale and there has been no progress for some time. Please rebase. and open a new PR with test evidence.

PR #550, opened 2019-01-04
EKF: add flt_mag_align_complete to control_status flags

CarlOlsson:
@priseborough I added the flt_mag_align_complete to the control_status flags since it controls quite some things internally in the EKF


mhkabir:
Nitpick, but do we still need the `flt_` prefix?

CarlOlsson:
True, I can remove it if no one speaks up against it

priseborough:
If this is merged, then a corresponding PR is required upstream for this https://github.com/PX4/Firmware/blob/master/msg/estimator_status.msg#L25

PR #551, opened 2019-01-06
replace <cfloat> with <float.h>

dagar:
- <cfloat> isn't available in the NuttX c++ library

https://github.com/PX4/Firmware/issues/11152


dagar:
No difference in the generated code, this is safe to merge.

PR #559, opened 2019-01-23
EKF: add method to get the terrain variance

CarlOlsson:


EKF/terrain_estimator.cpp
```diff
@@ -217,3 +217,8 @@ void Ekf::checkRangeDataContinuity()
 		_range_data_continuous = false;
 	}
 }
+
+void Ekf::get_terrain_var(float *ret)
+{
+	memcpy(ret, &_terrain_var, sizeof(float));
```
dagar: This looks fine, and is consistent with the existing API, but how about just returning the value directly?

    float Ekf::get_terrain_var() const { return _terrain_var; }

Simple, safe, one liner directly in the header.

CarlOlsson: nice! I added it to ekf.h, was this how you meant?

PR #560, opened 2019-01-23
EKF: scale wind process noise with low pass filtered height rate

CarlOlsson:
This PR scales the wind process noise with low pass filtered height rate in order to faster learn the wind states at high altitude gradients

image


CarlOlsson:
Would be cool if someone could merge or comment here

Issue #545, opened 2019-01-01
TECS throttle and pitch calculation have got some quirks in them


almaaro:
Hello, 
I'll explain some parts of throttle and pitch calculation in tecs.cpp (at least how I've understood it):

`// Calculate the demanded true airspeed`
`_update_speed_setpoint();`
- This calculates _TAS_rate_setpoint from the airspeed error and _speed_error_gain.

`// Calculate the demanded height`
`_update_height_setpoint(hgt_setpoint, baro_altitude);`
- This calculates _hgt_rate_setpoint from the altitude error and _height_error_gain and the demanded climb/descend rate and _height_setpoint_gain_ff
1. quirk: why is _height_setpoint_gain_ff even a parameter? Why it should be something else than 1? If it is not 1, it will always result in a constant altitude error during climbs/descends. 

`// Calculate the specific energy values required by the control loop`
`_update_energy_estimates();`
- This just calculates height, airspeed, height rate and airspeed rate setpoints to specific energies.

`// Calculate the throttle demand`
`_update_throttle_setpoint(throttle_cruise, rotMat);`
- First it calculates the predicted throttle from the STE_rate_setpoint.
2. quirk: 
`// Calculate gain scaler from specific energy error to throttle
float STE_to_throttle = 1.0f / (_throttle_time_constant * (_STE_rate_max - _STE_rate_min));`
- If I am getting it right, STE_to_throttle is just a proportional term to change STE_error to throttle output assuming a linear relationship between throttle and the STE rate. But this assumes that _STE_rate_max is achieved at 100% throttle, not throttle_max.

3. quirk    
    `// Add proportional and derivative control feedback to the predicted throttle and constrain to throttle limits
_throttle_setpoint = (_STE_error + _STE_rate_error * _throttle_damping_gain) * STE_to_throttle + throttle_predicted;`
- This is taking the _STE_error again into account in throttle calculation. _update_speed_setpoint() and _update_height_setpoint() already converted the airspeed and altitude errors into rate setpoints, which was already used for throttle calculation.

- After this the integrator term is added.
- My suggestion would be to replace STE_to_throttle with just 1.0f / _throttle_time_constant and to add the derivative and integrative terms (the _throttle_integ_state would need to be calculated before this) directly into STE_rate_setpoint like this:
`_throttle_integ_state = _throttle_integ_state + (_STE_error * _integrator_gain) * _dt;`

[...]

PR #562, opened 2019-01-26
TECS throttle and pitch tuning algorithm improvements

almaaro:
Hello,

**Motivation**
The TECS algorithm is working fine. However, the tuning is sub-optimal. The pitch and throttle calculation variables are more or less dependent on each other. For example, the SKE weighting interferes with the height rate setpoints (as well as with the max climb/descend rates) and the pitch time constant has an added effect with the height rate gain. More on the quirks: https://github.com/PX4/ecl/issues/545

**Idea** 
So this basically consists of 3 parts:
1. The desired specific energy quantity rates for height and airspeed are calculated only in _update_height_setpoint() and _update_speed_setpoint() (no intereference from _update_throttle_setpoint() or _update_pitch_setpoint())

2. Updated pitch calculation
- The STE and SKE weights will be limited to a maximum of 1 (for example if SKE weighting is set to 0.5 then SPE weighting will be 1, not 1.5) to allow the SKE weighting be adjustable [0,1] while the SPE weighting is 1 (to avoid interference with the height setpoint gains) and vice versa.
- As menitioned above, _update_pitch_setpoint() doesn't calculate any height rate setpoints (omitted the pitch time constant, which is basically a proportional gain to calculate the height rate setpoint from the height error).
- To take the effect of the SPE weighting into account, calculate _SPE_rate_setpoint_pitch which gives out the actual height rate setpoint (for the throttle calculation).

3. Updated thottle calculation
- As with the pitch, _update_throttle_setpoint() doesn't calculate the desired STE rate by any means
- The throttle time constant is omitted, because the throttle calculation only utilizes the feed forward, proportional error and integral error terms of the STE rate setpoint (and compares them to the max and min STE rate limits to interpolate the throttle).

**Considerations**
- This effects a lot of the parameters in FW configuration making it harder to integrate with the older systems.
- Test data is waiting for the snow to melt ;)

Best regards,
almaaro


priseborough:
Please resubmit when you have rebased and removed the commits not related to the TECS improvements.

Also it must to be back compatible with exisiting tuning parameters and cannot be accepted until it is so a way to achieve this needs to be found.

Thanks.

PR #563, opened 2019-01-29
TECS underspeed respect max throttle parameter

dagar:
A small standalone fix from @almaaro's pull request. https://github.com/PX4/ecl/pull/538

> Also changed the throttle setpoint in underspeed conditions from 100% to the user defined max value. No more burnt ESCs.

### DIFF
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -315,8 +315,8 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 
 	// Calculate the throttle demand
 	if (_underspeed_detected) {
-		// always use full throttle to recover from an underspeed condition
-		_throttle_setpoint = 1.0f;
+		// always use max throttle to recover from an underspeed condition
+		_throttle_setpoint = _throttle_setpoint_max;
 
 	} else {
 		// Adjust the demanded total energy rate to compensate for induced drag rise in turns.


priseborough:
In the past I have used the PWM max value to prevent ESC's burning. The reason for that is that is otherwise possible to burn them out in manual mode.

I recommend that that be the approach used, not this fix. 

That and the the desire to use 'reserve' power was the reason for use of 100% throttle. This change would leave a user with reduction in stall or under-speed performance due to altitude  or low battery.

sfuhrer:
@priseborough this emerged again with https://github.com/PX4/Firmware/pull/15321, and I think many are confused that the throttle is allowed above that value. Adjusting the limit for the PWM is ofc feasible, but not very practical, plus setting FW_THR_MAX is just too temping in that case.

Lets first align on what FW_THR_MAX actually stands for: Currently it is mainly used as a tuning parameter inside TECS, as the "trim throttle needed for max climbing at FW_T_CLMB_MAX". So if people want to reduce the max climbing speed they reduce FW_T_CLMB_MAX, and then also need to reduce FW_THR_MAX, right? 

So if then we would use FW_THR_MAX for emergencies and not 100% like now, we could run into issues. To avoid that we would almost need to a second parameter, like FW_T_CLMB_THR, and then we could use FW_THR_MAX as a hard limit to set the throttle to during underspeed, plus also in manual flight. 
What's your opinion?

PR #570, opened 2019-02-04
EKF: use hagl estimate if valid for when to trigger yaw reset on takeoff

CarlOlsson:


priseborough:
Code looks OK, but where is the test log?

CarlOlsson:
here is a sitl logfile
https://logs.px4.io/plot_app?log=8adde0ee-f540-4764-8328-2afdd6a7bc5b

CarlOlsson:
@priseborough

PR #574, opened 2019-02-18
[WIP] EKF: Not require a mag sensor data to initialize if fusion is disabled

ArkadiuszNiemiec:
I have a board without a mag sensor and want to fly indoors with the yaw data form external vision system. This change does not require data from mag sensor to initialize the ekf if the `EKF2_MAG_TYPE` is set to `5 (None)`. The problem was described in this issue: https://github.com/PX4/Firmware/issues/11359

Tested on current PX4 master in flight with LIDAR, flow and external vision system.


ArkadiuszNiemiec:
It needs more testing and discussions, currently generating local.pos x and y as 0.0.

Issue #575, opened 2019-02-27
Only 3D mag fusion is broken


CarlOlsson:
If `EKF2_MAG_TYPE = 2` (only 3D mag fusion) the estimator will continuously trigger yaw resets
https://github.com/PX4/ecl/blob/1378ec179717448404c74819871d149217350b19/EKF/control.cpp#L1542-L1545

since mag_align_complete is set to false here
https://github.com/PX4/ecl/blob/1378ec179717448404c74819871d149217350b19/EKF/control.cpp#L1378-L1378

To reproduce run
`make px4_sitl gazebo`, set `EKF2_MAG_TYPE = 2` and restart simulation

![image](https://user-images.githubusercontent.com/7795133/53489382-57402000-3a91-11e9-88ea-2f418edab1de.png)
![image](https://user-images.githubusercontent.com/7795133/53489415-69ba5980-3a91-11e9-9903-cca3a2dc80fd.png)

dagar:
Any idea when it was last functional?

CarlOlsson:
No, I never used it before. Just wanted to test something in SITL and noticed that the vehicle never got GPS lock

PR #578, opened 2019-03-01
EKF: Fix bug causing continual yaw reset when EKF2_MAG_TYPE = 2

priseborough:
Fixes https://github.com/PX4/ecl/issues/575

SITL test log here: https://logs.px4.io/plot_app?log=2e403ed3-2189-443b-b72d-504a9541ac1d with EKF2_MAG_TYPE = 2

Print statements at each call to resetMagHeading() were use to verify that reset only occurred at the following events:

1. State vector initialisation (does not increment quaternion reset count)
2. Completion of tilt alignment
3. Commencement of GPS aiding when local declination was available
4. When 1.5 metres was cleared after takeoff

Reset Counter:
screen shot 2019-03-02 at 9 09 11 am

Earth Field States:
screen shot 2019-03-02 at 9 09 38 am

XYZ Innovations:
screen shot 2019-03-02 at 9 10 17 am

Height:
screen shot 2019-03-02 at 9 11 19 am

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1533,7 +1533,7 @@ void Ekf::controlMagFusion()
 			}
 
 		} else if (_params.mag_fusion_type == MAG_FUSE_TYPE_3D) {
-			if (!_control_status.flags.mag_3D && _control_status.flags.yaw_align && (_flt_mag_align_start_time > 0)) {
+			if (!_control_status.flags.mag_3D && _control_status.flags.yaw_align) {
 				// only commence 3-axis fusion when yaw is aligned and field states set
 				_control_status.flags.mag_3D = true;
 			}
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1533,16 +1533,10 @@ void Ekf::controlMagFusion()
 			}
 
 		} else if (_params.mag_fusion_type == MAG_FUSE_TYPE_3D) {
-			// if transitioning into 3-axis fusion mode, we need to initialise the yaw angle and field states
-			if (!_control_status.flags.mag_3D || !_control_status.flags.mag_align_complete) {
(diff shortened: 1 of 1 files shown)


EKF/control.cpp
```diff
@@ -1538,16 +1538,10 @@ void Ekf::controlMagFusion()
 			}
 
 		} else if (_params.mag_fusion_type == MAG_FUSE_TYPE_3D) {
-			// if transitioning into 3-axis fusion mode, we need to initialise the yaw angle and field states
-			if (!_control_status.flags.mag_3D || !_control_status.flags.mag_align_complete) {
-				_control_status.flags.mag_align_complete = resetMagHeading(_mag_sample_delayed.mag);
-				_control_status.flags.yaw_align = _control_status.flags.yaw_align || _control_status.flags.mag_align_complete;
+			if (!_control_status.flags.mag_3D && _control_status.flags.yaw_align && (_flt_mag_align_start_time > 0)) {
```
CarlOlsson: Why do you need to check for `_flt_mag_align_start_time > 0` here?

CarlOlsson: @priseborough

priseborough: Originally it was to prevent the use of 3-axis mag fusion when on the ground whichcan casue problems due to ground based magnetic anomalies.

CarlOlsson: `_flt_mag_align_start_time` is set to the [imu time on bootup](https://github.com/PX4/ecl/blob/cd2e0739c6151ab9f444fd5ec190c89610db8279/EKF/ekf_helper.cpp#L713-L713) the [first time we trigger resetMagheading()](https://github.com/PX4/ecl/blob/cd2e0739c6151ab9f444fd5ec190c89610db8279/EKF/ekf.cpp#L265-L265) so the check here will never be false

priseborough: I'll fix that and rebase.

Issue #581, opened 2019-03-05
Unused parameters warning


fvantienen:
We now need to compile with `-Wno-unused-parameter` in order to not get any warnings due to the `estimator_interface.h` having functions with unused parameters. Wouldn't it be nicer to define the `__attribute__((unused))` for these parameters specifically?
This way we don't need to compile the whole autopilot project with the `unused-parameter` flag.

dagar:
Are these parameters completely unused in ecl or only in your particular use case?
Are you building with cmake?

fvantienen:
These following functions have unused parameters:
- `virtual bool collect_gps(uint64_t time_usec, struct gps_message *gps) { return true; }` (`estimator_interface.h`, all parameters)
- `bool Ekf::collect_gps(uint64_t time_usec, struct gps_message *gps)` (`gps_checks.cpp`, only `time_usec`)

I am not building with cmake, we use our own Makefile system.

dagar:
Let's just update them. I'll fix the PX4/Firmware usage.

fvantienen:
You can then also remove the `-Wno-unused-parameter` from the cmake file in this repo.

PR #584, opened 2019-03-05
EKF drop unused timestamp from collect_gps() and pass data by const reference

dagar:
- fixes https://github.com/PX4/ecl/issues/581
 - PX4/Firmware side https://github.com/PX4/Firmware/pull/11594

 @fvantienen

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,6 @@ if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
 			-Wextra
 			-Werror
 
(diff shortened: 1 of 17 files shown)


fvantienen:
The only warning I still get now is:
```
estimator_interface.h:172:46: warning: unused parameter ‘gps’ [-Wunused-parameter]
  virtual bool collect_gps(const gps_message &gps) { return true; }
```

But next to that it works fine.

dagar:
Ok, I think I've got them all now. Please review.

fvantienen:
Tested and works!

PR #585, opened 2019-03-07
ground effect: removed dependency on local position

RomanBapst:
Ground effect compensation activation logic for systems without a range finder has been implemented in
https://github.com/PX4/Firmware/pull/11592

However, the ekf would deactivate ground effect compensation when the vehicle was in_air and not close to the takeoff altitude. This means that ground effect compensation did not work if the user took off from some altitude and tried to land the drone at a higher altitude.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -985,12 +985,9 @@ void Ekf::controlHeightFusion()
 				_hgt_sensor_offset = 0.0f;
 			}
 
-			// Turn off ground effect compensation if it times out or sufficient height has been gained
-			// since takeoff.
+			// Turn off ground effect compensation if it times out
 			if (_control_status.flags.gnd_effect) {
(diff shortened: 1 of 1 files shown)


RomanBapst:
@priseborough I think we could introduce this logic again as soon as we have explicit knowledge about the flight stage of the vehicle (e.g. when a takeoff takes place). However, I would prefer to handle this outside of the EKF.

RomanBapst:
Idea: If ground effect flag set before takeoff then wait for sufficient height to be gained.

priseborough:
Can we change the logic so that if the ground effect flag has been set before takeoff the height check is applied, but not if it has been set afterwards.

priseborough:
I will merge this given that 
https://github.com/PX4/Firmware/pull/11592 is now in the flight testing stage.

PR #586, opened 2019-03-08
EKF: fix covariance and output filter buffer initialization

CarlOlsson:
This PR fixes several issues that were present in the initialization of the attitude covariances and the output filter buffer.

- Issue 1: The attitude covariances were initialized [before the attitude](https://github.com/PX4/ecl/blob/a85d3a43edc4ac4b85ae21159510b9c26f7359e6/EKF/ekf.cpp#L235-L253) which does not work properly since the initialization is [dependent on the attitude](https://github.com/PX4/ecl/blob/a85d3a43edc4ac4b85ae21159510b9c26f7359e6/EKF/ekf_helper.cpp#L1462-L1478).
- Issue 2: After the attitude was initialized we [rotate the yaw angle](https://github.com/PX4/ecl/blob/a85d3a43edc4ac4b85ae21159510b9c26f7359e6/EKF/ekf.cpp#L265-L265). In this function we also rotated all attitude solutions in the output buffer. However, the data in the buffer did not include the tilt rotation. So the "yaw rotation" was applied along a completely different axis.
- Issue 3: At the end of the initialization we [align the output filter](https://github.com/PX4/ecl/blob/a85d3a43edc4ac4b85ae21159510b9c26f7359e6/EKF/ekf_helper.cpp#L379-L379) to match the state estimate at the delayed time horizon. However, the rotation was applied in the wrong direction.
- Issue 4: The `_output_new` quaternion was [not initialized to a valid quaternion](https://github.com/PX4/ecl/blob/a85d3a43edc4ac4b85ae21159510b9c26f7359e6/EKF/ekf.cpp#L62-L62)

This PR changes the EKF initialization to:
1) init tilt
2) init yaw
3) init tilt uncertainty
4) init yaw uncertainty

This have quite large consequences. Below is a logfile with high temperature gradient replayed with master code and this PR.

On master the delta velocity bias states diverge before takeoff triggering a reset of the covariance matrix while on the PR they are stable

**master/PR**
image

This is because on the PR the uncertainty in the delta velocity bias aligned with gravity decrease much faster since this axis is more observable than the axes aligned with the horizontal plane.

**master/PR**
Note that this plot is from a tailsitter with body x axis facing upwards
image

This is a plot comparing the attitude estimate at the current time and at the delayed time horizon right after the EKF initialization. (This is normally not logged since the ekf2_main.cpp only logs the attitude estimate when tilt is aligned. So for testing the attitude estimate at the current time is added to estimator status)

**master/PR**
image

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -60,6 +60,7 @@ bool Ekf::init(uint64_t timestamp)
 	_output_new.vel.setZero();
 	_output_new.pos.setZero();
 	_output_new.quat_nominal.setZero();
+	_output_new.quat_nominal(0) = 1.0f;
 
 	_delta_angle_corr.setZero();
 	_imu_down_sampled.delta_ang.setZero();
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -256,13 +256,10 @@ bool Ekf::initialiseFilter()
 		// update transformation matrix from body to world frame
 		_R_to_earth = quat_to_invrotmat(_state.quat_nominal);
 
-		// calculate the averaged magnetometer reading
-		Vector3f mag_init = _mag_filt_state;
-
 		// calculate the initial magnetic field and yaw alignment
 		// Get the magnetic declination
 		calcMagDeclination();
-		_control_status.flags.yaw_align = resetMagHeading(mag_init);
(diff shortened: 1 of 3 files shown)


mrivi:
I have noticed a big drift in the z position when leaving the simulation running for a couple of minutes with the vehicle on the ground unarmed. This PR fixes this issue. 

Current master:
z_drift

This PR:
no_drift

PR #588, opened 2019-03-09
Add unit tests for DataValidator

tstellanova:
Added a few unit tests for the `DataValidator` class that cover happy path and common failure paths. In some cases the existing interface lacks documentation of expected behavior (for example `DataValidator::confidence` has some undocumented side-effects, setting `_error_mask`, and `put` performs a variety of calculations), so these tests codify some of the "as-built" behavior of the class.  I did not yet attempt to test the vibration-calculating code, which is unspecified but appears to be key to the operation of `DataValidatorGroup`.

### DIFF
--- a/validation/CMakeLists.txt
+++ b/validation/CMakeLists.txt
@@ -44,3 +44,4 @@ if(BUILD_TESTING)
 
 	add_subdirectory(tests)
 endif()
+
--- a/validation/tests/CMakeLists.txt
+++ b/validation/tests/CMakeLists.txt
@@ -1,6 +1,6 @@
 ############################################################################
 #
-#   Copyright (c) 2018 ECL Development Team. All rights reserved.
+#   Copyright (c) 2015-2018 ECL Development Team. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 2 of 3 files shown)


dagar:
Looks like a good start.

Not that we've defined this anywhere, but would you mind running it through PX4's astyle? 
From Firmware `./Tools/astyle/fix_code_style.sh validation/tests/*.cpp`.

priseborough:
These changes appear to be causing a problem with PX4/Firmware CI testing.

http://ci.px4.io:8080/blue/organizations/jenkins/PX4_misc%2FFirmware-compile_mac/detail/PR-11630/2/pipeline

tstellanova:
It looks like they are passing in this CI:
https://codecov.io/gh/PX4/ecl/src/master/validation/tests/test_data_validator.cpp
I'm not familiar with the other CI. @dagar Any recommendations on how to ensure this test binary builds in _that_ CI ?

tstellanova:
My initial thought is that it appears the `ci.px4.io` CI build is treating the `ecl_test_data_validator` binary as one of the sitl unit tests, similar to eg `autodeclination`. Not sure why it's deciding to do that.

dagar:
I'll limit the ecl tests to the standalone build. https://github.com/PX4/ecl/pull/591

PR #590, opened 2019-03-12
EKF: Fix innovation in fuseDeclination()

CarlOlsson:
Before we only calculated the `_mag_declination` variable before mag fusion, leading to the fact that in fuseDeclination() the innovation was not zero (since mag fusion updated the states) even though mag_align_complete was true.

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -550,7 +550,7 @@ private:
 	// It is used to align the yaw angle after launch or takeoff for fixed wing vehicle.
 	bool realignYawGPS();
 
-	// calculate the magnetic declination to be used by the alignment and fusion processing
+	// Return the magnetic declination in radians to be used by the alignment and fusion processing
 	float getMagDeclination();
 
 	// reset position states of the ekf (only horizontal position)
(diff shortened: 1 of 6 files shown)


EKF/ekf.h
```diff
@@ -553,7 +551,7 @@ class Ekf : public EstimatorInterface
 	bool realignYawGPS();
 
 	// calculate the magnetic declination to be used by the alignment and fusion processing
```
priseborough: Now that this is returning the value, please ensure the header documentation indicates the units, eg radians

EKF/ekf_helper.cpp
```diff
@@ -767,28 +767,28 @@ bool Ekf::resetMagHeading(Vector3f &mag_init, bool increase_yaw_var, bool update
 }
 
 // Calculate the magnetic declination to be used by the alignment and fusion processing
```
priseborough: Suggest

// Return the magnetic declination in radians to be used by the alignment and fusion processing.

EKF/ekf.h
priseborough: Suggest:

// Returns the magnetic declination in radians to be used by the alignment and fusion processing.

PR #597, opened 2019-04-14
Implemented use of optical flow for terrain estimation

RomanBapst:
This PR allows optical flow measurements to be used for terrain estimation. I see the following two cases when this could be useful:
- terrain estimation for fixed wing flight and auto land (potentially using the flow sensor at higher altitude and a lidar when close to the ground)
@priseborough already did this in another project and I believe he had good success with the [pmw3901](https://www.seeedstudio.com/Flow-Breakout-Board-p-2949.html) optical flow sensor.
- enhancement of the detection logic for ground effect (see https://github.com/PX4/Firmware/pull/11592 )

Flow measurements will only be fused in the terrain estimator if no range data is being fused and if the primary EKF is not fusing flow.
So far this has only been tested in SITL using iris_opt_flow model. Two logs are being provided, one where the terrain estimate is initialized to 50m on the ground in order to see if the filter recovers.
The second log shows a normal flight in position control over a flat surface. As the vehicle starts to move the terrain height becomes observable and the estimated terrain variance decreases. When the vehicle holds position the terrain variance increases again.

**Terrain initialized to 50m** 
terrain_flow

**Normal flight over flat ground**
terrain_flow2

**Logs:**
Terrain initialized to 50m:
https://logs.px4.io/plot_app?log=e47c69ed-125c-4a82-a1df-dc0553904b63

Normal flight:
https://logs.px4.io/plot_app?log=a3864090-d60c-42af-8fc8-f8465ded7790

**How to test:**
Currently this feature can only be tested if optical flow in the main filter is deactivated and if no range data is available. Therefore, the following two steps are required for testing:
1) Set EKF2_AID_MASK to only use GPS
2) Comment out [this line](https://github.com/PX4/Firmware/blob/master/src/modules/simulator/simulator_mavlink.cpp#L312) to avoid the system publishing range finder data.

**Required action items:**

- [x]  test on real vehicle (multicopter and fixed wing)
- [ ] - make sure we do not break existing optical flow & terrain estimation via lidar
- [x] - potentially adjust innovation gate size, optical flow measurement noise for robustness

### DIFF
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -56,7 +56,8 @@ bool Ekf::initHagl()
 		// success
 		return true;
 	} else if (_flow_for_terrain_data_ready) {
-		_terrain_vpos = _state.pos(2);
+		// initialise terrain vertical position to origin as this is the best guess we have
+		_terrain_vpos = fmaxf(0.0f,  _state.pos(2));
 		_terrain_var = 100.0f;
 		return true;
 	} else if (!_control_status.flags.in_air) {
@@ -213,6 +214,9 @@ void Ekf::fuseFlowForTerrain()
 
 	// Calculate observation matrix for flow around the vehicle x axis
 	float Hx = vel_body(1) * t0 /(pred_hagl * pred_hagl);
+
+	// Constrain terrain variance to be non-negative
+	_terrain_var = fmaxf(_terrain_var, 0.0f);
 
 	// Cacluate innovation variance
 	_flow_innov_var[0] = Hx * Hx * _terrain_var + R_LOS;
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -112,7 +112,7 @@ void Ekf::runTerrainEstimator()
 			_flow_for_terrain_data_ready = false;
 		}
 
-		//constrain _terrain_vpos to be a minimum of _params.rng_gnd_clearance larger than _state.pos(2)
+		// constrain _terrain_vpos to be a minimum of _params.rng_gnd_clearance larger than _state.pos(2)
 		if (_terrain_vpos - _state.pos(2) < _params.rng_gnd_clearance) {
 			_terrain_vpos = _params.rng_gnd_clearance + _state.pos(2);
 		}
@@ -207,8 +207,12 @@ void Ekf::fuseFlowForTerrain()
 
 	float t0 = q0*q0 - q1*q1 - q2*q2 + q3*q3;
 
+	// constrain terrain to minimum allowed value and predict height above ground
+	_terrain_vpos = fmaxf(_terrain_vpos, _params.rng_gnd_clearance + _state.pos(2));
+	float pred_hagl = _terrain_vpos - _state.pos(2);
+
 	// Calculate observation matrix for flow around the vehicle x axis
-	float Hx = vel_body(1) * t0 /((_terrain_vpos - _state.pos(2)) * (_terrain_vpos - _state.pos(2)));
+	float Hx = vel_body(1) * t0 /(pred_hagl * pred_hagl);
 
 	// Cacluate innovation variance
 	_flow_innov_var[0] = Hx * Hx * _terrain_var + R_LOS;
@@ -217,7 +221,7 @@ void Ekf::fuseFlowForTerrain()
 	float Kx = _terrain_var * Hx / _flow_innov_var[0];
 
 	// calculate prediced optical flow about x axis
-	float pred_flow_x = vel_body(1) * earth_to_body(2,2) / (_terrain_vpos - _state.pos(2));
+	float pred_flow_x = vel_body(1) * earth_to_body(2,2) / pred_hagl;
 
 	// calculate flow innovation (x axis)
 	_flow_innov[0] = pred_flow_x - opt_flow_rate(0);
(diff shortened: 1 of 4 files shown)


EKF/terrain_estimator.cpp
```diff
@@ -162,6 +170,100 @@ void Ekf::fuseHagl()
...
+
+	// do not perform measurement update if badly conditioned
+	if (flow_test_ratio <= 1.0f) {
+		_terrain_vpos += Kx * _flow_innov[0];
+		// guard against negative variance
+		_terrain_var = fmaxf(_terrain_var - KxHxP, 0.0f);
+	}
+
+	// Calculate observation matrix for flow around the vehicle y axis
+	float Hy = -vel_body(0) * t0 /((_terrain_vpos - _state.pos(2)) * (_terrain_vpos - _state.pos(2)));
```
RomanBapst: Guard against division by zero

RomanBapst:
Results from test flight on a fixed wing (Quanum observer, cruise speed 15 m/s), this was a flight over more or less flat terrain.
image

This is a zoom-in into the region when the vehicle landed. Noticed that at the time of touch-down the estimated distance to bottom is about 26cm (indicated by the spike in the accelerometer).
image

RomanBapst:
Results from a flight over a hill with altitudes above 100m.
image

RomanBapst:
Replay log for flat terrain:
https://logs.px4.io/plot_app?log=925aeea4-6679-4452-904e-1a1144e45bc2

Replay log for flight over hill:
https://logs.px4.io/plot_app?log=869142b1-7eeb-41d5-828a-3f7fefd848fe

**NOTE**
The optical flow orientation on these flight was not correct and need to be adjusted if you want to replay the logs. You can add the following code [here](https://github.com/PX4/ecl/blob/master/EKF/estimator_interface.cpp#L411)

```
float helper = optflow_sample_new.flowRadXY(0);
optflow_sample_new.flowRadXY(0) = -optflow_sample_new.flowRadXY(1);
optflow_sample_new.flowRadXY(1) = helper;
```

RomanBapst:
This is a replayed log file of when the plane was flying over a hill:
https://logs.px4.io/plot_app?log=7b365163-79fc-46e7-8615-1b46d08db4a6

EKF/terrain_estimator.cpp
```diff
@@ -55,7 +55,10 @@ bool Ekf::initHagl()
 		_terrain_var = sq(_params.range_noise);
 		// success
 		return true;
-
+	} else if (_flow_for_terrain_data_ready) {
+		_terrain_vpos = _state.pos(2);
```
priseborough: This is a guess unless we are on the ground. Might be better to set to height of origin as that is where terrain is most likely to be after takeoff.

priseborough: If on ground, using current position is better.

EKF/terrain_estimator.cpp
```diff
@@ -162,6 +170,104 @@ void Ekf::fuseHagl()
...
+	// Note: _flow_sample_delayed.gyroXYZ is the negative of the body angular velocity, thus use minus sign
+	Vector3f vel_rel_imu_body = cross_product(-_flow_sample_delayed.gyroXYZ / _flow_sample_delayed.dt, pos_offset_body);
+
+	// calculate the velocity of the sensor in the earth frame
+	Vector3f vel_rel_earth = _state.vel + _R_to_earth * vel_rel_imu_body;
+
+	// rotate into body frame
+	Vector3f vel_body = earth_to_body * vel_rel_earth;
+
+	float t0 = q0*q0 - q1*q1 - q2*q2 + q3*q3;
```
priseborough: Which derivation did you use to obtain this? If you have a Matlab file, add it here and provide a link to it in the documentation: https://github.com/PX4/ecl/tree/master/EKF/matlab/scripts/Terrain%20Estimator

EKF/terrain_estimator.cpp
```diff
@@ -162,6 +170,104 @@ void Ekf::fuseHagl()
...
+
+	// constrain terrain to minimum allowed value and predict height above ground
+	_terrain_vpos = fmaxf(_terrain_vpos, _params.rng_gnd_clearance + _state.pos(2));
+	float pred_hagl = _terrain_vpos - _state.pos(2);
+
+	// Calculate observation matrix for flow around the vehicle x axis
+	float Hx = vel_body(1) * t0 /(pred_hagl * pred_hagl);
+
+	// Cacluate innovation variance
+	_flow_innov_var[0] = Hx * Hx * _terrain_var + R_LOS;
```
priseborough: Add protection against Hx * Hx * _terrain_var being less then zero.

EKF/terrain_estimator.cpp
```diff
@@ -162,6 +170,104 @@ void Ekf::fuseHagl()
...
+		_terrain_vpos += Kx * _flow_innov[0];
+		// guard against negative variance
+		_terrain_var = fmaxf(_terrain_var - KxHxP, 0.0f);
+	}
+
+	// Calculate observation matrix for flow around the vehicle y axis
+	float Hy = -vel_body(0) * t0 /(pred_hagl * pred_hagl);
+
+	// Calculuate innovation variance
+	_flow_innov_var[1] = Hy * Hy * _terrain_var + R_LOS;
```
priseborough: Add protection against Hy * Hy * _terrain_var being less then zero.

priseborough: _terrain_var being protected against being negative above so should be OK.

EKF/terrain_estimator.cpp
priseborough: There are many places in code that touch _terrain_var and it is not forced to the non-negative immediately prior.

EKF/terrain_estimator.cpp
RomanBapst: Done

EKF/terrain_estimator.cpp
RomanBapst: @priseborough I added a python folder which has the script for the derivation.

EKF/terrain_estimator.cpp
RomanBapst: Done

EKF/terrain_estimator.cpp
RomanBapst: Done

RomanBapst:
@priseborough I addressed your comments.

PR #599, opened 2019-05-01
Fix up link to EKF docs

hamishwillee:
Once user guide became versioned, this was broken.


hamishwillee:
Thanks. @mhkabir If you can merge, please do so. I don't have rights here.

PR #600, opened 2019-05-01
tecs: rename variables

CarlOlsson:
comes together with https://github.com/PX4/Firmware/pull/11944

no functional change

* add const
* rename variables to match the updated tecs_status message

### DIFF
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -74,7 +74,7 @@ void TECS::update_vehicle_state_estimates(const float equivalent_airspeed, const
 	}
 
 	if (reset_altitude) {
-		_vert_pos_state = altitude;
+		_altitude_state = altitude;
 
 		if (vz_valid) {
 			_vert_vel_state = -vz;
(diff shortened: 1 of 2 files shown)


tecs/tecs.cpp
```diff
@@ -361,12 +361,12 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 		if (_integrator_gain > 0.0f) {
 			// Calculate throttle integrator state upper and lower limits with allowance for
 			// 10% throttle saturation to accommodate noise on the demand.
-			float integ_state_max = _throttle_setpoint_max - _throttle_setpoint + 0.1f;
-			float integ_state_min = _throttle_setpoint_min - _throttle_setpoint - 0.1f;
+			const float integ_state_max = _throttle_setpoint_max - _throttle_setpoint + 0.1f;
+			const float integ_state_min = _throttle_setpoint_min - _throttle_setpoint - 0.1f;
 
 			// Calculate a throttle demand from the integrated total energy error
```
priseborough: I would like to see the term  'specific total energy' used consistently in the comments.

tecs/tecs.cpp
```diff
@@ -400,19 +400,19 @@ void TECS::_detect_uncommanded_descent()
 {
 	/*
 	 * This function detects a condition that can occur when the demanded airspeed is greater than the
-	 * aircraft can achieve in level flight. When this occurs, the vehicle will continue to reduce height
+	 * aircraft can achieve in level flight. When this occurs, the vehicle will continue to reduce altitude
 	 * while attempting to maintain speed.
 	*/
 
 	// Calculate rate of change of total specific energy
```
priseborough: 'specific total energy'

RomanBapst:
@CarlOlsson Overall this looks good to me. For me it's not entirely clear why some variables are written in capital letters (_SPE_rate_setpoint) and some variables not (_total_energy_rate_error).
What could also be confusing about this is that the former variable has "specific" in it's name but the latter one not.

`float SEB_correction = _energy_distribution_error + _energy_distribution_rate_error * _pitch_damping_gain + SEB_rate_setpoint * _pitch_time_constant;`

Same applies for the line above. I think we should either speak of "energy balance" or "energy distribution" but not mix both together.

RomanBapst:
@CarlOlsson By the way if this is good to merge I'm happy to test in on my fixed wing. I anyway wanted to try auto landing using flow for terrain.

CarlOlsson:
I agree with all comments. The main reason I did the PR was that the logged fields in the ulog file should match the variable names in TECS.
I won't have time to address the comments in the coming days but it should be safe to flight test at least

priseborough:
The capital letters are where individual words have been abbreviated, eg SEB is specific energy balance, SKE is specific kinetic energy, etc. The definitions are contained in the header file and include the units.

CarlOlsson:
The problem with capital letters is that you can't have them capital in the topic. But I guess we could just make them lower case instead.

RomanBapst:
@CarlOlsson Ah I see, that was your intention. I would make them small letter in the topic.

priseborough:
@CarlOlsson and @dagar I'm OK with this being merged provided https://github.com/PX4/Firmware/pull/11944 is also OK.

priseborough:
@CarlOlsson to enable this to go in independent of https://github.com/PX4/Firmware/pull/11944 before it bit-rots, are you able to break it into two? The first with the inward facing changes that maintain interface compatibility that we can merge now and the second with the outward facing changes that require https://github.com/PX4/Firmware/pull/11944.

CarlOlsson:
Since I don't have time to work on this at the moment and there are several comments that needs to be addressed I close here for now. Will hopefully come back to it at some point.
Main issues to solve are
1) topic field names should match variable names in TECS
2) we need to explicitly state everywhere which airspeed that is used (EAS/TAS)

PR #601, opened 2019-05-08
Implemented Kahan summation algorithm for adding process noise

RomanBapst:
Each time there is a covariance matrix update we add process noise to the variance of each state.
For the delta angle and the delta velocity bias variances these contributions can be so small compared to the actual variance that they get lost due to numerical limitations when dealing with floating point variables.

One solution for this problem is to use the Kahan summation algorithm which relies on an accumulator variable which captures the small increments that would get lost otherwise.

For more details, see https://en.wikipedia.org/wiki/Kahan_summation_algorithm

Thanks to @priseborough for helping debug this issue and thanks to @jkflying for suggesting the algorithm.

Delta velocity z bias variance prior to this change:
image (4)

Delta velocity z bias variance after this change:
image (3)

I also compared the results of this PR against the previous code when the delta velocity bias noise was set very high (0.3 m/s/s/s) in order to see if both implementations lead to similar variances. The figure below shows that this was the case.

variance!UNITO-UNDERSCORE!diff

This is a replay log file from an S500 which has been used for validation:
https://logs.px4.io/plot_app?log=7ebfbbc0-d9ed-4e4f-92f0-e9ab73f0016f

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -54,6 +54,9 @@ void Ekf::initialiseCovariance()
 		}
 	}
 
+	_delta_angle_bias_var_accum.setZero();
+	_delta_vel_bias_var_accum.setZero();
+
 	// calculate average prediction time step in sec
 	float dt = FILTER_UPDATE_PERIOD_S;
 
@@ -430,8 +433,15 @@ void Ekf::predictCovariance()
 	nextP[12][12] = P[12][12];
 
 	// add process noise that is not from the IMU
-	for (unsigned i = 0; i <= 12; i++) {
+	for (unsigned i = 0; i <= 9; i++) {
 		nextP[i][i] += process_noise[i];
+	}
+
+	// process noise contribution for delta angle states can be very small compared to
+	// the variances, therefore use algorithm to minimise numerical error
+	for (unsigned i = 10; i <=12; i++) {
+		const int index = i-10;
+		nextP[i][i] = kahanSummation(nextP[i][i], process_noise[i], _delta_angle_bias_var_accum(index));
 	}
 
 	// Don't calculate these covariance terms if IMU delta velocity bias estimation is inhibited
@@ -485,14 +495,18 @@ void Ekf::predictCovariance()
 		nextP[15][15] = P[15][15];
 
 		// add process noise that is not from the IMU
+		// process noise contributiton for delta velocity states can be very small compared to
+		// the variances, therefore use algorithm to minimise numerical error
 		for (unsigned i = 13; i <= 15; i++) {
-			nextP[i][i] += process_noise[i];
+			const int index = i-13;
+			nextP[i][i] = kahanSummation(nextP[i][i], process_noise[i], _delta_vel_bias_var_accum(index));
 		}
 
 	} else {
 		// Inhibit delta velocity bias learning by zeroing the covariance terms
 		zeroRows(nextP, 13, 15);
 		zeroCols(nextP, 13, 15);
+		_delta_vel_bias_var_accum.setZero();
 	}
 
 	// Don't do covariance prediction on magnetic field states unless we are using 3-axis fusion
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -363,6 +363,9 @@ private:
 
 	float P[_k_num_states][_k_num_states] {};	///< state covariance matrix
 
+	Vector3f _delta_vel_bias_var_accum;		///< kahan summation algorithm accumulator for delta velocity bias variance
+	Vector3f _delta_angle_bias_var_accum;	///< kahan summation algorithm accumulator for delta angle bias variance
+
 	float _vel_pos_innov[6] {};	///< NED velocity and position innovations: 0-2 vel (m/sec),  3-5 pos (m)
 	float _vel_pos_innov_var[6] {};	///< NED velocity and position innovation variances: 0-2 vel ((m/sec)**2), 3-5 pos (m**2)
 	float _aux_vel_innov[2] {};	///< NE auxiliary velocity innovations: (m/sec)
@@ -706,5 +709,11 @@ private:
 
 	// uncorrelate quaternion states from other states
 	void uncorrelateQuatStates();
+
+	// Use Kahan summation algorithm to get the sum of "sum_previous" and "input".
+	// This function relies on the caller to be responsible for keeping a copy of
+	// "accumulator" and passing this value at the next iteration.
+	// Ref: https://en.wikipedia.org/wiki/Kahan_summation_algorithm
+	float kahanSummation(float sum_previous, float input, float &accumulator) const;
 
 };
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -1722,3 +1722,11 @@ void Ekf::save_mag_cov_data()
 		}
 	}
 }
+
+float Ekf::kahanSummation(float sum_previous, float input, float &accumulator) const
+{
+	float y = input - accumulator;
+	float t = sum_previous + y;
+	accumulator = (t - sum_previous) - y;
+	return t;
+}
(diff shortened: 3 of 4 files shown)


EKF/covariance.cpp
```diff
@@ -431,7 +431,12 @@ void Ekf::predictCovariance()
 
 	// add process noise that is not from the IMU
 	for (unsigned i = 0; i <= 12; i++) {
-		nextP[i][i] += process_noise[i];
+		if (i < 10) {
```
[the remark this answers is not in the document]
priseborough: Would be more efficient that way.

EKF/ekf.h
```diff
@@ -359,6 +359,9 @@ class Ekf : public EstimatorInterface
 
 	float P[_k_num_states][_k_num_states] {};	///< state covariance matrix
 
+	Vector3f _delta_vel_bias_var_accum;		///< kahan summation algorithm accumulator for delta velocity bias variance
```
priseborough: Should also explicitly zero these in the initialiseCovariance() function.

EKF/covariance.cpp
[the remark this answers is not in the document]
RomanBapst: Done

EKF/ekf.h
RomanBapst: Done

RomanBapst:
@jkflying @priseborough I addressed your comments.

EKF/covariance.cpp
```diff
@@ -430,8 +433,15 @@ void Ekf::predictCovariance()
 	nextP[12][12] = P[12][12];
 
 	// add process noise that is not from the IMU
-	for (unsigned i = 0; i <= 12; i++) {
-		nextP[i][i] += process_noise[i];
+	for (unsigned i = 0; i <= 9; i++) {
+			nextP[i][i] += process_noise[i];
+	}
+
+	// process noise contributiton for delta angle states can be very small compared to
```
CarlOlsson:
```suggestion
	// process noise contribution for delta angle states can be very small compared to
```

priseborough:
There are no issues with the code, however I am hesitant to merge this so close to a major release because it requires a change to the default parameter in ekf2 module which will need flight hours on different platforms to validate the new process noise value.

priseborough:
PX4/Firmware v1.9.0-rc2 has been tagged and from now on and i have been told it will now only update with bug fixes, so we can now carry on with getting a sensible value for the noise variance, include that default with this PR and also the corresponding one on the firmware side. I will do some replay testing tomorrow.

priseborough:
My testing on replay is showing that the kahanSummation function is not working as expected and the _delta_vel_bias_var_accum variable is always returning as zero.

See debug output added here: https://github.com/priseborough/ecl/tree/pr601

Replay with with EKF2_ACC_B_NOISE = 0.0006 gives:

Screen Shot 2019-05-24 at 11 47 10 am

At a larger values of EKF2_ACC_B_NOISE = 0.06 that just raise the variance off the lower limit I get the same variance before and after the PR changes are applied, eg:

Before:

before

After:

after

bkueng:
@priseborough in order to ensure the compiler does not optimize the algorithm away, we'll need to disable `-funsafe-math-optimizations` (in particular `-fassociative-math`, which is set by `-funsafe-math-optimizations`). It's here: https://github.com/PX4/Firmware/blob/master/cmake/px4_add_common_flags.cmake#L54

jkflying:
@bkueng Maybe another way would be to declare the accumulator `volatile`, or to add `__attribute__ ((optimize("no-fast-math"))) ` above the Kahan function declaration?

bkueng:
That's also fine with me, but I think we need to be very careful with enabling flags like `-fassociative-math` on a global scope.

jkflying:
It was probably added to reduce flash size, I know that often things like matrix calculations get enormously reduced using these flags. But there might be other cases where this is causing an issue, so yeah, we really need to be careful.

EKF/ekf_helper.cpp
```diff
@@ -1722,3 +1722,10 @@ void Ekf::save_mag_cov_data()
 		}
 	}
 }
+
+float Ekf::kahanSummation(float sum_previous, float input, float &accumulator) {
+	float y = input - accumulator;
```
jkflying: Otherwise
```suggestion
	volatile float y = input - accumulator;
```
and

EKF/ekf_helper.cpp
```diff
@@ -1722,3 +1722,10 @@ void Ekf::save_mag_cov_data()
...
+	float t = sum_previous + y;
```
jkflying:
```suggestion
	volatile float t = sum_previous + y;
```

priseborough:
I will try the method using volatiles in replay.

priseborough:
I have confirmed on replay that making the t and y floats volatiles fixes the problem reported. 

Screen Shot 2019-05-28 at 9 08 09 pm

dagar:
I'd rather explicitly disable the optimization with the attribute than use a volatile.

Yet another option would be moving `Ekf::kahanSummation` to a dedicated compilation unit and removing the optimization flag at the cmake level .

jkflying:
IMO the optimization flag itself is dangerous, but if it reduces binary size a lot and potentially increases performance I understand, even if it might be problematic here.

The attribute won't work on `clang` builds, and there is no equivalent way of doing it inline like that for `clang`. I really don't want there to be different behavior in something as fundamental as the ECL based on compiler used. It would be impossible to debug.

The CMake and different compilation unit will work, but then the kahan summation is scattered across a few different files, and it seems easy for a change to be made in the build system (CMake include orders, for example) which would silently break this. We don't have unit tests to make sure it is working, and if we did they would need to be in Release builds with all optimizations on for me to really trust them.

`volatile` is ugly from a "that's not what `volatile` is meant for" perspective, but seems to me the most likely to be robust, least effort, and won't silently break in the future.

I'm a bit torn here. If we aren't planning to support `clang`, then the attribute is better, but then we need to make it clear that `clang` is broken.

dagar:
I'm not opposed to dropping `-fassociative-math` either across the entire `ecl/EKF` or PX4-wide for that matter. The flash savings are minor and so far I'm not seeing any significant performance differences on the bench (obviously needs significantly more testing).

image

### Relevant GCC math optimizations

 * `-funsafe-math-optimizations` # Enables -fno-signed-zeros, -fno-trapping-math, -fassociative-math and -freciprocal-math
   * `-fno-signed-zeros`	# Allow optimizations for floating-point arithmetic that ignore the signedness of zero
   * `-fno-trapping-math`	# Compile code assuming that floating-point operations cannot generate user-visible traps
   * `-fassociative-math`	# Allow re-association of operands in series of floating-point operations
   * `-freciprocal-math`	# Allow the reciprocal of a value to be used instead of dividing by the value if this enables optimizations

  * `-fno-math-errno`		# Do not set errno after calling math functions that are executed with a single instruction, e.g., sqrt

LorenzMeier:
Yes, sounds like we should be dropping those in general.

dagar:
Here's a PX4/Firmware side PR for testing. https://github.com/PX4/Firmware/pull/12096

jkflying:
Ok, so how about we use the `volatile` as a temporary workaround, and once we've updated the flags in the firmware we can get rid of the `volatile`.

dagar:
> Ok, so how about we use the `volatile` as a temporary workaround, and once we've updated the flags in the firmware we can get rid of the `volatile`.

That's fine as long as we don't completely move on and forget about https://github.com/PX4/Firmware/pull/12096.

Here's the cmake level solution. https://github.com/PX4/ecl/blob/e4c31c30f746205a973dfae7aaef8dc5c3429d0c/EKF/CMakeLists.txt#L58

jkflying:
> Here's the cmake level solution.
> 
> https://github.com/PX4/ecl/blob/e4c31c30f746205a973dfae7aaef8dc5c3429d0c/EKF/CMakeLists.txt#L58

I think you mean `-fno-associative-math`?

dagar:
> > Here's the cmake level solution.
> > https://github.com/PX4/ecl/blob/e4c31c30f746205a973dfae7aaef8dc5c3429d0c/EKF/CMakeLists.txt#L58
> 
> I think you mean `-fno-associative-math`?

Yep :smirk:. It's just an example of how additional flags can be set per source file.

priseborough:
So we wait for the firmware change, ensure there are no unexpected effects, then merge this. My limited replay testing so far has shown that the exisiting tuning parameter is OK to use, but that was one flight log from one vehicle.

Issue #604, opened 2019-05-21
Error in ekf_helper.cpp


ecpknpaul:
Hi:
ekf_helper.cpp:
line 512:
`// Note q1 *= q2 is equivalent to q1 = q2 * q1 `
I think it should be `q1 *= q2 is equivalent to q1 = q1 * q2`, because in file Quaternion.hpp, line 272-276:

> void operator*=(const Quaternion &other)
 {
 Quaternion &self = *this;
 self = self * other;
 }

and I think in file ekf_helper.cpp, line 519
`_output_new.quat_nominal = _state_reset_status.quat_change * _output_new.quat_nominal;`
should be:
`_output_new.quat_nominal = _output_new.quat_nominal * _state_reset_status.quat_change;`

Am I right？Looking forward to your reply.

CarlOlsson:
Hi,

yes you are right! I forgot to fix those lines in [this PR](https://github.com/PX4/ecl/pull/586)
I did not notice since it was not visible in the logs since the error only affects one single output.
How did you find it? 
I pushed a fix [here](https://github.com/PX4/ecl/pull/605)

ecpknpaul:
Hi,
I just read the code,  and find the quaternion multiplication is in the wrong order.

priseborough:
Yes, the comment and code are inconsistent and we should be using the same convention for multiplication order as the Eigen library. Please provide log evidence of the error because if the order is wrong it should be visible in a log.

PR #605, opened 2019-05-21
 ekf: correct quaternion multiplication order

CarlOlsson:
fixes https://github.com/PX4/ecl/issues/604

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -378,8 +378,8 @@ void Ekf::resetHeight()
 // align output filter states to match EKF states at the fusion time horizon
 void Ekf::alignOutputFilter()
 {
-	// calculate the quaternion delta between the output and EKF quaternions at the EKF fusion time horizon
-	Quatf q_delta = _output_sample_delayed.quat_nominal.inversed() *  _state.quat_nominal;
+	// calculate the quaternion rotation delta from the EKF to output observer states at the EKF fusion time horizon
+	Quatf q_delta = _state.quat_nominal * _output_sample_delayed.quat_nominal.inversed();
 	q_delta.normalize();
 
 	// calculate the velocity and position deltas between the output and EKF at the EKF fusion time horizon
@@ -388,16 +388,16 @@ void Ekf::alignOutputFilter()
 
 	// loop through the output filter state history and add the deltas
 	for (uint8_t i = 0; i < _output_buffer.get_length(); i++) {
-		_output_buffer[i].quat_nominal *= q_delta;
+		_output_buffer[i].quat_nominal = q_delta * _output_buffer[i].quat_nominal;
 		_output_buffer[i].quat_nominal.normalize();
 		_output_buffer[i].vel += vel_delta;
 		_output_buffer[i].pos += pos_delta;
 	}
 
-	_output_new.quat_nominal *= q_delta;
+	_output_new.quat_nominal = q_delta * _output_new.quat_nominal;
 	_output_new.quat_nominal.normalize();
 
-	_output_sample_delayed.quat_nominal *= q_delta;
+	_output_sample_delayed.quat_nominal = q_delta * _output_sample_delayed.quat_nominal;
 	_output_sample_delayed.quat_nominal.normalize();
 }
 
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -509,7 +509,7 @@ bool Ekf::realignYawGPS()
 			_state_reset_status.quat_change = quat_before_reset.inversed() * _state.quat_nominal;
 
 			// add the reset amount to the output observer buffered data
-			// Note q1 *= q2 is equivalent to q1 = q2 * q1
+			// Note q1 *= q2 is equivalent to q1 = q1 * q2
 			for (uint8_t i = 0; i < _output_buffer.get_length(); i++) {
 				_output_buffer[i].quat_nominal *= _state_reset_status.quat_change;
 			}
@@ -750,7 +750,7 @@ bool Ekf::resetMagHeading(Vector3f &mag_init, bool increase_yaw_var, bool update
 	if (update_buffer) {
 		// add the reset amount to the output observer buffered data
 		for (uint8_t i = 0; i < _output_buffer.get_length(); i++) {
-			// Note q1 *= q2 is equivalent to q1 = q2 * q1
+			// Note q1 *= q2 is equivalent to q1 = q1 * q2
 			_output_buffer[i].quat_nominal *= _state_reset_status.quat_change;
 		}
 
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -506,17 +506,16 @@ bool Ekf::realignYawGPS()
 			_flt_mag_align_start_time = _imu_sample_delayed.time_us;
 
 			// calculate the amount that the quaternion has changed by
-			_state_reset_status.quat_change = quat_before_reset.inversed() * _state.quat_nominal;
+			_state_reset_status.quat_change = _state.quat_nominal * quat_before_reset.inversed();
 
 			// add the reset amount to the output observer buffered data
-			// Note q1 *= q2 is equivalent to q1 = q1 * q2
 			for (uint8_t i = 0; i < _output_buffer.get_length(); i++) {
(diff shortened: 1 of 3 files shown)


priseborough:
I will go through the signs. The code is supposed to use the Eigen library convention.

Also if https://github.com/PX4/ecl/blob/f6eaa92dddf0e4f932227241339d62e5b0cdbf56/EKF/ekf_helper.cpp#L509 is giving opposite sign to expected, then it would explain why reversed order is working.

priseborough:
More testing required.

priseborough:
I will attempt to test this tomorrow in SITL by adding a height varying magnetic declination angle  offset here: https://github.com/PX4/sitl_gazebo/blob/master/src/gazebo_magnetometer_plugin.cpp#L173

EKF/ekf_helper.cpp
```diff
@@ -509,14 +509,14 @@ bool Ekf::realignYawGPS()
 			_state_reset_status.quat_change = quat_before_reset.inversed() * _state.quat_nominal;
 
 			// add the reset amount to the output observer buffered data
-			// Note q1 *= q2 is equivalent to q1 = q2 * q1
+			// Note q1 *= q2 is equivalent to q1 = q1 * q2
```
priseborough: This correct but confusing given the use of the Eigen library. I propose we remove this comment line and change line 514 to:

_output_buffer[i].quat_nominal = _state_reset_status.quat_change * _output_buffer[i].quat_nominal;

to make the multiplication order more obvious.

EKF/ekf_helper.cpp
```diff
@@ -509,14 +509,14 @@ bool Ekf::realignYawGPS()
...
 			for (uint8_t i = 0; i < _output_buffer.get_length(); i++) {
 				_output_buffer[i].quat_nominal *= _state_reset_status.quat_change;
 			}
 
 			// apply the change in attitude quaternion to our newest quaternion estimate
 			// which was already taken out from the output buffer
-			_output_new.quat_nominal = _state_reset_status.quat_change * _output_new.quat_nominal;
+			_output_new.quat_nominal = _output_new.quat_nominal * _state_reset_status.quat_change;
```
priseborough: This is the wrong way round - we use the Eigen library convention.

EKF/ekf_helper.cpp
```diff
@@ -750,13 +750,13 @@ bool Ekf::resetMagHeading(Vector3f &mag_init, bool increase_yaw_var, bool update
 	if (update_buffer) {
 		// add the reset amount to the output observer buffered data
 		for (uint8_t i = 0; i < _output_buffer.get_length(); i++) {
-			// Note q1 *= q2 is equivalent to q1 = q2 * q1
+			// Note q1 *= q2 is equivalent to q1 = q1 * q2
```
priseborough: Suggest we remove this line and change line 754 to:

_output_buffer[i].quat_nominal = _state_reset_status.quat_change * _output_buffer[i].quat_nominal;

to make the use of the eigen library order convention more obvious.

EKF/ekf_helper.cpp
```diff
@@ -750,13 +750,13 @@ bool Ekf::resetMagHeading(Vector3f &mag_init, bool increase_yaw_var, bool update
...
 			_output_buffer[i].quat_nominal *= _state_reset_status.quat_change;
 		}
 
 		// apply the change in attitude quaternion to our newest quaternion estimate
 		// which was already taken out from the output buffer
-		_output_new.quat_nominal = _state_reset_status.quat_change * _output_new.quat_nominal;
+		_output_new.quat_nominal = _output_new.quat_nominal * _state_reset_status.quat_change;
```
priseborough: This is incorrect - it needs to follow the Eigen library convention.

priseborough:
I think the code is correct.

Testing in SITL with a large positive ground level declination offset gives the following results with 

master:

master

This PR (note that the reversed order in this PR only applies to a single output frame and does not show up in the log):

pr605

Replacing the q1 *= q2 with q1 = q2 * q1:

*= replaced

CarlOlsson:
@priseborough Sorry to reopen this PR, but I am still not convinced that the code is correct on master.

In order to test the quaternion multiplication I did the following:
1) changed the magnetic declination lookup to always return 45 degrees in SITL
2) disable the saving of magnetic declination in order to have reproducible tests
3) set EKF2_REQ_GPS_H to 10s in order to log the heading reset when GPS lock is achieved

I then printed the quaternions in the output buffer before and after the reset, as well as the output_new nominal quaternion before and after the reset.

Running `make px4_sitl gazebo HEADLESS=1` on master gives:

Buffer before reset:
`_output_buffer[i].quat_nominal = 0.984, -0.012, -0.019, 0.178`
Buffer after reset:
`_output_buffer[i].quat_nominal = 0.842, -0.004, -0.022, 0.539`
Q nominal before reset:
`_output_new.quat_nominal = 0.984, -0.012, -0.019, 0.178`
Q nominal after reset:
`_output_new.quat_nominal = 0.842, 0.006, -0.035, 0.539`

As can be seen the output_new quatertion does not match the quaterion in the output buffer.
If I do the same test after changing the quaterion multiplication order, or simply replacing 
```
_output_new.quat_nominal = _state_reset_status.quat_change * _output_new.quat_nominal;
```
with
```
_output_new.quat_nominal *= _state_reset_status.quat_change;
```

I get:
Buffer before reset:
`_output_buffer[i].quat_nominal = 0.984, -0.012, -0.019, 0.178`
 Buffer after reset:
`_output_buffer[i].quat_nominal = 0.842, -0.004, -0.023, 0.539`
Q nominal before reset:
`_output_new.quat_nominal = 0.984, -0.012, -0.019, 0.178`
Q nominal after reset:
`_output_new.quat_nominal = 0.842, -0.004, -0.023, 0.539`

As can be seen now the two quaternions perfectly match

EKF/ekf_helper.cpp
priseborough: Have checked the library and the self multiplication operator does reverse the order compared to the eigen convention which is not good. I recommend we discontinue use of the self multiplication operator to avoid confusion and replace it with_output_buffer[i].quat_nominal = _output_buffer[i].quat_nominal * _state_reset_status.quat_change;

priseborough: It might be best to change line 509 so that the delta quaternion it produces is from old -> new and can be applied using an order that is consistent with use of the quaternion product in other places when an incremental rotation is applied, eg:

_state_reset_status.quat_change =  * _state.quat_nominal * quat_before_reset.inversed();

EKF/ekf_helper.cpp
priseborough: The sign of _state_reset_status.quat_change was reversed.

EKF/ekf_helper.cpp
priseborough: Suggest we remove this comment line, fix the sign in calculation of _state_reset_status.quat_change and discontinue use of the *= operator

priseborough:
@CarlOlsson Thanks for looking at that. The reversed order in the self product operator was confusing me. I have pushed patches which discontinues use of the operator and makes the calculation and application of a delta quaternion consistent throughout the code.

priseborough:
SITL test log here: https://logs.px4.io/plot_app?log=feaca793-0fdd-4c14-bc63-e14e1a53f0e5

priseborough:
Yaw reset behaviour appears normal:

Screen Shot 2019-06-13 at 7 43 00 am

Because the bug affected a single frame of the output and also after the output was published, its effect was not observable in the logs, however provided we apply the delta quaternion in the code using a consistent method, eg q_new = q_delta * q_old, then correct usage can be determined.

priseborough:
A second SITL test was performed https://logs.px4.io/plot_app?log=99ade2e6-d51a-465c-add8-57fef0b48347 with the takeoff in POSCTL mode to check that the delta rotation published by the vehicle_attitude message was correct and that the attitude controller uses it correctly to reset the yaw set point and prevent a large yaw change.

Screen Shot 2019-06-14 at 11 29 26 am

This functionality appears to be operating correctly

jkflying:
Could we remove the self-multiplication from the library entirely? As in edit the library to remove those functions. Otherwise I think this issue will keep showing up.

priseborough:
I support that notion, but it will need to be dealt with in another PR.

PR #606, opened 2019-05-23
fixed calculation of magentic table values by using floor instead of int casting

sverling:
When calculating the min_lat it was incorrectly casting to int which for negativ values will round up rather than round down. This will then result in a lat_scale/lon_scale that is forced to zero as lat/lon is a smaller value than min_lat/min_lon.

### DIFF
--- a/geo_lookup/geo_mag_declination.cpp
+++ b/geo_lookup/geo_mag_declination.cpp
@@ -49,6 +49,7 @@
 #include <stdint.h>
 
 using math::constrain;
+using math::floorf;
 
 /** set this always to the sampling in degrees for the table below */
 static constexpr float SAMPLING_RES = 10.0f;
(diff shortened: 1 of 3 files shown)


LorenzMeier:
Thanks!

dagar:
@sverling @MaEtUgR actually I would drop this entirely and use floorf from the c math library. Otherwise you'll need to replicate the change PX4 side.

sverling:
Sounds goot to me. Main reason i did it this way was i wasn't sure if i can just add another library to it. On my branch (not latest PX4 master, so this might not hold true anymore) it would compile correctly on PX4 side as there it is using the correct math libraries and I only had to add the function to compile for standalone ecl compilation

dagar:
@sverling likely just missing the math.h include ecl side. https://github.com/PX4/ecl/pull/608

PR #614, opened 2019-06-10
EKF: Fix bug allowing state initialisation to a bad GPS height

priseborough:
If GPS is being used as the primary height source (EKF2_HGT_MODE = 1) then the first 3D fix from the GPS is used to set the height origin. This can cause problems because these erarly estimates are inaccurate and can result in the EKF origin being out by 30m or more. The GPS height solution then improves and the large rapid change in measurement causes it to be rejected by the EKF until the height rejection timeout, resulting in a reset. The following figure shows the large innovations and the reset caused by this behaviour.

before

With this fix applied, replay of the same data shows the problem to be fixed.

after

This fix requires that the same GPS quality check used to determine if GPs can be used for navigation are also required before it can be used as the primary height sensor. The _gps_hgt_faulty class variable has been renamed _gps_hgt_intermittent to be consistent with its usage and unnecessary resets of this variable have been removed.

Flight testing with EKF2_HGT_MODE = 1 is required

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -80,12 +80,12 @@ void Ekf::controlFusionModes()
 
 	}
 
-	// check faultiness (before pop_first_older_than) to see if we can change back to original height sensor
+	// check for intermittent data (before pop_first_older_than)
 	const baroSample &baro_init = _baro_buffer.get_newest();
 	_baro_hgt_faulty = !((_time_last_imu - baro_init.time_us) < 2 * BARO_MAX_INTERVAL);
 
 	const gpsSample &gps_init = _gps_buffer.get_newest();
-	_gps_hgt_faulty = !((_time_last_imu - gps_init.time_us) < 2 * GPS_MAX_INTERVAL);
+	_gps_hgt_intermittent = !((_time_last_imu - gps_init.time_us) < 2 * GPS_MAX_INTERVAL);
 
 	// check for arrival of new sensor data at the fusion time horizon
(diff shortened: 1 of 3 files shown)


priseborough:
Flight test with EKF2_HGT_MODE = 0 (use Baro)

Screen Shot 2019-06-10 at 3 37 05 pm

Same hardware and location tested with EKF2_HGT_MODE = 1 (use GPS). EKF switches from baro to GPS when GPS preflight checks pass:

Screen Shot 2019-06-10 at 3 39 33 pm

jkflying:
Logically it makes sense - we shouldn't be using it for primary height if it isn't good enough for navigation. My only criticism is that I dislike solving continuous systems by adding discontinuities. Would it be possible to solve this by using an extremely large variance estimate on the initial GPS height? The EKF would then have a large variance on the height, which wouldn't have as much trouble adjusting to a rapidly changing height as the variance decreases.

priseborough:
We should also update the parameter description for EKF2_HGT_MODE to recommend turning on bit position index 3 in EKF2_GPS_CHECK if EKF2_HGT_MODE = 1 so that the vertical position accuracy metric check is enabled.

priseborough:
@jkflying Unfortunately it is the setting of the height datum that causes most of the the problems. Also the GPS solution can drift in the first 30 seconds at a rate that makes the EKF unable to distinguish between the GPS change in height offset and the IMU error.

Issue #602, opened 2019-05-10
Suboptimal wind estimation initialization


CarlOlsson:
[Here](https://logs.px4.io/plot_app?log=264d2d6d-026c-4ee0-a277-cd826c993ec6) is a sitl mission with gazebo plane where the vehicle takes of facing east and does a loiter circle
![bokeh_plot](https://user-images.githubusercontent.com/7795133/57535465-94067e00-7342-11e9-902f-b4e142144e1d.png)
Synthetic sideslip fusion was disabled in the test

One would obviously expect the wind uncertainty in the east direction to be initialized with a lower uncertainty than in north direction. However this is not the case
![image](https://user-images.githubusercontent.com/7795133/57536948-90282b00-7345-11e9-9fbd-bdb9961df72a.png)

Here is another example of airspeed fusion commencing after transition to cruise on a vtol platform 
![image](https://user-images.githubusercontent.com/7795133/57537047-c49be700-7345-11e9-9fec-60c790364cb0.png)

As can be seen the uncertainty in east is in the beginning much higher than north.
I guess this has to do with the fact that the airspeed fusion measurement model does not incorporate the fact that the pitot tube measures the airspeed only in one direction, more info [here](https://github.com/PX4/ecl/issues/399).

We should add this as an alternative fusion method.

CarlOlsson:
An intermediate step would be to use the fact that we know that the pitot measures the airspeed in one direction for the covariance intialization in the same way as we use it in the wind state initialization
https://github.com/PX4/ecl/blob/a27a43eafa8f4dd514e89984f5394260a36ea4f6/EKF/airspeed_fusion.cpp#L267-L267

priseborough:
That is correct. The derivation assumes an omni directional sensor. I can derive an alternative observation model that uses a cosine law.

CarlOlsson:
okey awesome, I started with that some time ago but got stuck on fact that you have to take the partial derivate of the discontinues function.
If you know away around it that would be awesome!

priseborough:
The initialisation of wind states and covariances does take direction into account - see 

https://github.com/PX4/ecl/blob/master/EKF/covariance.cpp#L920-L921

and

https://github.com/PX4/ecl/blob/master/EKF/airspeed_fusion.cpp#L266

However, if the pitch angle is high, then the initialisation could be incorrect due to the use of yaw angle from a 321 rotation sequence.

priseborough:
[...]

PR #615, opened 2019-06-11
EKF: update wind covariance initialization

CarlOlsson:
Update the initialization of the wind state covariance submatrix as discussed [here](https://github.com/PX4/ecl/issues/602)

Here is a SITL mission where the plane take off facing east and then turns to the north. Synthetic sideslip fusion is turned off
master/pr
image

As can be seen on the PR the initial wind uncertainty in east direction is lower than in north which makes sense.

@priseborough Compared to our precious discussion I did not set the wind along body y to zero in the covariance initialization. When looking through the derivation I simply did not see a reason for that, would be happy if you could review

The derivation is done using the following sympy script
```
import sympy as sp
Wx, Wy, yaw, R_TAS, initial_wind_var_body_y, R_yaw = sp.symbols('Wx Wy yaw R_TAS initial_wind_var_body_y R_yaw')
Wn = Wx * sp.cos(yaw) - Wy * sp.sin(yaw)
We = Wx * sp.sin(yaw) + Wy * sp.cos(yaw)

Wn_Wx = sp.diff(Wn, Wx)
Wn_Wy = sp.diff(Wn, Wy)
Wn_yaw = sp.diff(Wn, yaw)
We_Wx = sp.diff(We, Wx)
We_Wy = sp.diff(We, Wy)
We_yaw = sp.diff(We, yaw)

G = sp.Matrix([[Wn_Wx, Wn_Wy, Wn_yaw],[We_Wx, We_Wy, We_yaw]])
wind_body_cov = sp.Matrix([[R_TAS, 0.0, 0.0], [0.0,initial_wind_var_body_y, 0.0], [0.0, 0.0, R_yaw]])
wind_cov = G * wind_body_cov * G.T

print('P[22][22] = ' + str(wind_cov[0,0]))
print('P[22][23] = ' + str(wind_cov[0,1]))
print('P[23][22] = ' + str(wind_cov[1,0]))
print('P[23][23] = ' + str(wind_cov[1,1]))
```

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -894,29 +894,26 @@ void Ekf::zeroMagCov()
 void Ekf::resetWindCovariance()
 {
 	if (_tas_data_ready && (_imu_sample_delayed.time_us - _airspeed_sample_delayed.time_us < (uint64_t)5e5)) {
-		// Use airspeed and zer sideslip assumption to set initial covariance values for wind states
+		// Derived using EKF/matlab/scripts/Inertial Nav EKF/wind_cov.py
+		// TODO: explicitly include the sideslip angle in the derivation
+		Eulerf euler321(_state.quat_nominal);
+		const float euler_yaw = euler321(2);
+		const float R_TAS = sq(math::constrain(_params.eas_noise, 0.5f, 5.0f) * math::constrain(_airspeed_sample_delayed.eas2tas, 0.9f, 10.0f));
+		const float initial_sideslip_uncertainty = math::radians(15.0f);
+		const float initial_wind_var_body_y = sq(_airspeed_sample_delayed.true_airspeed * sinf(initial_sideslip_uncertainty));
+		const float R_yaw = sq(math::radians(10.0f));
 
-		// calculate the wind speed and bearing
-		float spd = sqrtf(sq(_state.wind_vel(0)) + sq(_state.wind_vel(1)));
-		float yaw = atan2f(_state.wind_vel(1), _state.wind_vel(0));
+		// rotate wind velocity into earth frame aligned with vehicle yaw
+		const float Wx = _state.wind_vel(0) * cosf(euler_yaw) + _state.wind_vel(1) * sinf(euler_yaw);
+		const float Wy = -_state.wind_vel(0) * sinf(euler_yaw) + _state.wind_vel(1) * cosf(euler_yaw);
 
-		// calculate the uncertainty in wind speed and direction using the uncertainty in airspeed and sideslip angle
-		// used to calculate the initial wind speed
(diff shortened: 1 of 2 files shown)


EKF/covariance.cpp
```diff
@@ -923,28 +923,19 @@ void Ekf::resetWindCovariance()
...
-		float sin_cos_yaw = sin_yaw * cos_yaw;
-		float spd_2 = sq(spd);
-		P[22][22] = R_yaw * spd_2 * sin_yaw_2 + R_spd * cos_yaw_2;
-		P[22][23] = - R_yaw * sin_cos_yaw * spd_2 + R_spd * sin_cos_yaw;
-		P[23][22] = P[22][23];
-		P[23][23] = R_yaw * spd_2 * cos_yaw_2 + R_spd * sin_yaw_2;
+		Eulerf euler321(_state.quat_nominal);
+		const float euler_yaw = euler321(2);
+		const float R_TAS = sq(math::constrain(_params.eas_noise, 0.5f, 5.0f) * math::constrain(_airspeed_sample_delayed.eas2tas, 0.9f, 10.0f));
+		const float intitial_sideslip_uncertainty = math::radians(15.0f);
```
priseborough: check variable naming. I think you intended 'initial_sideslip_uncertainty'

priseborough:
I've gone through the equations and have done some testing using different inout values including varying the hardcoded values for  initial_sideslip_uncertainty and R_yaw.

Something I noticed when doing this is  that when the wind estimate is is zero, the yaw angle uncertainty has no effect on the resultant P matrix entries. For example if:

euler_yaw = 90 deg flying east
wind_vel_0 = 0
wind_vel_1 = 0
true_airspeed = 15 m/s
R_TAS = 4.0
initial_sideslip_uncertainty = 0
R_yaw = (deg2rad * 10.0)^2;

The P(22,23) sub matrix produced is:

P =

    0.0000    0.0000
    0.0000    4.0000

In this instance I would have expected when flying east with a yaw uncertainty and zero diselip uncertainty that there would be some uncertainty in the North velocity component due to the yaw uncertainty.

Should we the equations for Wn and We explicitly include  the sideslip angle? eg:

Wn = Wx * sp.cos(yaw+beta) - Wy * sp.sin(yaw+beta)
We = Wx * sp.sin(yaw+beta) + Wy * sp.cos(yaw+beta)

CarlOlsson:
@priseborough As we already discussed I agree with your points above, unfortunately I don't have time to implement and test this at the moment.
However, we have done a couple of thousand flights with the current logic and it seems to work well. So how do you feel about adding a todo comment regarding the sideslip angle in the code and merging this as it is?

CarlOlsson:
okey will do the changes

CarlOlsson:
@priseborough I finally got the time to push the derivation script. For now I put it in the matlab folder even though it is written in python. I guess we will have to clean that up at some point

LorenzMeier:
@CarlOlsson This would need to be rebased. Thanks!

CarlOlsson:
Done

PR #616, opened 2019-06-11
tecs: update load factor and STE_to_throttle

CarlOlsson:
@priseborough I was cleaning up some branches and realized I never pushed these two things which we discussed a while back

### DIFF
--- a/tecs/tecs.cpp
+++ b/tecs/tecs.cpp
@@ -319,11 +319,20 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 		_throttle_setpoint = 1.0f;
 
 	} else {
-		// Adjust the demanded total energy rate to compensate for induced drag rise in turns.
-		// Assume induced drag scales linearly with normal load factor.
-		// The additional normal load factor is given by (1/cos(bank angle) - 1)
+		/*
+		Adjust the demanded total energy rate to compensate for induced drag rise in turns.
+		Assume induced drag scales with lift coefficient/normal load factor squared.
+		Normal load factor is given by cos(pitch) / (cos(roll)*(1 + sin^2(pitch))
+		Pitch is angle of X axis above horizontal plane (second rotation angle from Tait-Bryan 321 sequence)
+		Roll is angle of Y axis below horizontal plane (second rotation angle from Tait-Bryan 312 sequence)
+		*/
+		float roll = asinf(rotMat(2, 1));
+		float cosRoll = constrain(cosf(roll), 0.1f, 1.0f);
+		float sinePitch = constrain(-rotMat(2, 0), -1.0f, 1.0f);
+		float pitch = asinf(sinePitch);
+		float load_factor = cosf(pitch) / (cosf(roll)*(1.0f + sinePitch * sinePitch));
 		float cosPhi = sqrtf((rotMat(0, 1) * rotMat(0, 1)) + (rotMat(1, 1) * rotMat(1, 1)));
-		STE_rate_setpoint = STE_rate_setpoint + _load_factor_correction * (1.0f / constrain(cosPhi, 0.1f, 1.0f) - 1.0f);
+		STE_rate_setpoint = STE_rate_setpoint + _load_factor_correction * (load_factor * load_factor - 1.0f);
 
 		// Calculate a predicted throttle from the demanded rate of change of energy, using the cruise throttle
 		// as the starting point. Assume:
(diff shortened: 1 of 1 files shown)


tecs/tecs.cpp
```diff
@@ -323,7 +323,7 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 		// Assume induced drag scales linearly with normal load factor.
 		// The additional normal load factor is given by (1/cos(bank angle) - 1)
 		float cosPhi = sqrtf((rotMat(0, 1) * rotMat(0, 1)) + (rotMat(1, 1) * rotMat(1, 1)));
-		STE_rate_setpoint = STE_rate_setpoint + _load_factor_correction * (1.0f / constrain(cosPhi, 0.1f, 1.0f) - 1.0f);
+		STE_rate_setpoint = STE_rate_setpoint + _load_factor_correction * (1.0f / constrain(cosPhi * cosPhi, 0.1f, 1.0f) - 1.0f);
```
jkflying: Does it make sense to take the sqrt above if we are just going to be squaring it here anyway?

RomanBapst: @CarlOlsson Can you also please verify the function for calculating the cosine of roll angle?
I plugged in the values into the formula and I end up with:
`cos(roll)^2 + sin(roll)^2 * sin(pitch)^2`
This is only exact if the pitch angle is zero which is probably not generally a bad assumption.
However, I prefer if we use the matrix lib to calculate the roll angle and then just take the cosine of that. Otherwise we lose the benefit of the test coverage.

Could you also please update the comment above the formula you changed, I don't think it reflects the calculations anymore, does it?

RomanBapst: @CarlOlsson I assume the rationale behind this is the quadratic shape of Cd as a function of Cl?

priseborough: There are two considerations here:

1) Induced drag scales with CL^2 and is more logically scaled using  load factor^2
2) Roll angle from a Tait-Bryan 321 sequence (what the library returns) is not useful for this type of compensaton because it gives incorrect results for normal load factor at large pitch angles. I think roll from a Tait-Bryan 312 (Yaw->Roll->Pitch) sequence would be more useful.

I will have a look at this and get back to you tomorrow.

priseborough: The roll from a Tait-Bryan 312 sequnece is = asinf(_R_to_earth(2, 1)) 

See the Ekf::fuseHeading() function in the EKF for an example of where this rotation definition is used.

It is possible the exisiting code in TECS is trying to do a pitch compensation, but I think the Roll from a 312 sequence would be more understandable.

priseborough: Whatever method is used should be tested to correctly calculate the load factor for climbing and descending turns.

RomanBapst: @priseborough makes sense

priseborough: See attached for derivation of pitch angle compensation where ignoring angle of attack, pitch angle is defined as angle of X axis above horizontal.

priseborough: Therefore the load factor calculated  both centripetal, forwards and vertical acceleration components would be:

cos(pitch) / (cos(roll)*(1 + sin^2(pitch))

where pitch is angle of X axis above horizontal plane (second rotation angle from Tait-Bryan 321 sequence) and roll angle is angle of Y axis relative to horizontal plane (second rotation angle from Tait-Bryan 312 sequence)

tecs/tecs.cpp
```diff
@@ -323,7 +323,7 @@ void TECS::_update_throttle_setpoint(const float throttle_cruise, const matrix::
 		// Assume induced drag scales linearly with normal load factor.
```
priseborough: This comment needs to be amended, eg:

		// Assume induced drag scales with lift coefficient/normal load factor squared.

priseborough:
@CarlOlsson I've pushed some changes aligned with the discussion above. You can Skype me to discuss.

CarlOlsson:
Thanks for the update Paul! We will do some testing in the coming weeks and report back

priseborough:
@CarlOlsson Any progress on this?

CarlOlsson:
sorry I have not had the time to test it properly. Will try to do it as soon as possible

CarlOlsson:
@priseborough Can you please clarify the last step in your derivation above (from the picture).
1 / (cos(theta) + sin(theta) * sin(theta) / cos(theta)) = cos(theta)

CarlOlsson:
@priseborough Ping

CarlOlsson:
@priseborough Ping

priseborough:
@CarlOlsson I've fixed the error in the last line of the derivation for the pitch angle compensation. I will push a commit to fix the corresponding line in code.

priseborough:
@RomanBapst are you interested in evaluating this?

PR #620, opened 2019-06-29
control.cpp: clean up extra code judge

xdwgood:
```
resetVelocity();
resetPosition();
```
always return `ture,` so we don't need judge more, just direct `resetVelocity();resetPosition();`

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -590,10 +590,8 @@ void Ekf::controlGpsFusion()
 
 					}
 
-					if (_control_status.flags.gps) {
-						ECL_INFO("EKF commencing GPS fusion");
-						_time_last_gps = _time_last_imu;
-					}
+					ECL_INFO("EKF commencing GPS fusion");
+					_time_last_gps = _time_last_imu;
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


dagar:
Logically your change looks correct, but was something else intended here?

xdwgood:
I don't know, it may be that the previous two functions will return false in some cases, Paul should know more

priseborough:
Can please you explain the problem change fixes and show evidence of how you have tested it? Thanks.

xdwgood:
only code logic extra....Always return true Why do you need an if judgment?

xdwgood:
I found out when reading the code, not from flight test.

xdwgood:
@dagar As you said, the logical change of code is just a clean up of the code.But the problem raised by paul makes me feel confused...As long as the person who has read the code, it is easy to know that this is just a cleanup, unless he does not see the code at all. Is the clean up of ecl code not welcome?But it doesn't matter, thank you and Paul for your comments.

priseborough:
Are you saying that it is a non functional change (dies not change behaviour) ? Have you done any testing that exercises the changed lines of code?

xdwgood:
If this change is acceptable I can do a flight test using the H520 :smiley:

PR #622, opened 2019-07-17
Fixed airspeed scaling

RomanBapst:
Small enhancement to the wind estimation library.
1) Added API which allows to disable true airspeed scale factor learning but instead use a pre-set value.
Airspeed scale factor errors are affected by the particular installation of the pitot tube. Once the scale factor has been learned it should not change anymore.

2) Added method which notifies if a reset of the filter has taken place in the current cycle.

### DIFF
--- a/airdata/WindEstimator.hpp
+++ b/airdata/WindEstimator.hpp
@@ -81,6 +81,7 @@ public:
 		wind_var[0] = _P(0, 0);
 		wind_var[1] = _P(1, 1);
 	}
+	bool get_wind_estimator_reset() { return _wind_estimator_reset; }
 
 	void set_wind_p_noise(float wind_sigma) { _wind_p_var = wind_sigma * wind_sigma; }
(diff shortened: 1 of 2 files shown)


RomanBapst:
@dagar Any idea why this fails? Seems like it fails to checkout the submodule:
image

airdata/WindEstimator.hpp
```diff
@@ -121,6 +126,9 @@ class WindEstimator
 	uint64_t _time_rejected_tas =
 		0;		///<timestamp of when true airspeed measurements have consistently started to be rejected
 
+	bool _wind_estimator_reset = false; ///< wind estimator was reset in this cycle
+	float _enforced_airspeed_scale{-1}; ///< by default we want to estimate the true airspeed scale factor (see enforce_airspeed_scale(...) )
```
xdwgood: -1.0f

RomanBapst: fixed in the follow up PR

PR #623, opened 2019-07-18
Fixed airspeed scaling

RomanBapst:
Replaces #622


RomanBapst:
@sfuhrer Could you please review this?

airdata/WindEstimator.hpp
```diff
@@ -121,6 +126,9 @@ class WindEstimator
 	uint64_t _time_rejected_tas =
 		0;		///<timestamp of when true airspeed measurements have consistently started to be rejected
 
+	bool _wind_estimator_reset = false; ///< wind estimator was reset in this cycle
+	float _enforced_airspeed_scale{-1.0f}; ///< by default we want to estimate the true airspeed scale factor (see enforce_airspeed_scale(...) )
```
sfuhrer: not 100% sure if by default we should the scale estimation should be on or not. When it's on the airspeed validity checks are more or less turned off (what we though maybe want per default anyway)

Issue #627, opened 2019-08-06
EKF heading drifts when EKF2_MAG_TYPE = 4


priseborough:
The EKF2_MAG_TYPE = 5 parameter setting is supposed to enable operation without magnetometer in situations where there are high levels of magnetic interference and no observations in an earth frame of reference that require an absolute heading such as GPS. Examples where this option may be used is indoor using optical flow.

Testing with EKF2_AID_MASK = 0 (to disable GPs use) and EKF2_MAG_TYPE = 4 has showed that the heading drifts on the ground because when the vehicle is stationary, the last non-moving heading is not being used as a yaw observation as was intended in the original design. Also there are no yaw innovations at any time resulting in a steady growth in quaternion variance that may eventually destabilise the covariance matrix.

The following log can be replayed to illustrate this effect:

https://logs.px4.io/plot_app?log=ac2c156b-14f2-46d3-87d8-3100353e1823

The test procedure involved powering the vehicle when resting on the ground, picking it up and performing a 360 degree yaw rotation before placing it back down again before being picked up and rotated 360deg in the opposite direction, placed back down and de-powered.

The following log shows the result of this data relayed using current master:

https://logs.px4.io/plot_app?log=d1023af3-5b3c-43e2-a7e6-c3e7362af495

The parameter change forced during replay was SDLOG_MODE = 3

AndreasAntener:
Paul, could the same happen with EKF2_MAG_TYPE 5? We sometimes see a drift as well (right from boot) but using aid mask 129. I'm trying to reproduce and provide a log.

AndreasAntener:
Local test with mag type 5, interestingly this now shows an oscillation. Not sure what the data drop is about. Also it calms down quickly after GPS heading is available this time. It's quite hard to reproduce exactly as it happened on the field, but I hope this at least shows the initial cause (still trying to get the exact effect we've had).

https://review.px4.io/plot_app?log=7eb93866-a59d-429b-80e3-e0cde7a3e03d

![Screenshot 2019-08-08 at 15 07 38](https://user-images.githubusercontent.com/5750020/62705730-4a373d80-b9ee-11e9-83ae-62cef381d12b.png)

AndreasAntener:
Oh, what I forgot to mention is that the drift (or oscillation) is matched by the gyro bias. It doesn't follow the actual gyro drift. Not sure if that is part of the cause or just another symptom. At least, from what I can see is, that is has nothing to do with gyro drift.

PR #628, opened 2019-08-06
EKF: Fix on ground yaw drift when using EKF2_MAG_TYPE = 4

priseborough:
Fixes https://github.com/PX4/ecl/issues/627

Tested using replay log:

Before change log: https://logs.px4.io/plot_app?log=d1023af3-5b3c-43e2-a7e6-c3e7362af495

The yaw angle continued to drift when on ground
euler angles - before

Quaternion variance continues to grow - if left unchecked this will eventually destabilise the filter
variances - before

After change log: https://logs.px4.io/plot_app?log=d4591d33-40c6-4fd2-bd99-a892f2a308d6

Yaw doesn't drift
euler angles - after

Quaternion variance when resting on ground doesn't continue to grow:
variances - after

Short flight test of changes here: https://logs.px4.io/plot_app?log=eb89707c-d3b9-4175-8806-8e01d5d6f9ae

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1521,7 +1521,7 @@ void Ekf::controlMagFusion()
 				save_mag_cov_data();
 			}
 
-		} else if (_params.mag_fusion_type == MAG_FUSE_TYPE_HEADING) {
+		} else if (_params.mag_fusion_type == MAG_FUSE_TYPE_HEADING || _params.mag_fusion_type == MAG_FUSE_TYPE_INDOOR) {
 			// always use heading fusion
 			_control_status.flags.mag_hdg = true;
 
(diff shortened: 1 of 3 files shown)


priseborough:
@RomanBapst Ping

RomanBapst:
@bkueng Isn't this the problem we have when using gps heading and the vehicle is waiting on the ground for an RTK lock?

mhkabir:
> Isn't this the problem we have when using gps heading and the vehicle is waiting on the ground for an RTK lock?

bkueng:
> @bkueng Isn't this the problem we have when using gps heading and the vehicle is waiting on the ground for an RTK lock?

Yes looks exactly like that.

Issue #625, opened 2019-07-28
GPS heading fusion does not block mag heading fusion


AndreasAntener:
From the parameter documentation I understand that with GPS heading enabled, and mag fusion set to automatic, the mag yaw should only be used until the GPS heading is available. But for that I think there is a condition missing similar to the one already in place for EV heading fusion in https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L1377, otherwise the mag fusion is re-enabled right after GPS heading fusion disables it (and they will both run).

I'd appreciate if someone can clarify that this is the general idea behind it. Also, heading fusion never falls back to mag if GPS becomes unavailable?

AndreasAntener:
If we assume magnetometer values are not safe to use in certain areas we don't want it use both or fall back to mag.

What happens if GPS heading is not updated anymore for a long time? This case is not explicitly handled yet as far as I can see. It seems to just continue integrating the gyro.

AndreasAntener:
This is from a bench test, you can see the innovation spikes which have roughly the length of the difference between mag and GPS heading. With MAG_TYPE 0. I don't think that's the idea.

<img width="1375" alt="Screenshot 2019-07-31 at 09 28 59" src="https://user-images.githubusercontent.com/5750020/62192490-02714000-b376-11e9-9e8b-14ead86d4f61.png">

RomanBapst:
@AndreasAntener As far as I know gps heading was tested but I don't think the logic is complete (as you have correctly stated). Do you have a replay log available which I could use to test some changes?
I do have a setup in the office in case you don't have one available. During the replay I can stop gps data to make sure that it switches over between gps and mag correctly.

AndreasAntener:
No replay log yet. Trying to reproduce locally. Similar issue like Paul reported occurs as well: https://github.com/PX4/ecl/issues/627

I will push the small change I did, which makes it switch correctly as far as I can see, and it works well, but we many times we have a drift initially which won't come to rest.

AndreasAntener:
Issue reproduced on PX4 1.9.2: https://review.px4.io/plot_app?log=703082d0-c57f-40a8-bdd4-5194de2233f5

![Screenshot 2019-08-08 at 15 02 24](https://user-images.githubusercontent.com/5750020/62705390-9d5cc080-b9ed-11e9-9233-45d57bec38ca.png)

priseborough:
[...]

PR #629, opened 2019-08-06
EKF: Stop mag fusion when GPS heading fusion is active

AndreasAntener:
Review needed, initial suggestion for fixing https://github.com/PX4/ecl/issues/625

This changes removes the mag innovation spikes from the plot in the issue and it can look like this now:

Screenshot 2019-08-06 at 21 15 44

We see the heading drift in this plot though, so if we can get rid of that too it would be in an initial good working state.


priseborough:
This fixes the immediate bug. The heading drift is a more complex issue that I will address in a separate PR after getting some suitable log data for replay testing.

PR #634, opened 2019-08-20
Fix alignment of external local frame 

kamilritz:
These changes fix the alignment between the local origin frame and an external fixed frame of the external vision system. The changes contain:
- fix the computation of the orientation error between the frames
- enable access to the estimated alignment rotation
- removed mag suffix
- enable the computation of the orientation error only based on MASK_ROTATE_EV
- fixes wrong labeling of get_ekf2ev_quaternion function (also see: PR #635)

EKF_AID_MASK_63
@priseborough

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -256,7 +256,7 @@ public:
 	void get_ekf_soln_status(uint16_t *status);
 
 	// return the quaternion defining the rotation from the EKF to the External Vision reference frame
-	void get_ekf2ev_quaternion(float *quat);
+	void get_ev2ekf_quaternion(float *quat);
 
 	// use the latest IMU data at the current time horizon.
 	Quatf calculate_quaternion() const;
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -255,7 +255,7 @@ public:
 	// return a bitmask integer that describes which state estimates can be used for flight control
 	void get_ekf_soln_status(uint16_t *status);
 
-	// return the quaternion defining the rotation from the EKF to the External Vision reference frame
+	// return the quaternion defining the rotation from the External Vision to the EKF reference frame
 	void get_ev2ekf_quaternion(float *quat);
 
 	// use the latest IMU data at the current time horizon.
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -1593,10 +1593,9 @@ void Ekf::setControlEVHeight()
 // and calculate a rotation matrix which rotates EV measurements into the EKF's navigation frame
 void Ekf::calcExtVisRotMat()
 {
-	// calculate the quaternion delta between the EV and EKF reference frames at the EKF fusion time horizon
+	// Calculate the quaternion delta that rotates from the EV to the EKF reference frame at the EKF fusion time horizon.
 	Quatf q_error = _state.quat_nominal * _ev_sample_delayed.quat.inversed();
 	q_error.normalize();
-	q_error.canonicalize();
 
 	// convert to a delta angle and apply a spike and low pass filter
 	Vector3f rot_vec = q_error.to_axis_angle();
@@ -1623,9 +1622,7 @@ void Ekf::calcExtVisRotMat()
 
 	// convert filtered vector to a quaternion and then to a rotation matrix
 	q_error.from_axis_angle(_ev_rot_vec_filt);
-	q_error.normalize();
-	q_error.canonicalize();
-	_ev_rot_mat = Dcmf(q_error); // rotation from EV reference to EKF reference
+	_ev_rot_mat = quat_to_invrotmat(q_error); // rotation from EV reference to EKF reference
 
 }
 
@@ -1633,7 +1630,7 @@ void Ekf::calcExtVisRotMat()
 // and update the rotation matrix which rotates EV measurements into the EKF's navigation frame
 void Ekf::resetExtVisRotMat()
 {
-	// calculate the quaternion delta between the EV and EKF reference frames at the EKF fusion time horizon
+	// Calculate the quaternion delta that rotates from the EV to the EKF reference frame at the EKF fusion time horizon.
 	Quatf q_error = _state.quat_nominal * _ev_sample_delayed.quat.inversed();
 	q_error.normalize();
 
@@ -1650,7 +1647,7 @@ void Ekf::resetExtVisRotMat()
 	}
 
 	// reset the rotation matrix
-	_ev_rot_mat = Dcmf(q_error); // rotation from EV reference to EKF reference
+	_ev_rot_mat = quat_to_invrotmat(q_error); // rotation from EV reference to EKF reference
 }
 
 // return the quaternions for the rotation from External Vision system reference frame to the EKF reference frame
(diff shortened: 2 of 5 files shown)


jkflying:
@dagar we're getting SWIG failures here since the `matrix` hash in `ecl` CI wasn't updated since Feb, is `ecl` CI maybe behind on some other related updates or something?

dagar:
> @dagar we're getting SWIG failures here since the `matrix` hash in `ecl` CI wasn't updated since Feb, is `ecl` CI maybe behind on some other related updates or something?

Let's make a quick attempt to fix it, but otherwise those tests can be temporarily disabled if they get in the way. Those tests don't provide much value at the moment.

priseborough:
Can you please split this PR into two parts. The first to fix any internal inconsistencies in handling of rotations and the second for improvements to use of VIO. I have a problem with rotating VIO data using an assumption of a continued common origin with the EKF.

kamilritz:
> Can you please split this PR into two parts. The first to fix any internal inconsistencies in handling of rotations and the second for improvements to use of VIO. I have a problem with rotating VIO data using an assumption of a continued common origin with the EKF.

This PR contains only the fix for calculation of the alignment quaternion, together with an interface such that the aligned visual odometry can be looked at in the log file. The fix is separated from the rest in the first commit: e6ccc6c442cee2d1c5d2cb5f83793c5de37aefb3

EKF/ekf_helper.cpp
```diff
@@ -1593,9 +1593,8 @@ void Ekf::setControlEVHeight()
 // and calculate a rotation matrix which rotates EV measurements into the EKF's navigation frame
 void Ekf::calcExtVisRotMat()
 {
-	// calculate the quaternion delta between the EKF and EV reference frames at the EKF fusion time horizon
-	Quatf quat_inv = _ev_sample_delayed.quat.inversed();
-	Quatf q_error =   quat_inv * _state.quat_nominal;
+	// calculate the quaternion delta between the EV and EKF reference frames at the EKF fusion time horizon
```
priseborough: // Calculate the quaternion delta that rotates from the EV to the EKF reference frame at the EKF fusion time horizon.

EKF/ekf_helper.cpp
```diff
@@ -1623,17 +1622,17 @@ void Ekf::calcExtVisRotMat()
...
 }
 
 // reset the estimated misalignment between the EV navigation frame and the EKF navigation frame
 // and update the rotation matrix which rotates EV measurements into the EKF's navigation frame
 void Ekf::resetExtVisRotMat()
 {
-	// calculate the quaternion delta between the EKF and EV reference frames at the EKF fusion time horizon
-	Quatf quat_inv = _ev_sample_delayed.quat.inversed();
-	Quatf q_error =   quat_inv * _state.quat_nominal;
+	// calculate the quaternion delta between the EV and EKF reference frames at the EKF fusion time horizon
```
priseborough: // calculate the quaternion delta from the EV to the EKF reference frame at the EKF fusion time horizon

EKF/ekf_helper.cpp
```diff
@@ -1623,17 +1622,17 @@ void Ekf::calcExtVisRotMat()
 
 	// convert filtered vector to a quaternion and then to a rotation matrix
 	q_error.from_axis_angle(_ev_rot_vec_filt);
-	_ev_rot_mat = quat_to_invrotmat(q_error);
+	q_error.normalize();
+	_ev_rot_mat = Dcmf(q_error); // rotation from EV reference to EKF reference
```
priseborough: There is already a PR dealing with how the quat_to_invrotmat function is used - see https://github.com/PX4/ecl/pull/635

EKF/ekf_helper.cpp
```diff
@@ -1649,7 +1648,7 @@ void Ekf::resetExtVisRotMat()
 	}
 
 	// reset the rotation matrix
-	_ev_rot_mat = quat_to_invrotmat(q_error);
+	_ev_rot_mat = Dcmf(q_error); // rotation from EV reference to EKF reference
```
priseborough: There is already a PR dealing with how the quat_to_invrotmat function is used - see https://github.com/PX4/ecl/pull/635

EKF/ekf_helper.cpp
```diff
@@ -1596,6 +1596,7 @@ void Ekf::calcExtVisRotMat()
 	// calculate the quaternion delta between the EV and EKF reference frames at the EKF fusion time horizon
 	Quatf q_error = _state.quat_nominal * _ev_sample_delayed.quat.inversed();
 	q_error.normalize();
+	q_error.canonicalize();
```
priseborough: This step appears unnecessary. If a canonical form is required for an external accessor, then the operation should be done there.

EKF/ekf_helper.cpp
```diff
@@ -1623,6 +1624,7 @@ void Ekf::calcExtVisRotMat()
 	// convert filtered vector to a quaternion and then to a rotation matrix
 	q_error.from_axis_angle(_ev_rot_vec_filt);
 	q_error.normalize();
+	q_error.canonicalize();
```
priseborough: This step appears unnecessary. If a canonical form is required for an external accessor, then the operation should be done there.

EKF/ekf.h
```diff
@@ -256,7 +256,7 @@ class Ekf : public EstimatorInterface
 	void get_ekf_soln_status(uint16_t *status);
 
 	// return the quaternion defining the rotation from the EKF to the External Vision reference frame
```
priseborough: // return the quaternion defining the rotation from the External Vision to the EKF reference frame

priseborough:
These changes are more extensive than required to fix the quaternion multiplication order error in the calculation of the EV to EKF rotation matrix, however the remaining changes will make debugging of alignment issues easier for integrators of VIO so are worth the additional resource utilisation.

The only remaining issue is why the conversion to canonical form has been added in two locations that appear to be unnecessary for internal use within the EKF.

priseborough:
The core change which is the fix to the rotation order in the calculation of the EV to EKF rotation matrix has been verified on replay using the https://review.px4.io/plot_app?log=6e27829d-714a-40ef-b9f9-856248630329 data set. This data set contains a combined yaw and pitch rotaton of the vehicle which is necessary to test the changes as the error does not show when the vehicle is close to level with small roll/pitch angles.

The quaternions in this data set show a yaw misalignment between the VIO and EKF solutions, but minimal difference in roll/pitch so the expected result for the EV to EKF rotation matrix is that the off diagonal terms in the third row/column are close to zero.

Print statements were added to the Ekf::calcExtVisRotMat() function to output the rotation matrix and the ekf2 replay was used with EKF2_AID_MASK 73 and 
EKF2_GPS_CHECK 31.

The old rotation order gave:

R = [0.849,0.516,-0.113]
    [-0.516,0.765,-0.386]
    [-0.112,0.386,0.916]

The new rotation order gave:

R = [0.758,0.652,0.027]
    [-0.652,0.758,-0.004]
    [-0.023,-0.014,1.000]

kamilritz:
@priseborough 
> // return the quaternion defining the rotation from the External Vision to the EKF reference frame
// return the quaternion defining the rotation from the External Vision to the EKF reference frame

I changed the comment in my newest commits.

I also reverted to `quat_to_invrotmat()` to respect PR https://github.com/PX4/ecl/pull/635.

The canonicalization of the quaternion is also removed. I was not sure if it would be needed, but it is also working without it.

The latest version of this PR is tested to work with the log 
> https://review.px4.io/plot_app?log=6e27829d-714a-40ef-b9f9-856248630329

priseborough:
Thank you @kamilritz

PR #635, opened 2019-08-22
EKF: Use math library for quaternion to rotation matrix conversions

priseborough:
This change fixes the following issues:

1) The quat_to_invrotmat function produced the same rotation matrix as the math library operator, but the name and comments indicated otherwise. This function was being used in several places where the math library could be used instead.
2) In several places where an inverse rotation was required, the rotation matrix was being calculated, then transposed.

The following changes have been made

1) Wherever the required rotation matrix can be calculated directly from the quaternion using the math library operator, this is done.
2) The quat_to_invrotmat function has been modified so that it outputs the inverse rotation to that produced by the math library and the comments have been updated to make this clear.
2) Wherever an inverse rotation matrix is required from a quaternion, the quat_to_invrotmat function is used instead of calculating the matrix and then transposing it.

SITL log attached:

https://review.px4.io/plot_app?log=397c0931-225d-423d-bf55-0a1014075998

### DIFF
--- a/EKF/drag_fusion.cpp
+++ b/EKF/drag_fusion.cpp
@@ -82,8 +82,7 @@ void Ekf::fuseDrag()
 	rel_wind(0) = vn - vwn;
 	rel_wind(1) = ve - vwe;
 	rel_wind(2) = vd;
-	Dcmf earth_to_body(_state.quat_nominal);
-	earth_to_body = earth_to_body.transpose();
+	Dcmf earth_to_body = quat_to_invrotmat(_state.quat_nominal);
 	rel_wind = earth_to_body * rel_wind;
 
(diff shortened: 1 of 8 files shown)


priseborough:
@RomanBapst can you please review this. I would like to get this in before we deal with https://github.com/PX4/ecl/pull/634

priseborough:
https://github.com/PX4/ecl/pull/634 has been merged so I'll fix a couple of minor conflicts and rebase.

mhkabir:
@priseborough can we get this in?

kamilritz:
@priseborough @RomanBapst Would be great to have this in

RomanBapst:
@kamilritz Looks good to me.

PR #636, opened 2019-08-23
Fixing wrong label of quaternion direction

kamilritz:
Some comments were wrongly stating that the EKF quaternion was representing the rotation from "nav to body" instead from "body to nav"


EKF/mag_fusion.cpp
```diff
@@ -537,7 +537,7 @@ void Ekf::fuseHeading()
 		H_YAW[2] = t8*t14*(-q1*t3+q1*t4+q1*t5+q1*t6-q0*q2*q3*2.0f)*2.0f;
 		H_YAW[3] = t8*t14*(q0*t3-q0*t4+q0*t5+q0*t6-q1*q2*q3*2.0f)*2.0f;
 
-		/* Calculate the 312 sequence euler angles that rotate from earth to body frame
+		/* Calculate the 312 sequence euler angles that rotate from body to earth frame
```
priseborough: The RPY angles returned by this calculation are from earth to body. For example if the F body axis is aligned with the positive E axis and R body axis is aligned with the negative N axis (+90 degrees yaw), then _R_to_earth(0, 1) = -1 , _R_to_earth(1, 1) = 0 and yaw = atan2f(-_R_to_earth(0, 1), _R_to_earth(1, 1)) = atan2(1,0) = +90 degrees.

kamilritz: Thanks. I removed this line from the commit.

PR #638, opened 2019-09-05
implemented synthesized magnetometer Z measurement

RomanBapst:
This PR intends to improve yaw / magnetic field vector estimation for setups where the quality of the magnetomter Z component is bad. This is e.g. the case when there is a strong coupling between motor current and magnetometer Z measurement.

If this feature is enabled and the global position on earth is known then the estimator will calculate the magnetometer Z component based on the knowledge of the direction and strength of the magnetic field vector.
The synthesized component is used for the heading fusion but not for the 3D fusion. In the 3D fusion we rely on the x/y components of the magnetometer measurement to estimate the 3D magnetic field vector.

**Follow up work**
1) Since we don't fuse the synthetic Z measurement for 3D fusion we need to make sure to disable the mag Z bias state.
2) If we have the knowledge about the earth magnetic field vector we should add it as a direct fusion. That should be very cheap as we can observe the states directly. Currently we only fuse the declination as an observation in order to put some constraints on the earth fields states, however, we can go even further and constrain the vector itself.

**Testing**
I have tested this PR on a small racer vehicle. In order to do realistic tests I distorted the magnetometer Z measurements from the sensor as a function of the demanded throttle.

Heading innovations using magnetometer Z data:

bad_heading

Heading innovation using synthesized measurement:
heading_good

I also did a SITL test where I set the max tilt angle of the drone to 70 degrees and flew a maneuver where the drone would spiral around in yaw with a 70 degrees bank angle.
I misused the hagl_innov field of the ekf2_innovations message to store the synthesized value for the magnetomter Z component. As seen in the plot below the actual sensor measurement is calculated quite accurately.
image

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -357,6 +357,9 @@ struct parameters {
 
 	// control of on-ground movement check
 	float is_moving_scaler{1.0f};		///< gain scaler used to adjust the threshold for the on-ground movement detection. Larger values make the test less sensitive.
+
+	// compute synthetic magnetomter Z value if possible
+	int32_t synthesize_mag_z{0};
 };
 
 struct stateSample {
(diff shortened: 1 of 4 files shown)


priseborough:
@RomanBapst can you please push a patch for the conflict.

RomanBapst:
@priseborough Done.

PR #639, opened 2019-09-11
Flow takeoff improvements

nicovanduijn:
This PR improves the behavior when taking off using only optical flow. Specifically:
 - It publishes a "fake range measurement" when the vehicle is on the ground and range sensor data contains garbage. This change now publishes the value of the parameter `rng_gnd_clearance` in that case.
- It adds the range sensor quality to the terrain estimator. Fundamentally, the information is fused into the existing `_rng_hgt_faulty` flag. It is also used in the decision-making of the above point, as well as in the initialization of the height above ground `initHagl()`.

I've tested this on two different vehicles on PX4 with different range sensors:
- Using the CM8JL65, a lidar sensor which is quite accurate from 0.1m ~ 6m
- Using the MB12xx, a sonar which starts working around 0.65m ~ 5m

The drivers of the two vary a bit: the CM8JL65 doesn't actually set the quality flag, but correctly reports the saturated min/max values, while my (custom) driver for the MB12xx reports garbage but correctly identifies it as such and sets the quality metric accordingly.

Logs:
CM8JL65: https://review.px4.io/plot_app?log=01deedd4-2afb-4b09-ad37-726814424121
image

MB12xx: https://review.px4.io/plot_app?log=bc610be1-089b-4b57-90f6-937d97326797
image

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -133,6 +133,7 @@ struct baroSample {
 struct rangeSample {
 	float       rng;	///< range (distance to ground) measurement (m)
 	uint64_t    time_us;	///< timestamp of the measurement (uSec)
+	uint8_t	    quality; ///< quality indicator between 0 and 255
 };
 
 struct airspeedSample {
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -131,9 +131,9 @@ struct baroSample {
 };
 
 struct rangeSample {
-	float       rng;	///< range (distance to ground) measurement (m)
+	float       rng;	    ///< range (distance to ground) measurement (m)
 	uint64_t    time_us;	///< timestamp of the measurement (uSec)
-	uint8_t	    quality; ///< quality indicator between 0 and 255
+	int8_t	    quality;    ///< Signal quality in percent (0...100%), where 0 = invalid signal, 100 = perfect signal, and -1 = unknown signal quality.
 };
 
 struct airspeedSample {
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1198,7 +1198,7 @@ void Ekf::checkRangeDataValidity()
 		return;
 	} else {
 		// reset fault status when we get new data
-		_rng_hgt_faulty = false;
+		_rng_hgt_faulty = (_range_sample_delayed.quality == 0);
 	}
 
 	// Check if excessively tilted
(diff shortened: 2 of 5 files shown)


JohannesBrand:
Do you have before-after plots?

JohannesBrand:
Great! Thanks for the explanation.

EKF/common.h
```diff
@@ -133,6 +133,7 @@ struct baroSample {
 struct rangeSample {
 	float       rng;	///< range (distance to ground) measurement (m)
 	uint64_t    time_us;	///< timestamp of the measurement (uSec)
+	uint8_t	    quality; ///< quality indicator between 0 and 255
```
priseborough: The signal quality metric in the PX4 distance_sensor topic is cast as a int8 with a range between 0 and 100 when quality is known and -1 when it is not. We should keep that mapping.

EKF/control.cpp
```diff
@@ -1244,7 +1244,7 @@ void Ekf::checkRangeDataValidity()
 			}
 
 			_control_status.flags.rng_stuck = true;
-			_rng_hgt_faulty = true;
+			_rng_hgt_faulty = (_range_sample_delayed.quality == 0);
```
priseborough: If the range reading is stuck we should flag it as unhealthy regardless of quality metric

EKF/estimator_interface.cpp
```diff
@@ -314,7 +314,7 @@ void EstimatorInterface::setAirspeedData(uint64_t time_usec, float true_airspeed
 	}
 }
 
-void EstimatorInterface::setRangeData(uint64_t time_usec, float data)
+void EstimatorInterface::setRangeData(uint64_t time_usec, float data, uint8_t quality)
```
priseborough: It is preferable to use the same data type and definitions for quality as the distance_sensor topic.

EKF/estimator_interface.h
```diff
@@ -191,7 +191,7 @@ class EstimatorInterface
 	void setAirspeedData(uint64_t time_usec, float true_airspeed, float eas2tas);
 
 	// set range data
-	void setRangeData(uint64_t time_usec, float data);
+	void setRangeData(uint64_t time_usec, float data, uint8_t quality);
```
priseborough: Better to use type and definition as per distance_sensor topic.

EKF/common.h
nicovanduijn: good point

EKF/control.cpp
nicovanduijn: nice catch, this is blatantly wrong

Issue #631, opened 2019-08-07
External vision velocity estimate usage


kamilritz:
From the vehicle_visual_odometry msg only the position estimate is used at the moment. Current VIO systems, such as the Realsense T265, output also a velocity estimates. It could be interesting to enable the EKF to also handle EV velocities, since these are likely to be continuous. While the position estimate from the EV system can include jumps due to re-localization or loop-closures. 

I would be happy to investigate this further.

mhkabir:
I'm working on this right now.

mhkabir:
Done!

Issue #397, opened 2018-02-08
EV and GPS fusion


ChristophTobler:
Currently the only way of fusing EV and GPS is by rotating EV measurements and fuse them as incremental position changes (assuming that the EV measurements are not in NED which is almost always the case). The downside of this is that we can't fully make use of the EV's performance. With the incremental position fusion the vehicle drifts more.

I see two options:

1. We keep the yaw from the magnetometer all the time (actual NED) and add an offset and rotation to https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L289. That way we fuse the actual position.

2. We use yaw from EV if GPS isn't valid and switch to the mag if GPS gets valid and reset all EV related parameters (should also work the other way around). So if no GPS is available the can fuse actual EV position and if GPS is available we use the incremental approach.

mhkabir:
The first approach sounds better to me.

EliaTarasov:
In a number of cases it's better to avoid using magnetometer at all. I would vote for the second option, because it the first case we are totally depend on mag.

ChristophTobler:
I implemented option 1 ([branch](https://github.com/PX4/ecl/tree/ekf_ev_pos_fusion_1)) as it didn't require too much changes and did some tests.

Current Implementation (incremental pos fusion) doesn't work very good. When stationary on the ground the position drifts a lot. In air it's better (not sure why...) and can be used to fly in position mode. But the performance isn't very good... see below (EV is what the vehicle actually did -> almost ground truth)
![ev_incremental_fusion](https://user-images.githubusercontent.com/14769636/36027639-6a5501c4-0d9c-11e8-9b1f-34aacde0e7a4.png)

When using option 1, performance is much better and there is no drift.
![ev_abs_fusion_with_rot_offset](https://user-images.githubusercontent.com/14769636/36027662-80888e34-0d9c-11e8-9950-da1e0d6ae470.png)

I also did some testing in sitl with switching GPS on and off and switching ev on and off (hardcoded counters) and it seems to work nicely. But this would need so more testing and review. I could open a PR with that branch if we all agree.

All tests include the rotation fix https://github.com/PX4/ecl/pull/396
logs
* master (incremental pos fusion): https://review.px4.io/plot_app?log=c933c22f-ea92-4882-b186-8283543cfc9c
* option 1 without GPS: https://review.px4.io/plot_app?log=6d6f16b9-58d0-4c95-8e0b-48e1f363ff74
[...]

PR #642, opened 2019-09-17
External vision velocity

kamilritz:
This PR allows ECL to fuse velocity measurements from an external vision sensor such as the Realsense T265. In addition, it also improves on the external vision position fusion, such that the external vision measurement fusion is no usable with several EKF_AID_MASK configurations. 

Supported configuration for the EKF_AID_MASK are:
1      -     GPS
73    -     GPS + EV_POS + ROTATE_EV (incremental ev pos fusion, not recommended, use EV_VEL instead)
321  -     GPS + EV_VEL  + ROTATE_EV (recommended)
24    -     EV_POS + EV_YAW
72    -     EV_POS + ROTATE_EV
272  -     EV_VEL + EV_YAW
320  -     EV_VEL + ROTATE_EV
280  -     EV_POS + EV_VEL + EV_YAW
328  -     EV_POS + EV_VEL + ROTATE_EV

If you use no GPS information, you are now free to decide if you want your EKF reference frame to be aligned with true ENU frame (use ROTATE_EV) or aligned with the external vision reference frame (use EV_YAW).

The EKF_AID_MASK = 9 (GPS + EV_POS) is not supported and should not be used.

All these configurations are tested in EKF replay on log: https://review.px4.io/plot_app?log=bd994025-2312-4ee9-a7b2-f6ab2d65c6b8 and the majority of the configurations were validated on real world test flights.
Make sure to have EVV_NOISE=0.1, EVV_GATE=3, EVP_NOISE=0.1 and EVP_GATE=5.

@priseborough Thank you very much for your great help and your valuable inputs.
#631 #397

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -80,8 +80,8 @@ struct flow_message {
 };
 
 struct ext_vision_message {
-	Vector3f posNED;	///< XYZ position in earth frame (m) - Z must be aligned with down axis
-	Vector3f velNED;	///< XYZ velocity in earth frame (m/sec) - Z must be aligned with down axis
+	Vector3f pos;	///< XYZ position in earth frame (m) - Z must be aligned with down axis
+	Vector3f vel;	///< XYZ velocity in earth frame (m/sec) - Z must be aligned with down axis
 	Quatf quat;		///< quaternion defining rotation from body to earth frame
 	float posErr;		///< 1-Sigma horizontal position accuracy (m)
 	float hgtErr;		///< 1-Sigma height accuracy (m)
@@ -153,8 +153,8 @@ struct flowSample {
 };
 
 struct extVisionSample {
-	Vector3f posNED;	///< XYZ position in earth frame (m) - Z must be aligned with down axis
-	Vector3f velNED;	///< XYZ velocity in earth frame (m/sec) - Z must be aligned with down axis
+	Vector3f pos;	///< XYZ position in earth frame (m) - Z must be aligned with down axis
+	Vector3f vel;	///< XYZ velocity in earth frame (m/sec) - Z must be aligned with down axis
 	Quatf quat;		///< quaternion defining rotation from body to earth frame
 	float posErr;		///< 1-Sigma horizontal position accuracy (m)
 	float hgtErr;		///< 1-Sigma height accuracy (m)
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -260,8 +260,8 @@ struct parameters {
 	float pos_noaid_noise{10.0f};		///< observation noise for non-aiding position fusion (m)
 	float baro_noise{2.0f};			///< observation noise for barometric height fusion (m)
 	float baro_innov_gate{5.0f};		///< barometric and GPS height innovation consistency gate size (STD)
-	float posNE_innov_gate{5.0f};		///< GPS horizontal position innovation consistency gate size (STD)
-	float vel_innov_gate{5.0f};		///< GPS velocity innovation consistency gate size (STD)
+        float gps_pos_innov_gate{5.0f};		///< GPS horizontal position innovation consistency gate size (STD)
+        float gps_vel_innov_gate{5.0f};		///< GPS velocity innovation consistency gate size (STD)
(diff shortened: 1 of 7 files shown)


mhkabir:
> If you use no GPS information, you are now free to decide if you want your EKF reference frame to be aligned with true ENU frame (use ROTATE_EV) or aligned with the external vision reference frame (use EV_YAW).

If you're not using GPS, the recommended way would still be to use EV_YAW, no? Since ROTATE_EV case still needs to depend on the magnetometer for heading.

@jkflying @priseborough @kamilritz Also along this thread, more generally, we should be able to handle two very important cases where we are configured to run GPS + vision:

1. Starting up in a GPS denied situation where mag cannot be relied on, and therefore you don't know the true NED frame alignment. In this case, we should probably do absolute EV position and EV yaw fusion (EVP, EVV, EV_YAW) till it is detected that we are outdoors. At that point the reference frame can be reset to true NED, and relative fusion continued (GPS, EVV, ROTATE_EV). This will cause a jump in the local frame's rotation and position.

2. Flying outdoors, then entering an area where only vision is viable, with both GPS and mag being unreliable. In this case, we would need to fuse the absolute EV pos and *rotated* EV Yaw to prevent excessive drifting from only vel fusion. This case differs from (1) in that the NED alignment is already known, and we may want to retain some of that information. 

I see two ways to handle this:
(i) Only update the EV rotation quaternion and EV pos offset (see below) when global heading observations are available (outdoor mode), and use it to pre-rotate the EV yaw before absolute fusion. We would additionally need to track the position offset between EV and EKF, and use this to first transform EV meas into the EKF frame before absolute fusion. To be noted - while indoors, the local frame will drift slowly in rotation from true NED.
(ii) Do a hard reset to the EV frame (like proposed to do to NED at the end of (1)) when we switch to indoor mode. This will cause a jump in local frame rotation and position. 

Both of these will require us to really robustify indoor mode detection, and good mag/gpsyaw handling to be really clear about when we have reliable global heading available. The initialization of the GPS origin for both the above also needs to be discussed.

Automatically handling these cases will get rid of a lot of the parameter combos which may be confusing for new users. 

This is mostly just a collection of my thoughts. We should probably sit down and discuss this quickly so that we can implement.

priseborough:
The default behaviour should be to use relative heading from EV_YAW when operating without GPS due to problems with magnetometer performance in an indoor environment, however we should still allow the user to use an absolute heading reference to aid with development testing if that is what they want to do.

EKF/common.h
```diff
@@ -80,10 +80,12 @@ struct flow_message {
 };
 
 struct ext_vision_message {
-	Vector3f posNED;	///< measured NED position relative to the local origin (m)
-	Quatf quat;		///< measured quaternion orientation defining rotation from NED to body frame
+	Vector3f posNED;	///< XYZ position in earth frame (m) - Z must be aligned with down axis
```
mhkabir: I think it is now more appropriate for these to be renamed to pos and vel, without the NED prefixes.

EKF/common.h
```diff
@@ -151,10 +153,12 @@ struct flowSample {
 };
 
 struct extVisionSample {
-	Vector3f posNED;	///< measured NED position relative to the local origin (m)
-	Quatf quat;		///< measured quaternion orientation defining rotation from NED to body frame
+	Vector3f posNED;	///< XYZ position in earth frame (m) - Z must be aligned with down axis
+	Vector3f velNED;	///< XYZ velocity in earth frame (m/sec) - Z must be aligned with down axis
```
mhkabir: Same as above

EKF/common.h
kamilritz: Yes, that is true

priseborough:
I think we should do this in tow stages. This PR adds the underlying fusion modes required and is the basis for better handover logic. I support merging this and them using specific use cases like the ones @mhkabir raised to further develop the logic and corner case handling.

mhkabir:
@priseborough Thanks for merging. 

Next time, for large PRs like this, please try to squash commits before you merge because otherwise we end up with this a bunch of these minor non-building commits which makes bisecting a pain. Thanks!

PR #644, opened 2019-09-18
Add timestamp to ECL print statements,

kamilritz:
which is helpful for EKF replay debugging.

@priseborough Is `_imu_sample_delayed.time_us` the right time to stamp the messages?

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -71,7 +71,6 @@ void Ekf::controlFusionModes()
 
 			} else if (_control_status.flags.rng_hgt) {
 				ECL_INFO("EKF aligned, (range height, IMU buf: %i, OBS buf: %i)", (int)_imu_buffer_length, (int)_obs_buffer_length);
-
 			} else {
 				ECL_ERR("EKF aligned, (unknown height, IMU buf: %i, OBS buf: %i)", (int)_imu_buffer_length, (int)_obs_buffer_length);
 			}
@@ -181,14 +180,14 @@ void Ekf::controlExternalVisionFusion()
 				if (_params.fusion_mode & MASK_USE_EVPOS && !_control_status.flags.ev_pos) {
 					_control_status.flags.ev_pos = true;
 					resetPosition();
-					ECL_INFO("EKF commencing external vision position fusion");
+					ECL_INFO_TIMESTAMPED("EKF commencing external vision position fusion");
 				}
 
 				// turn on use of external vision measurements for velocity
 				if (_params.fusion_mode & MASK_USE_EVVEL && !_control_status.flags.ev_vel) {
 					_control_status.flags.ev_vel = true;
 					resetVelocity();
-					ECL_INFO("EKF commencing external vision velocity fusion");
+					ECL_INFO_TIMESTAMPED("EKF commencing external vision velocity fusion");
 				}
 
 				if ((_params.fusion_mode & MASK_ROTATE_EV) && !(_params.fusion_mode & MASK_USE_EVYAW)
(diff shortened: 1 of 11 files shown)


dagar:
Can you do this without double precision math? These print statements are already relatively expensive on real hardware (stm32f4).

kamilritz:
@dagar Double multiplication is now removed. The time is now printed in microseconds.

kamilritz:
@jkflying Thanks, this is now resolved

dagar:
> @dagar Double multiplication is now removed. The time is now printed in microseconds.

You're still casting to double though?

mhkabir:
@kamilritz Just print the microseconds timestamp using `%llu` without casting.

kamilritz:
@mhkabir Are you sure that this is working on all platforms. To me it seemed that some platforms interpret uint64_t as long unsigned int (e.g in EKF replay - make px4_sitl none) and some as long long unsigned int.

mhkabir:
You're right - PRIu64 should be more portable. https://en.cppreference.com/w/c/types/integer

kamilritz:
@dagar @mhkabir It is now using uint64 directly and is adding the timestamps in posix builds only.

EKF/gps_yaw_fusion.cpp
```diff
@@ -174,7 +174,7 @@ void Ekf::fuseGpsAntYaw()
 
 		// we reinitialise the covariance matrix and abort this fusion step
 		initialiseCovariance();
-		ECL_ERR("EKF GPS yaw fusion numerical error - covariance reset");
+		ECL_ERR("%.0lf: EKF GPS yaw fusion numerical error - covariance reset", (double)_imu_sample_delayed.time_us);
```
jkflying: Missed this one

mhkabir:
I'm not fully sold on the idea using the delayed sample's timestamp. I feel like a log message should be using the actually current time (like hrt_absolute_time(), as opposed to one which can be delayed based on the fusion horizon.

jkflying:
The issue is that during replay using the hrt_absolute_time() gives system time, which is useless.

priseborough:
@kamilritz Can you please push a patch for the conflict.

EKF/control.cpp
```diff
@@ -181,21 +180,22 @@ void Ekf::controlExternalVisionFusion()
...
-					ECL_INFO("EKF commencing external vision velocity fusion");
+					ECL_INFO_TIMESTAMPED("EKF commencing external vision velocity fusion");
 				}
 
 				if ((_params.fusion_mode & MASK_ROTATE_EV) && !(_params.fusion_mode & MASK_USE_EVYAW)
 					&& !_ev_rot_mat_initialised)  {
 					// Reset transformation between EV and EKF navigation frames when starting fusion
 					resetExtVisRotMat();
 					_ev_rot_mat_initialised = true;
+          ECL_INFO_TIMESTAMPED("EKF external vision aligned");
```
mhkabir: Indent

EKF/control.cpp
```diff
@@ -1682,7 +1682,7 @@ void Ekf::controlVelPosFusion()
 				_fuse_hpos_as_odom = false;
 
 				if (_time_last_fake_gps != 0) {
-					ECL_WARN("EKF stopping navigation");
+				ECL_WARN_TIMESTAMPED("EKF stopping navigation");
```
mhkabir: Indent

kamilritz: thanks

kamilritz:
@mhkabir @priseborough LGTM

kamilritz:
@priseborough @julianoes @bkueng it is squashed now, could someone please merge it. Thanks!

PR #645, opened 2019-09-19
data_validator_group: add a method to get the state of the sensor

dusan19:
Needed for
https://github.com/PX4/Firmware/pull/12989


kamilritz:
@dusan19 Is this still needed?

priseborough:
How has this been tested?

dusan19:
the px4 firmware commit that uses this will most likely not be merged anytime soon or at all.

Also there is no harm in merging this since it only extends the dataValidatorGroup interface such that it provides the error state of a specific sensor.

I tested this by calling this function from the sensors.cpp and printing the result.

PR #647, opened 2019-09-25
Fix print statements cast error on posix firmware builds

kamilritz:


kamilritz:
@julianoes We noticed that we did not passed Firmware CI based on posix fcu boards. This fixes it. Could you please merge this as well? Thanks!

mhkabir:
@kamilritz Why not use PRIu64 instead of this casting? (also mentioned this in the PR before)

PR #650, opened 2019-10-03
[Terrain estimator] Continuously reset terrain height on ground using known ground clearance

bresch:
Ground clearance is the best estimate we can have when we are on the ground and we should not rely on a distance sensor that could output bad data before flight. 
This also helps when the drone is carried over as it avoids starting with a crazy downward distance for optical flow scaling.

#### Prior to this PR:
The `dist_bottom` starts at 3 meters because it initializes on wrong range finder data.
bad_terrain_init

#### With this PR:
The `dist_bottom` starts from the ground clearance and reject properly measurements until they actually make sense.
rng_wrong_meas_on_gnd

### DIFF
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -45,33 +45,44 @@
 
 bool Ekf::initHagl()
 {
+	bool initialized = false;
 	// get most recent range measurement from buffer
 	const rangeSample &latest_measurement = _range_buffer.get_newest();
 
-	if (!_rng_hgt_faulty && (_time_last_imu - latest_measurement.time_us) < (uint64_t)2e5 && _R_rng_to_earth_2_2 > _params.range_cos_max_tilt) {
+	if (!_control_status.flags.in_air) {
+		// if on ground, do not trust the range sensor, but assume a ground clearance
+		_terrain_vpos = _state.pos(2) + _params.rng_gnd_clearance;
+		// use the ground clearance value as our uncertainty
+		_terrain_var = sq(_params.rng_gnd_clearance);
+		initialized = true;
+
+	} else if (!_rng_hgt_faulty
+		   && (_time_last_imu - latest_measurement.time_us) < (uint64_t)2e5
+		   && _R_rng_to_earth_2_2 > _params.range_cos_max_tilt) {
 		// if we have a fresh measurement, use it to initialise the terrain estimator
 		_terrain_vpos = _state.pos(2) + latest_measurement.rng * _R_rng_to_earth_2_2;
 		// initialise state variance to variance of measurement
 		_terrain_var = sq(_params.range_noise);
 		// success
-		return true;
+		initialized = true;
+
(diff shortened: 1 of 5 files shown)


EKF/terrain_estimator.cpp
```diff
@@ -80,7 +91,7 @@ void Ekf::runTerrainEstimator()
 	checkRangeDataContinuity();
 
 	// Perform initialisation check
```
priseborough: Suggest we add a comment to indicate that continually re-initializing on the ground is deliberate.

RomanBapst:
@bresch I was about to do another cleanup regarding the entire logic of checking range data validity.
There are some things I wanted to get rid of such as:
1) We are replacing range data with known ground clearance at multiple locations instead of doing it at a central point, e.g.
https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L1170
https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L1275
https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L1084
https://github.com/PX4/ecl/blob/master/EKF/terrain_estimator.cpp#L106

So I was wondering if we could consolidate this into a single location such that we don't need to have those checks all over the code. I think this would also solve your case as range data would automatically be set to the ground clearance when the vehicle is not in air.
What do you think?

bresch:
Note for myself: remove `if ((_rng_hgt_faulty || !_range_data_ready) && !_control_status.flags.in_air && _vehicle_at_rest &&`

bresch:
The last commit changes the `snake_case` functions to `camelCase` to follow the general convention. I also ran astyle for general formatting.
For now, the `camelCase` function naming isn't used for most of the EKF functions, but I think we should move towards this every time we do an update somewhere, do you agree? Or is this a no-no because it breaks the interface (i.e.:other projects than PX4 using ecl should update their code)

bresch:
todo: wrap new functions into old function call for backward compatibility

bresch:
@RomanBapst Can you review and approve again please?

EKF/terrain_estimator.cpp
```diff
@@ -283,21 +288,20 @@ void Ekf::fuseFlowForTerrain()
...
 {
 	return _hagl_valid;
 }
 
-// determine terrain validity
-void Ekf::update_terrain_valid()
+void Ekf::updateTerrainValidity()
 {
 	// we have been fusing range finder measurements in the last 5 seconds
-	bool recent_range_fusion = (_time_last_imu - _time_last_hagl_fuse) < 5*1000*1000;
```
RomanBapst: @bresch If I remember correctly this was done to avoid a cast from double to unit64, since 5e6 is stored as a double.

RomanBapst: @bresch I think we should use time literals as we used them in the firmware. Here it would be 5 * 1_s

priseborough:
Thanks @bresch

PR #652, opened 2019-10-08
Moving towards multi-sensor-per-same-measurement fusion

kamilritz:
The current EKF was designed to fuse velocity and position measurement in an efficient way. The current implementation does this by selecting a sensor source for each measurement and performing a measurement update containing measurement from different  sensors.
Fusing the same measurement from different sources, e.g. GPS velocity and EV velocity, is also possible. The issue is that this happens through the same global `_vel_pos_innov` variable. Which can potentially be used for multiple sensors in the same loop. In general, the whole EKF interface is not designed to support logging of  the same measurement from different sensors.

This PR tries to solve this issue by adapting the `fuseVelPoseHeight()` function. The function interface is changed, such that it takes multiple arguments and runs with a generic velPosHeight sensor. Therefore the computation of the height innovation is moved outside to the `controlHeightFusion()` function.
The `controlGPSFusion()`, `controlExternalVisionFusion()`, `controlHeightFusion()`, `controlAuxFusion()` and the new `controlFakePosFusion()` functions are all passing the innovation/-variances/-gates, observation-variances and a new fusion_mask as an argument in the function call to `fuseVelPoseHeight()`.

The necessary changes on the Firmware side are laid out [here](https://github.com/PX4/Firmware/pull/13127).

These changes were test flown with different velPos and height sources. The test flight is based on flying a square followed by a yaw spin. The flight performance was in all cases as expected. The logs of the test flights can be found here:

- [GPS and GPS height](https://review.px4.io/plot_app?log=25984331-b5d9-462f-91df-3bde5d5d8c1d)
- [GPS/Flow and Baro heihgt](https://review.px4.io/plot_app?log=f7beb996-93e1-4e24-bd94-af3a9fe876a6)
- [Flow and Range height](https://review.px4.io/plot_app?log=bb60ddde-6f71-4542-bf30-22423d0a8766)
- [EVPOS/EVVEL/EVYAW and vision height](https://review.px4.io/plot_app?log=115e72a7-f5cd-46b3-9616-711cd9474e0b)
- [EVVEL/EVYAW and vision height](https://review.px4.io/plot_app?log=c5159c1b-c6d4-4554-befc-9d250e36a384)
- [GPS/EVVEL/EVROT and vision height](https://review.px4.io/plot_app?log=a4e6d982-2a41-4a19-87c8-a8e0888ce70c)

@priseborough I would appreciate if you could have a look at this and give me rough feedback. So that I can move in the right direction.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -210,6 +210,12 @@ struct auxVelSample {
 #define RNG_MAX_INTERVAL  (uint64_t)2e5	///< Maximum allowable time interval between range finder  measurements (uSec)
 #define EV_MAX_INTERVAL   (uint64_t)2e5	///< Maximum allowable time interval between external vision system measurements (uSec)
 
+// VelPos measurement bundles indices
+#define HVEL 0	///< x and y velocity bundle index
+#define VVEL 1	///< z velocity bundle index
+#define HPOS 2	///< x and y position bundle index
+#define VPOS 3	///< z position bundle index
+
 // bad accelerometer detection and mitigation
 #define BADACC_PROBATION  (uint64_t)10e6	///< Period of time that accel data declared bad must continuously pass checks to be declared good again (uSec)
 #define BADACC_BIAS_PNOISE	4.9f	///< The delta velocity process noise is set to this when accel data is declared bad (m/sec**2)
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -367,8 +367,8 @@ struct stateSample {
 	Quatf  quat_nominal;	///< quaternion defining the rotation from body to earth frame
 	Vector3f    vel;	///< NED velocity in earth frame in m/s
 	Vector3f    pos;	///< NED position in earth frame in m
-	Vector3f    gyro_bias;	///< delta angle bias estimate in rad
-	Vector3f    accel_bias;	///< delta velocity bias estimate in m/s
+	Vector3f    delta_ang_bias;	///< delta angle bias estimate in rad
+	Vector3f    delta_vel_bias;	///< delta velocity bias estimate in m/s
 	Vector3f    mag_I;	///< NED earth magnetic field in gauss
 	Vector3f    mag_B;	///< magnetometer bias estimate in body frame in gauss
 	Vector2f    wind_vel;	///< wind velocity in m/s
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -210,12 +210,6 @@ struct auxVelSample {
 #define RNG_MAX_INTERVAL  (uint64_t)2e5	///< Maximum allowable time interval between range finder  measurements (uSec)
 #define EV_MAX_INTERVAL   (uint64_t)2e5	///< Maximum allowable time interval between external vision system measurements (uSec)
 
-// VelPos measurement bundles indices
-#define HVEL 0	///< x and y velocity bundle index
-#define VVEL 1	///< z velocity bundle index
-#define HPOS 2	///< x and y position bundle index
-#define VPOS 3	///< z position bundle index
-
 // bad accelerometer detection and mitigation
 #define BADACC_PROBATION  (uint64_t)10e6	///< Period of time that accel data declared bad must continuously pass checks to be declared good again (uSec)
 #define BADACC_BIAS_PNOISE	4.9f	///< The delta velocity process noise is set to this when accel data is declared bad (m/sec**2)
@@ -406,18 +400,19 @@ union fault_status_u {
 // define structure used to communicate innovation test failures
 union innovation_fault_status_u {
 	struct {
-		bool reject_vel_NED: 1;		///< 0 - true if velocity observations have been rejected
-		bool reject_pos_NE: 1;		///< 1 - true if horizontal position observations have been rejected
-		bool reject_pos_D: 1;		///< 2 - true if true if vertical position observations have been rejected
-		bool reject_mag_x: 1;		///< 3 - true if the X magnetometer observation has been rejected
-		bool reject_mag_y: 1;		///< 4 - true if the Y magnetometer observation has been rejected
-		bool reject_mag_z: 1;		///< 5 - true if the Z magnetometer observation has been rejected
-		bool reject_yaw: 1;		///< 6 - true if the yaw observation has been rejected
-		bool reject_airspeed: 1;	///< 7 - true if the airspeed observation has been rejected
-		bool reject_sideslip: 1;	///< 8 - true if the synthetic sideslip observation has been rejected
-		bool reject_hagl: 1;		///< 9 - true if the height above ground observation has been rejected
-		bool reject_optflow_X: 1;	///< 10 - true if the X optical flow observation has been rejected
-		bool reject_optflow_Y: 1;	///< 11 - true if the Y optical flow observation has been rejected
+		bool reject_hor_vel: 1;		///< 0 - true if horizontal velocity observations have been rejected
+		bool reject_ver_vel: 1;		///< 1 - true if vertical velocity observations have been rejected
+		bool reject_hor_pos: 1;		///< 2 - true if horizontal position observations have been rejected
+		bool reject_ver_pos: 1;		///< 3 - true if true if vertical position observations have been rejected
+		bool reject_mag_x: 1;		///< 4 - true if the X magnetometer observation has been rejected
+		bool reject_mag_y: 1;		///< 5 - true if the Y magnetometer observation has been rejected
+		bool reject_mag_z: 1;		///< 6 - true if the Z magnetometer observation has been rejected
+		bool reject_yaw: 1;		///< 7 - true if the yaw observation has been rejected
+		bool reject_airspeed: 1;	///< 8 - true if the airspeed observation has been rejected
+		bool reject_sideslip: 1;	///< 9 - true if the synthetic sideslip observation has been rejected
+		bool reject_hagl: 1;		///< 10 - true if the height above ground observation has been rejected
+		bool reject_optflow_X: 1;	///< 11 - true if the X optical flow observation has been rejected
+		bool reject_optflow_Y: 1;	///< 12 - true if the Y optical flow observation has been rejected
 	} flags;
 	uint16_t value;
 
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -157,15 +157,14 @@ void Ekf::controlFusionModes()
 	controlDragFusion();
 	controlHeightFusion();
 
-	// For efficiency, fusion of direct state observations for position and velocity is performed sequentially
-	// in a single function using sensor data from multiple sources (GPS, baro, range finder, etc)
-	controlVelPosFusion();
-
-	// Additional data from an external vision pose estimator can be fused.
+	// Additional data odoemtery data from an external estimator can be fused.
 	controlExternalVisionFusion();
 
-	// Additional NE velocity data from an auxiliary sensor can be fused
+	// Additional horizontal velocity data from an auxiliary sensor can be fused
 	controlAuxVelFusion();
+
+	// Fake position measurement for constraining drift when no other velocity or position measurements
+	controlFakePosFusion();
 
 	// check if we are no longer fusing measurements that directly constrain velocity drift
 	update_deadreckoning_status();
@@ -270,23 +269,13 @@ void Ekf::controlExternalVisionFusion()
 			}
 		}
 
-		// determine if we should start using the height observations
-		if (_params.vdist_sensor_type == VDIST_SENSOR_EV) {
-			// don't start using EV data unless data is arriving frequently
-			if (!_control_status.flags.ev_hgt && ((_time_last_imu - _time_last_ext_vision) < (2 * EV_MAX_INTERVAL))) {
-				setControlEVHeight();
-				resetHeight();
-			}
-		}
-
-		// determine if we should use the vertical position observation
-		if (_control_status.flags.ev_hgt) {
-			_fuse_height = true;
-		}
+		bool ev_fuse_mask[4]{};
+		float ev_obs_var[6]{};
+		float ev_innov_gate[4]{};
 
 		// determine if we should use the horizontal position observations
 		if (_control_status.flags.ev_pos) {
-			_fuse_pos = true;
+			ev_fuse_mask[HPOS] = true;
 
 			// correct position and height for offset relative to IMU
 			Vector3f pos_offset_body = _params.ev_pos_body - _params.imu_pos_body;
@@ -316,11 +305,11 @@ void Ekf::controlExternalVisionFusion()
 					ev_delta_pos = _ev_rot_mat * ev_delta_pos;
 
 					// use the change in position since the last measurement
-					_vel_pos_innov[3] = _state.pos(0) - _hpos_pred_prev(0) - ev_delta_pos(0);
-					_vel_pos_innov[4] = _state.pos(1) - _hpos_pred_prev(1) - ev_delta_pos(1);
+					_ev_vel_pos_innov[3] = _state.pos(0) - _hpos_pred_prev(0) - ev_delta_pos(0);
+					_ev_vel_pos_innov[4] = _state.pos(1) - _hpos_pred_prev(1) - ev_delta_pos(1);
 
 					// observation 1-STD error, incremental pos observation is expected to have more uncertainty
-					_posObsNoiseNE = fmaxf(_ev_sample_delayed.posErr, 0.5f);
+					ev_obs_var[3] = ev_obs_var[4] = fmaxf(_ev_sample_delayed.posErr, 0.5f);
 				}
 
 				// record observation and estimate for use next time
@@ -334,10 +323,10 @@ void Ekf::controlExternalVisionFusion()
 				if (_params.fusion_mode & MASK_ROTATE_EV) {
 					ev_pos_meas = _ev_rot_mat * ev_pos_meas;
 				}
-				_vel_pos_innov[3] = _state.pos(0) - ev_pos_meas(0);
-				_vel_pos_innov[4] = _state.pos(1) - ev_pos_meas(1);
+				_ev_vel_pos_innov[3] = _state.pos(0) - ev_pos_meas(0);
+				_ev_vel_pos_innov[4] = _state.pos(1) - ev_pos_meas(1);
 				// observation 1-STD error
-				_posObsNoiseNE = fmaxf(_ev_sample_delayed.posErr, 0.01f);
+				ev_obs_var[3] = ev_obs_var[4] = fmaxf(_ev_sample_delayed.posErr, 0.01f);
 
 				// check if we have been deadreckoning too long
 				if ((_time_last_imu - _time_last_pos_fuse) > _params.reset_timeout_max) {
@@ -351,16 +340,12 @@ void Ekf::controlExternalVisionFusion()
 			}
 
 			// innovation gate size
-			_posInnovGateNE = fmaxf(_params.ev_pos_innov_gate, 1.0f);
-		}else{
-			_vel_pos_innov[3] = 0.0f;
-			_vel_pos_innov[4] = 0.0f;
+			ev_innov_gate[HPOS] = fmaxf(_params.ev_pos_innov_gate, 1.0f);
 		}
 
 		// determine if we should use the velocity observations
 		if (_control_status.flags.ev_vel) {
-			_fuse_hor_vel = true;
-			_fuse_vert_vel = true;
+			ev_fuse_mask[HVEL] = ev_fuse_mask[VVEL] = true;
 
 			Vector3f vel_aligned{_ev_sample_delayed.vel};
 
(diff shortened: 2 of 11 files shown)


EKF/ekf_helper.cpp
```diff
@@ -1196,18 +1196,22 @@ bool Ekf::reset_imu_bias()
...
+	*gps_vel = sqrtf(math::max(math::max(_gps_vel_pos_test_ratio[0], _gps_vel_pos_test_ratio[1]), _gps_vel_pos_test_ratio[2]));
+	// return the largest NE GPS position innovation test ratio
+	*gps_pos = sqrtf(math::max(_gps_vel_pos_test_ratio[3], _gps_vel_pos_test_ratio[4]));
+	// return the largest external vision velocity innovation test ratio
+	*ev_vel = sqrtf(math::max(math::max(_ev_vel_pos_test_ratio[0], _ev_vel_pos_test_ratio[1]), _ev_vel_pos_test_ratio[2]));
+	// return the largest horizontal external vision position innovation test ratio
+	*ev_pos = sqrtf(math::max(_ev_vel_pos_test_ratio[3], _ev_vel_pos_test_ratio[4]));
 	// return the vertical position innovation test ratio
-	*hgt = sqrtf(_vel_pos_test_ratio[5]);
+	*hgt = sqrtf(_gps_vel_pos_test_ratio[5])+sqrtf(_ev_vel_pos_test_ratio[5]);
```
priseborough: I think we should return the maximum.

EKF/vel_pos_fusion.cpp
```diff
@@ -45,119 +45,136 @@
 #include <ecl.h>
 #include <mathlib/mathlib.h>
 
-void Ekf::fuseVelPosHeight()
+/**
+ * Update the EKF state with velocity and position measurements.
+ *
+ * @param innov  	Input/Output	Innovation which is fuse_mask is set to false or do not pass
```
priseborough: Can you please reword the description for this. I have read it a few times and am not sure what it means.

kamilritz: You are totally right, this sentence does not make any sense. Thanks for noticing.

kamilritz: I think this line was part of an outdated version. Please have a look at the newest version.

EKF/ekf_helper.cpp
kamilritz: I am still not sure what would be the appropriate solution. Can you give me some reasoning why we should use the maximum?

kamilritz:
I seem to break the python tests. But I do not manage to build ECL locally as a standalone. So I can not run and check the python tests. Any advise here? @priseborough

EKF/ekf_helper.cpp
```diff
@@ -1217,8 +1363,8 @@ void Ekf::get_ekf_soln_status(uint16_t *status)
 	soln_status.flags.const_pos_mode = !soln_status.flags.velocity_horiz;
 	soln_status.flags.pred_pos_horiz_rel = soln_status.flags.pos_horiz_rel;
 	soln_status.flags.pred_pos_horiz_abs = soln_status.flags.pos_horiz_abs;
-	bool gps_vel_innov_bad = (_vel_pos_test_ratio[0] > 1.0f) || (_vel_pos_test_ratio[1] > 1.0f);
-	bool gps_pos_innov_bad = (_vel_pos_test_ratio[3] > 1.0f) || (_vel_pos_test_ratio[4] > 1.0f);
+	bool gps_vel_innov_bad = (_gps_vel_pos_test_ratio[0] > 1.0f);
```
bresch: why only checking the X axis here?

EKF/ekf_helper.cpp
```diff
@@ -1182,22 +1324,26 @@ bool Ekf::reset_imu_bias()
 // Innovation Test Ratios - these are the ratio of the innovation to the acceptance threshold.
 // A value > 1 indicates that the sensor measurement has exceeded the maximum acceptable level and has been rejected by the EKF
 // Where a measurement type is a vector quantity, eg magnetometer, GPS position, etc, the maximum value is returned.
-void Ekf::get_innovation_test_status(uint16_t *status, float *mag, float *vel, float *pos, float *hgt, float *tas, float *hagl, float *beta)
+void Ekf::get_innovation_test_status(uint16_t *status, float *mag, float *gps_vel, float *gps_pos, float *ev_vel, float *ev_pos, float *hgt, float *tas, float *hagl, float *beta)
```
bresch: pass references?

EKF/ekf.h
```diff
@@ -57,58 +57,95 @@ class Ekf : public EstimatorInterface
 	// should be called every time new data is pushed into the filter
 	bool update();
 
-	// gets the innovations of velocity and position measurements
-	// 0-2 vel, 3-5 pos
-	void get_vel_pos_innov(float vel_pos_innov[6]);
+	// gets the GPS innovations of velocity and position measurements
```
bresch: All those comments are redundant as the names of the functions are clear. However, if the intent is to generate documentation, you would need to add more details

EKF/control.cpp
```diff
@@ -1172,9 +1180,73 @@ void Ekf::controlHeightFusion()
 
 		}
 
-		_fuse_height = true;
+		height_fuse_mask[3] = true;
 	}
 
+	if (height_fuse_mask[3]) {
```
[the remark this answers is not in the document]
kamilritz: @bresch  No, I do not want to. I am not very happy how the height fusion is implemented at the moment and I will try to work on this next. If you want I can split it into two separate functions for now?

EKF/tests/pytest/test_utils.py
```diff
@@ -89,7 +89,7 @@ def initialized_ekf():
     dt_usec = 5000
 
     # Provide a few samples
-    for _ in range(1000):
+    for _ in range(2000):
```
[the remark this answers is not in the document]
kamilritz: @bresch Apparently. I can not see a clear reason why it should take longer to get the tilt aligned. But with out increasing the number of samples, it does not pass the test.

EKF/vel_pos_fusion.cpp
```diff
@@ -45,276 +45,190 @@
...
+			healthy = false;
 
-		for (int i = 0; i < _k_num_states; i++) {
-			if (P[i][i] < KHP[i][i]) {
-				// zero rows and columns
-				zeroRows(P, i, i);
-				zeroCols(P, i, i);
+			// update individual measurement health status
+			if (obs_index == 0) {
+				_fault_status.flags.bad_vel_N = true;
```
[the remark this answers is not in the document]
kamilritz: That is nice!

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
 	// sensor measurement
 	float calculate_synthetic_mag_z_measurement(Vector3f mag_meas, Vector3f mag_earth_predicted);
 
+	// stop GPS fusion
```
bresch:
```suggestion
```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopGpsFusion();
+
+	// stop GPS position fusion
```
bresch:
```suggestion
```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopGpsPosFusion();
+
+	// stop GPS velocity fusion
```
bresch:
```suggestion

```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopGpsVelFusion();
+
+	// stop GPS yaw fusion
```
bresch:
```suggestion

```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopGpsYawFusion();
+
+	// stop external vision fusion
```
bresch:
```suggestion

```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopEvFusion();
+
+	// stop external vision position fusion
```
bresch:
```suggestion

```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopEvPosFusion();
+
+	// stop external vision velocity fusion
```
bresch:
```suggestion

```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopEvVelFusion();
+
+	// stop external vision yaw fusion
```
bresch:
```suggestion

```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopEvYawFusion();
+
+	// stop auxiliary horizontal velocity fusion
```
bresch:
```suggestion

```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopAuxVelFusion();
+
+	// stop optical flow fusion
```
bresch:
```suggestion

```

EKF/ekf.h
```diff
@@ -724,4 +739,37 @@ class Ekf : public EstimatorInterface
...
+	void stopFlowFusion();
+
+	// Helper function to set velPos fault status
```
bresch:
```suggestion

```

EKF/estimator_interface.h
```diff
@@ -61,40 +61,84 @@ class EstimatorInterface
 	virtual bool init(uint64_t timestamp) = 0;
 	virtual bool update() = 0;
 
-	// gets the innovations of velocity and position measurements
-	// 0-2 vel, 3-5 pos
-	virtual void get_vel_pos_innov(float vel_pos_innov[6]) = 0;
+	// gets the GPS innovations of velocity and position measurements
```
bresch:
```suggestion

```

EKF/estimator_interface.h
```diff
@@ -61,40 +61,84 @@ class EstimatorInterface
...
+	virtual void getGpsVelPosInnov(float hvel[2], float &vvel, float hpos[2], float &vpos) = 0;
+	// gets the GPS innovation variances of velocity and position measurements
```
bresch:
```suggestion

```

EKF/tests/pytest/test_utils.py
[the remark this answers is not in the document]
bresch: The problem is that we need to know why. Maybe it worked before but shouldn't pass, maybe something broke now...

EKF/vel_pos_fusion.cpp
```diff
@@ -45,276 +45,175 @@
...
-			Kfusion[row] = P[row][state_index] / _vel_pos_innov_var[obs_index];
+	for (int i = 0; i < _k_num_states; i++) {
+		if (P[i][i] < KHP[i][i]) {
+			// zero rows and columns
+			zeroRows(P, i, i);
+			zeroCols(P, i, i);
+
+			healthy = false;
+
+			setVelPosFaultStatus(obs_index,true);
```
bresch: 1. It's okay here because it is clear enough that the true/false is the value you want to apply to the obs_index, but good practice is to wrap the function into an other one that sets the boolean for you.
2. I know that there is no astyle check for ecl, but try to follow the rues as close as possible :D 
```suggestion
			setVelPosFaultStatus(obs_index, true);
```

EKF/vel_pos_fusion.cpp
```diff
@@ -45,276 +45,175 @@
...
+
+		} else {
+			setVelPosFaultStatus(obs_index,false);
```
bresch:
```suggestion
			setVelPosFaultStatus(obs_index, false);
```

bresch:
Just as a side note, adding a comment to every variable and function isn't useful if it doesn't add any information: https://www.todaysoftmag.com/article/1120/clean-code-comments-and-formatting

EKF/ekf.h
```diff
@@ -57,58 +57,65 @@ class Ekf : public EstimatorInterface
...
+
+	void getDragInnov(float drag_innov[2]);
+
+	void getDragInnovVar(float drag_innov_var[2]);
+
+	void getDragInnovRatio(float drag_innov_ratio[2]);
+
+	void getAirspeedInnov(float &airspeed_innov);
+
+	void getAirspeedInnovVar(float &airspeed_innov_var);
```
dagar: If you're going to bother changing these how about we return by value whenever possible? A lot of them could then turn into inlined getters.

EKF/ekf.h
```diff
@@ -57,58 +57,65 @@ class Ekf : public EstimatorInterface
...
+
+	void getMagInnovRatio(float &mag_innov_ratio);
+
+	void getDragInnov(float drag_innov[2]);
+
+	void getDragInnovVar(float drag_innov_var[2]);
+
+	void getDragInnovRatio(float drag_innov_ratio[2]);
+
+	void getAirspeedInnov(float &airspeed_innov);
```
dagar:
```suggestion
	float getAirspeedInnov() { return _airspeed_innov; }
```

EKF/ekf_helper.cpp
```diff
@@ -852,72 +861,172 @@ void Ekf::calcEarthRateNED(Vector3f &omega, float lat_rad) const
...
+{
+	memcpy(hvel, _gps_vel_pos_innov_var+0, sizeof(float) * 2);
+	memcpy(&vvel, _gps_vel_pos_innov_var+2, sizeof(float) * 1);
+	memcpy(hpos, _gps_vel_pos_innov_var+3, sizeof(float) * 2);
+	memcpy(&vpos, _gps_vel_pos_innov_var+5, sizeof(float) * 1);
+}
+
+void Ekf::getGpsVelPosInnovRatio(float &hvel, float &vvel, float &hpos, float &vpos)
+{
+	memcpy(&hvel, _gps_vel_pos_test_ratio+HVEL, sizeof(float));
```
jkflying: These ones with the references are much easer to read as:
```suggestion
	hvel = _gps_vel_pos_test_ratio[HVEL];
```

EKF/ekf.h
dagar:
```suggestion
	float getAirspeedInnovVar() const { return _airspeed_innov_var; }
```

EKF/tests/pytest/test_utils.py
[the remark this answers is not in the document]
jkflying: This points to a change in behavior, in what should be a zero-behavior-change refactor. This should be investigated, maybe it was fixed due to a bug before, but I think due to it taking longer to diverge there is more likely a bug now.

EKF/vel_pos_fusion.cpp
```diff
@@ -45,276 +45,175 @@
 #include <ecl.h>
 #include <mathlib/mathlib.h>
 
-void Ekf::fuseVelPosHeight()
+/**
```
jkflying: This whole file is much cleaner, nice work!

EKF/ekf_helper.cpp
```diff
@@ -852,72 +861,172 @@ void Ekf::calcEarthRateNED(Vector3f &omega, float lat_rad) const
 	omega(2) = -CONSTANTS_EARTH_SPIN_RATE * sinf(lat_rad);
 }
 
-// gets the innovations of velocity and position measurements
-// 0-2 vel, 3-5 pos
-void Ekf::get_vel_pos_innov(float vel_pos_innov[6])
+void Ekf::getGpsVelPosInnov(float hvel[2], float &vvel, float hpos[2],  float &vpos)
+{
+	memcpy(hvel, _gps_vel_pos_innov+0, sizeof(float) * 2);
```
dagar: Returning matrix::Vectors is also fine.

kamilritz: can you give an example how you mean that?

dagar: Actually this one isn't a great example, but in other cases we could easily/safely return a Vector by value rather than passing in an array or pointer to fill.

Examples

``` C++
    // Current PX4/ecl:
    // we often pass in a pointer to fill with 1, 2, or 3 floats
    void estimator_interface::get_velocity(float *vel);
    // wind velocity pass in vector to fill with 2 floats
    void Ekf::get_wind_velocity(float *wind);


    // SUGGESTION:
    // whenever possible we should prefer simple and safe getters that return by value
    matrix::Vector3f estimator_interface::get_velocity() const;
    matrix::Vector2f Ekf::get_wind_velocity() const;

```

I'm mainly trying to say don't be afraid to return these thing as values. It's easier to use, harder to misuse, and not going to be any slower.

EKF/ekf.h
kamilritz: That is a nice idea. A lot of the float arrays that are returned are directly put into message fields, which are also a float array. So returning them by value as a Vector would be a lot of effort. 
There are a lot of other functions that do not follow the requested standard and I suggest that we try to clean this up in a separate PR.

EKF/ekf_helper.cpp
kamilritz: This is resolved in the newest version.

kamilritz:
> All good, except the python check that fails

@bresch I found out why I did not pass the original python tiltAlignment test. In the current EKF version we fuse fakePos measurements in the beginning since we do not have GPS measurements. The rate at which this is happening is determined by this [if statement](https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L1664). Since we fuse baro measurement early on, we set the variable `_fuse_height` to true at the rate of the baro measurements. In my previous version I did not use the `_fuse_height` variable anymore, but it was still in the if statement. There for the fake zero x and y position measurement were fused at a much lower rate causing the uncertainty in the orientation to converge slower. As an intermediate solution I set the `_fuse_height` variable to true every time height is fused. I checked with a debugger, that now the time of the successful tilt alignment during the test is unchanged. 

PS: the python tests based on the `initialized_ekf()` function are probably flawed, since they push in IMU sample with `delta_*_dt` in microseconds instead of seconds.

@priseborough LGTM

EKF/common.h
```diff
@@ -400,18 +400,19 @@ union fault_status_u {
...
+		bool reject_mag_x: 1;		///< 4 - true if the X magnetometer observation has been rejected
+		bool reject_mag_y: 1;		///< 5 - true if the Y magnetometer observation has been rejected
+		bool reject_mag_z: 1;		///< 6 - true if the Z magnetometer observation has been rejected
+		bool reject_yaw: 1;		///< 7 - true if the yaw observation has been rejected
+		bool reject_airspeed: 1;	///< 8 - true if the airspeed observation has been rejected
+		bool reject_sideslip: 1;	///< 9 - true if the synthetic sideslip observation has been rejected
+		bool reject_hagl: 1;		///< 10 - true if the height above ground observation has been rejected
+		bool reject_optflow_X: 1;	///< 11 - true if the X optical flow observation has been rejected
+		bool reject_optflow_Y: 1;	///< 12 - true if the Y optical flow observation has been rejected
 	} flags;
```
mhkabir: Why is this not per-sensor now?

EKF/ekf_helper.cpp
```diff
@@ -1731,3 +1858,75 @@ float Ekf::kahanSummation(float sum_previous, float input, float &accumulator) c
...
+	_control_status.flags.gps = false;
+	_control_status.flags.gps_hgt = false;
+	_gps_pos_innov.setZero();
+	_gps_pos_innov_var.setZero();
+	_gps_pos_test_ratio.setZero();
+}
+
+void Ekf::stopGpsVelFusion()
+{
+	_gps_vel_innov.setZero();
```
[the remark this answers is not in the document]
kamilritz: I agree. The control_status is logged in the estimator_status.msg. Where would we put the the GPS velocity flag in the filter_control_status struct? At the end? Are people okay with changed meaning of the control_status.gps flag?

EKF/ekf_helper.cpp
```diff
@@ -54,13 +54,15 @@ bool Ekf::resetVelocity()
 
 	// reset EKF states
 	if (_control_status.flags.gps && _gps_check_fail_status.value==0) {
+		ECL_INFO_TIMESTAMPED("EKF reset velocity to GPS");
```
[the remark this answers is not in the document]
kamilritz: That is true. I will remove them.

EKF/common.h
kamilritz: That is a good question. That is a point that is open for discussion.

EKF/common.h
```diff
@@ -400,18 +400,19 @@ union fault_status_u {
...
-		bool reject_mag_x: 1;		///< 3 - true if the X magnetometer observation has been rejected
-		bool reject_mag_y: 1;		///< 4 - true if the Y magnetometer observation has been rejected
-		bool reject_mag_z: 1;		///< 5 - true if the Z magnetometer observation has been rejected
-		bool reject_yaw: 1;		///< 6 - true if the yaw observation has been rejected
-		bool reject_airspeed: 1;	///< 7 - true if the airspeed observation has been rejected
-		bool reject_sideslip: 1;	///< 8 - true if the synthetic sideslip observation has been rejected
-		bool reject_hagl: 1;		///< 9 - true if the height above ground observation has been rejected
-		bool reject_optflow_X: 1;	///< 10 - true if the X optical flow observation has been rejected
-		bool reject_optflow_Y: 1;	///< 11 - true if the Y optical flow observation has been rejected
+		bool reject_hor_vel: 1;		///< 0 - true if horizontal velocity observations have been rejected
```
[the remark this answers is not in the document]
kamilritz: I like this idea very much. The current bitmask approach  also means that reordering the flags is not advisable.
I can help to implement that, but I think that this should go into a separate PR.

EKF/ekf_helper.cpp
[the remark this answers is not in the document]
mhkabir: I think both of those should be ok, but I defer to others.

kamilritz:
@mhkabir and @dagar thank you very much for your feedback. Your requests are in the process of being implemented and will be in new PR.

bresch:
@kamilritz Do you think you can start again on this topic? This would be a great improvement I think

kamilritz:
> @kamilritz Do you think you can start again on this topic? This would be a great improvement I think

@bresch Good point. I will put this on my list. But it will take some time until I will get to it. If you need it soon, then feel free to do it yourself.

bresch:
@kamilritz Okay, cool. It isn't something urgent but it would be great to have it for the next release (v1.12)

PR #654, opened 2019-10-11
Edit true airspeed setpoint 

moreba1:
https://github.com/PX4/Firmware/issues/13090


CarlOlsson:
Hi!
Why do you want to remove the conversion from equivalent airspeed to true airspeed in TECS?

moreba1:
Airspeed setpoint by user is true airspeed and don't need to coversion.

CarlOlsson:
Please explain more in depth your proposed change.
Why do you think the user should command TAS instead of EAS? With this change the aircraft will stall at minimum airspeed at high altitudes.

priseborough:
This is an unsafe change as indicatged by @CarlOlsson

PR #655, opened 2019-10-14
Enhancement: Increased robustness against 0 signal quality and cleanup range data handling

RomanBapst:
This PR attempts to cleanup the checks on the range finder data. Additionally it adds a hysteresis to declaring range finder data valid after signal quality was zero.

Specifically:
- whenever signal quality of the range data was 0 we require non-zero signal quality for 3 seconds before we accept the data to be used. This is to improve robustness against a known sensor failure mode where the signal quality toggles between 0 and non-zero values (see plot below)
- moved range finder data checks to a central location and created new file range_finder_checks.cpp in order to avoid loading control.cpp
- replaced _rng_hgt_faulty with _range_hgt_valid in order to avoid negation all over the place
- use better function name
- moved updateRangeDataContinuity() out of the terrain estimator and into range_finder_checks.cpp

I will provide SITL test data shortly.

Plot of range finder failure mode:

image

Notice how signal quality erratically jumps between 0 and 95% in the middle of the flight. The height reported by the sensor at that point was completely wrong.

### DIFF
--- a/EKF/range_finder_checks.cpp
+++ b/EKF/range_finder_checks.cpp
@@ -1,0 +1,144 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2019 Estimation and Control Library (ECL). All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ * 3. Neither the name ECL nor the names of its contributors may be
+ *    used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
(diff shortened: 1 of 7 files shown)


LorenzMeier:
The 3 seconds hold-off sounds like it is engineered for a specific sensor - can this be made a parameter and be set for that sensor?

RomanBapst:
Plot demonstrating 3 second hysteresis after signal quality was 0:
image

Log: https://mail.google.com/mail/u/0/#inbox/FMfcgxwDrlVmwVjrGGZSjKnTPrbflWXr

Plot demonstrating a height reset to range finder (simulated a step in range finder data of 5m):
image

Log: https://logs.px4.io/plot_app?log=d3182c9d-01e2-4d15-940f-00b7d2190af2

EKF/range_finder_checks.cpp
```diff
@@ -0,0 +1,144 @@
...
+	// Don't run the checks after this unless we have retrieved new data from the buffer
+	if (!_range_data_ready) {
+		return;
+	}
+
+	if (_range_sample_delayed.quality == 0) {
+		_time_bad_rng_signal_quality = _imu_sample_delayed.time_us;
+		_rng_hgt_valid = false;
+	} else if (_time_bad_rng_signal_quality > 0 && _imu_sample_delayed.time_us - _time_bad_rng_signal_quality > RNG_BAD_SIG_HYST) {
+		_time_bad_rng_signal_quality = 0;
```
bresch:
```suggestion
```

EKF/range_finder_checks.cpp
```diff
@@ -0,0 +1,146 @@
...
+					(_imu_sample_delayed.time_us - _range_sample_delayed.time_us);
+
+	_dt_last_range_update_filt_us = fminf(_dt_last_range_update_filt_us, 4e6f);
+
+	if (_dt_last_range_update_filt_us < 2e6f) {
+		_range_data_continuous = true;
+
+	} else {
+		_range_data_continuous = false;
+	}
```
bresch: `_range_data_continuous` doesn't need to be stored as it is only checked once and the result is straigntforward
```suggestion
	bool EKF::isRangeDataContinuous()
	{
	   return _dt_last_range_update_filt_us < 2e6f
	}
```

EKF/range_finder_checks.cpp
```diff
@@ -0,0 +1,146 @@
...
+	updateRangeDataContinuity();
+
+	// check if out of date
+	if ((_imu_sample_delayed.time_us - _range_sample_delayed .time_us) > 2 * RNG_MAX_INTERVAL) {
+		_rng_hgt_valid = false;
+		return;
+	}
+
+	// Don't allow faulty flag to clear unless range data is continuous
+	if (!_rng_hgt_valid && !_range_data_continuous) {
```
bresch:
```suggestion
	if (!_rng_hgt_valid && !isRangeDataContinuous()) {
```

RomanBapst:
@MaEtUgR I addressed your comments, they were good suggestions. Please have another look.

RomanBapst:
@bresch Sorry I meant you, not @MaEtUgR

RomanBapst:
@priseborough Could you please have a look as well?

priseborough:
Thanks @RomanBapst

PR #656, opened 2019-10-14
control: refactor rangeAidConditionsMet function

bresch:
Just because it took me way too long to read this function.
After this refactor, the comment at the top of the function is almost redundant, should we remove it?


RomanBapst:
@bresch Looks like a nice cleanup! Maybe you can rename "rangeAidConditionsMet" to "updateRangeAidEnabled" or "updateRangeAid" or something similar. The current name somehow makes me think that his function should return a boolean.

RomanBapst:
@bresch We need a test log where you can see the estimator switching in and out of range aid.
I usually use gazebo and add a a few boxes. Then I fly over them and verify that the drones does not change altitude.

bresch:
@RomanBapst I changed the names and I did a sitl test with boxes:

https://logs.px4.io/plot_app?log=4bc72eb4-4d74-4aa2-a068-cae8828313e7
It switches properly in and out of rang aid
2019-10-14_15-26-12_01_plot

bresch:
@RomanBapst Can you review again please?

PR #661, opened 2019-10-29
added python script for derivation of initial wind covariance matrix

RomanBapst:
This script derives the initial convariance matrix of the earth horizontal wind vector.

@CarlOlsson I was working on this to improve the initialization of the wind_estimation library we are using for airspeed fault detection. I noticed that you did something similar in #615 
@priseborough commented about the option to explicitly include the sideslip angle.
I think the derivation I did here could go into that direction.
When I set the yaw angle to 0 and both yaw and sideslip angle uncertainties to 0 I have 0 initial uncertainty in east direction. As soon as I have non-zero value for either yaw or sideslip uncertainty I see the contribution in east direction.

I've also tested the option to include the uncertainty of the ground speed in the derivation. However, using common variances for vx and vy from ekf2 I noticed that the contributions from those were negligible and I removed them.

Let me know what you think. 

### DIFF
--- a/EKF/python/wind_cov_init/derivation.py
+++ b/EKF/python/wind_cov_init/derivation.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+"""
+Created on Tue Oct 29 14:11:58 2019
+
+@author: roman
+"""
+
+from sympy import *
+
+################## Here are the variables you can change to see the effects on the cov matrix ###########################
+yaw_init =  0.5
+
+# ground speed in body frame (comes from ekf2)
+groundspeed_body_x_init = 5
+groundspeed_body_y_init = 5
+
+# true airspeed measured by pitot tube
+V_init = 7
+
+# heading variance
+R_yaw_init = rad(15.0)**2
+
+# sideslip variance
+R_beta_init = rad(15.0)**2
+
+# True airspeed measurement variance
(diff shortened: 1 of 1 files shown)


dagar:
A bit of a side note, but going forward could we start thinking about structuring any code generation so it's a bit more accessible to others that might want to work in this area?
1. Have generated code separated (different files) from hand written code and if possible actually do the generation at build time. This would make the core code easier to browse/understand and we could start automatically formatting ecl.
2. Include instructions or a custom target in cmake that actually runs the generation.

RomanBapst:
@dagar I would like that. I told you before I have written the ekf derivation in python which we could use at some point in the future if such an architecture is in place.

dagar:
> > Have generated code separated (different files) from hand written code and if possible actually do the generation at build time.
> 
> Sounds a bit scary to me to generate the code every time at build. How can we be sure that it gives the same results on all systems etc?

We wouldn't do it immediately, but I don't have it that scary (we already generate stuff allover the place). The incremental step would be to make it trivial to regenerate in place, then over time we'd see if it's necessary to lock down the sympy version.

priseborough:
Auto generation of algebra is scary because a change in toolchain version could introduce a bug that results in complete loss of estimation/attitude reference. This is more likely to occur IMO than a compiler error because the user base for the sympi is much smaller than most compilers such as gcc. Significant reversion testing would be required if there was  change in toolchain version.

PR #663, opened 2019-10-30
EKF increase delta velocity variance while sensor is clipping

dagar:
Mainly hacks for testing at this point.

Questions:
 ~~1. Is it worth handling this per axis? Clipping in Z is the most common.~~

Goes with ~~https://github.com/PX4/Firmware/pull/13103~~ https://github.com/PX4/Firmware/pull/14304.

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -246,6 +246,20 @@ void Ekf::predictCovariance()
 
 	dvxVar = dvyVar = dvzVar = sq(dt * accel_noise);
 
+	// Accelerometer Clipping
+	// delta velocity X: increase process noise if sample contained any X axis clipping
+	if (_imu_sample_delayed.delta_vel_clipping[0]) {
+		dvxVar = sq(dt * BADACC_BIAS_PNOISE);
+	}
+	// delta velocity Y: increase process noise if sample contained any Y axis clipping
+	if (_imu_sample_delayed.delta_vel_clipping[1]) {
+		dvyVar = sq(dt * BADACC_BIAS_PNOISE);
+	}
+	// delta velocity Z: increase process noise if sample contained any Z axis clipping
+	if (_imu_sample_delayed.delta_vel_clipping[2]) {
+		dvzVar = sq(dt * BADACC_BIAS_PNOISE);
+	}
+
 	// predict the covariance
 
 	// intermediate calculations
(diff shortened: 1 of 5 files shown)


priseborough:
What i would like to do here is to monitor the clipping count increase across each EKF state and covariance prediction  time step and scale the per axis sensor noise proportional to the clipping count increase on a per axis basis with the amount of increase determined from replay analysis .

We could add the clipping count to the IMU data uORB topic.

priseborough:
Alternatively we could publish the fraction of IMU samples affected in a given downsample IMU data message

priseborough:
So for example if the baseline noise is 0.3 and the clipping noise factor is 2.0, then the observation noise used would be 0.3 + 2.0 * clipping_fraction

dagar:
Tracking clipping per axis breaks down with any non-orthogonal sensor or board rotation or board level. In PX4/Firmware the new sensor status messages will still track the count per axis, but it will just be a total clip count passed through to the estimator.

EKF/covariance.cpp
```diff
@@ -260,18 +260,25 @@ void Ekf::predictCovariance()
...
-	float dvxVar, dvyVar, dvzVar;
-	float gyro_noise = math::constrain(_params.gyro_noise, 0.0f, 1.0f);
 	daxVar = dayVar = dazVar = sq(dt * gyro_noise);
+
 	float accel_noise = math::constrain(_params.accel_noise, 0.0f, 1.0f);
+	if (_imu_sample_delayed.delta_vel_clip_count > 0 && _imu_sample_delayed.delta_vel_samples > 0) {
+		// total clip count across every axis
+		const float clipping_fraction = _imu_sample_delayed.delta_vel_clip_count / (_imu_sample_delayed.delta_vel_samples * 3);
+
+		accel_noise += 2.0f * clipping_fraction;
```
priseborough: Let's make this a tuning parameter so we can use replay to iterate on a suitable value.

dagar: I've added _params->accel_noise_clipping_factor, which is EKF2_ACC_CLPFACT PX4 side.

dagar:
PX4/Firmware PR for testing on a pixracer or pixhawk 3 pro (icm20602 or icm20608g - https://github.com/PX4/Firmware/pull/13837

bresch:
Good log from boot with a lot of clipping: https://logs.px4.io/plot_app?log=c795c4d4-c17c-403a-9bb4-0098d14375c9
clipping

dagar:
This was tested in https://github.com/PX4/Firmware/pull/14304#issuecomment-609939435 and seems to be effective, although the flight log has too many dropouts to replay.

PR #664, opened 2019-11-01
Add access to EKF state reset

bresch:
I'm creating a logic in PX4/Firmware that checks for motion an resets the states or completely the EKF. To do so, I needed to increase the granularity of the init function.
I also need to know if the vehicle is at rest or not so I created a getter to reuse the "at_rest" variable of EKF2.

Required for https://github.com/PX4/Firmware/pull/13346

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -47,6 +47,13 @@
 bool Ekf::init(uint64_t timestamp)
 {
 	bool ret = initialise_interface(timestamp);
+	reset(timestamp);
+
+	return ret;
+}
+
+void Ekf::reset(uint64_t timestamp)
+{
 	_state.vel.setZero();
 	_state.pos.setZero();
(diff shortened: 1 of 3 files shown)


EKF/ekf.cpp
```diff
@@ -47,20 +47,14 @@
...
-	_state.vel.setZero();
-	_state.pos.setZero();
-	_state.gyro_bias.setZero();
-	_state.accel_bias.setZero();
-	_state.mag_I.setZero();
-	_state.mag_B.setZero();
-	_state.wind_vel.setZero();
-	_state.quat_nominal.setZero();
-	_state.quat_nominal(0) = 1.0f;
+	reset(timestamp);
```
kamilritz: A comment here could be helpful

bresch: It's basically THE ekf reset function, should I call it `resetAll` ?

PR #666, opened 2019-11-03
Jenkins: update container tags; use px4-dev-base-bionic

TSC21:
Purpose: have the containers updated with the current tags. Also, using the `px4-dev-ecl` container doesn't make sense anymore, as it just brought Python dependencies as an extra (https://github.com/PX4/containers/pull/208 removes that container).


TSC21:
@dagar feel free to merge as I don't have write permissions for this repo. Thanks!

PR #668, opened 2019-11-11
[continue:] Moving towards multi-sensor-per-same-measurement fusion

kamilritz:
This PR is exactly the same as: https://github.com/PX4/ecl/pull/652,
but added as a separate branch to ECL. Such that we can test the PR's compatibility with the Firmware's CI.

### DIFF
--- a/EKF/vel_pos_fusion.cpp
+++ b/EKF/vel_pos_fusion.cpp
@@ -45,276 +45,175 @@
 #include <ecl.h>
 #include <mathlib/mathlib.h>
 
-void Ekf::fuseVelPosHeight()
+/**
+ * Update the EKF state with velocity and position measurements sequentially. [(vx vy) (vz) (x y) (z)]
+ *
+ * @param innov  	Input		[vx vy vz x y z]
+ * @param innov_gate	Input		[vxy vz xy z]
+ * @param obs_var	Input		[vx vy vz x y z]
+ * @param fuse_mask	Input/Output	[vxy vz xy z]
+ * 					Specify which innovation components should be fused,
+ * 					components that do not pass innovations checks will be set to zero
+ * @param innov_var	Ouput		[vx vy vz x y z]
+ * @param test_ratio	Output		[vxy vz xy z]
+ */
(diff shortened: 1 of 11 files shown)


mhkabir:
Question: why do we even use the std. deviation in the estimator interface?
1. PX4 side uses variances (from cov matx)
2. Fusion uses variances

So we just end up uselessly square rooting and then squaring.

EKF/vel_pos_fusion.cpp
```diff
@@ -45,276 +45,174 @@
...
+		_time_last_ver_vel_fuse = _time_last_imu;
+		_innov_check_fail_status.flags.reject_ver_vel = false;
 
-		// Set observation noise variance and innovation consistency check gate size for the NE position observations
-		R[4] = R[3] = sq(_posObsNoiseNE);
-		gate_size[4] = gate_size[3] = _posInnovGateNE;
+		fuseVelPosHeight(innov(2),innov_var(2),2);
 
+		return true;
+	}else{
```
mhkabir:
```suggestion
	} else {
```

kamilritz:
@mhkabir I also noticed that and want get rid of the std deviation. I am in the process of cleaning this and some other things up. But these changes are based on top of this refactor. I am eager to get them in after this PR.

EKF/vel_pos_fusion.cpp
kamilritz: That is only the tip of the iceberg. At some point we really need a formatter.

mhkabir:
@kamilritz What is missing here? Let's bring this in soon.

kamilritz:
@mhkabir Glad you feel the same. Waiting on the firmware PR to be ready.

xdwgood:
@kamilritz  https://github.com/PX4/ecl/blob/master/EKF/vel_pos_fusion.cpp#L168

This is just `KP`, why does the code say it is `KHP`, where does `H` come from?  thank you very much :smiley:

kamilritz:
> Is it because the H matrix is 1?

Yes. To be a bit more precise. It is a vector of size 1x24 full of zeros with a single 1 at the location of the measurement.

PR #671, opened 2019-11-22
TECS: remove height rate complementary filter

dagar:
https://github.com/PX4/Firmware/pull/13527#issuecomment-557354415


priseborough:
Will need to be tested at PX4/Firmware level

kamilritz:
This PR changes the interface. With it going into EC> master, the current ECL master is not compatible to the Firmware at the moment.

PR #674, opened 2019-11-26
EKF: use Matrix cross product

dagar:


priseborough:
How have you checked that the order of the cross product hasn't changed due to this request? Because the operator is used for sensor  offset compensation, I don't think a order reversal would show up as a CI test failure>

jkflying:
ECL: (removed in this PR)
```C++
Vector3f EstimatorInterface::cross_product(const Vector3f &vecIn1, const Vector3f &vecIn2)
{
	Vector3f vecOut;
	vecOut(0) = vecIn1(1) * vecIn2(2) - vecIn1(2) * vecIn2(1);
	vecOut(1) = vecIn1(2) * vecIn2(0) - vecIn1(0) * vecIn2(2);
	vecOut(2) = vecIn1(0) * vecIn2(1) - vecIn1(1) * vecIn2(0);
	return vecOut;
}
```
Matrix:
```C++
    Vector3 cross(const Matrix31 & b) const {
        const Vector3 &a(*this);
        return {a(1)*b(2,0) - a(2)*b(1,0), -a(0)*b(2,0) + a(2)*b(0,0), a(0)*b(1,0) - a(1)*b(0,0)};
    }
```

Looks the same to me.

Issue #675, opened 2019-12-05
makeSymmetrical function


kamilritz:
The helper function [makeSymmetrical](https://github.com/PX4/ecl/blob/master/EKF/ekf_helper.cpp#L781) that makes part of the covariance matrix symmetric only considers element up to the last index. It therefore does not consider covariance entries of the states that come after the last index.

E.g. Matrix is 5 by 5. first = 2, last = 3
This affect following entries:

o x x o o
x o x o o
x x o o o
o o o o o
o o o o o

IMO it should affect: 

o x x o o
x o x x x
x x o x x
o x x o o
o x x o o

@priseborough 
@RomanBapst

PR #676, opened 2019-12-06
Make covariance matrices of type matrix::SquareMatrix

kamilritz:
and use [functions from Matrix](https://github.com/PX4/Matrix/pull/114) to reset covariances. These functions are tested with unit tests. 

Matrix::Matrix uses (i,j) to access a element, in comparison to the [i][j] access of the 2d float array. This affects portions of the code being auto generated with the MATLAB file. So therefore we would need to update the scripts that are creating the c++ code lines from the MATLAB derivation.

The templated matrix functions uses additional 630 bytes of flash memory. 

**Testing:**
Run a SITL plane simulation to get a [log](https://review.px4.io/plot_app?log=aa518e7d-2392-40b0-910e-9fd7844d617a). Replayed it with [master](https://review.px4.io/plot_app?log=42ce338a-2386-4aa3-9d0b-dcf4775ecbea) and this [PR](https://review.px4.io/plot_app?log=d8356b97-1cac-4bb8-b6da-75e041223976). The replayed behavior does in both cases not match the original log. Comparing the both replayed logs the curve are almost identical.  The only difference is a slight time shift. I am not sure what is causing this.
 When comparing the flight specs on flight review a small difference can be noticed. 

Notice that this is not fully a non-functional change, since the makeSymmetrical function is slightly adapted.

This solves issue #675


EKF/airspeed_fusion.cpp
```diff
@@ -222,7 +221,7 @@ void Ekf::fuseAirspeed()
 			// apply the covariance corrections
 			for (unsigned row = 0; row < _k_num_states; row++) {
 				for (unsigned column = 0; column < _k_num_states; column++) {
-					P[row][column] = P[row][column] - KHP[row][column];
+					P(row,column) = P(row,column) - KHP(row,column);
```
jkflying: This can now just be `P -= KHP;`

EKF/covariance.cpp
```diff
@@ -744,34 +729,34 @@ void Ekf::fixCovarianceErrors()
...
+	P.makeRowColSymmetric<13>(0);
 
 	// the following states are optional and are deactivated when not required
 	// by ensuring the corresponding covariance matrix values are kept at zero
 
 	// accelerometer bias states
 	if ((_params.fusion_mode & MASK_INHIBIT_ACC_BIAS) || _accel_bias_inhibit) {
-		zeroRows(P, 13, 15);
-		zeroCols(P, 13, 15);
+
```
jkflying: Extra line

EKF/mag_fusion.cpp
```diff
@@ -755,7 +754,7 @@ void Ekf::fuseHeading()
 	// apply covariance correction via P_new = (I -K*H)*P
 	// first calculate expression for KHP
 	// then calculate P - KHP
-	float KHP[_k_num_states][_k_num_states];
+	matrix::SquareMatrix<float, _k_num_states> KHP {};
```
jkflying: don't need the {}

EKF/mag_fusion.cpp
```diff
@@ -899,17 +897,17 @@ void Ekf::fuseDeclination(float decl_sigma)
 	// first calculate expression for KHP
 	// then calculate P - KHP
 	// take advantage of the empty columns in KH to reduce the number of operations
-	float KHP[_k_num_states][_k_num_states];
+	matrix::SquareMatrix<float, _k_num_states> KHP {};
```
jkflying: {}

EKF/optflow_fusion.cpp
```diff
@@ -442,7 +442,7 @@ void Ekf::fuseOptFlow()
 		// apply covariance correction via P_new = (I -K*H)*P
 		// first calculate expression for KHP
 		// then calculate P - KHP
-		float KHP[_k_num_states][_k_num_states];
+		matrix::SquareMatrix<float, _k_num_states> KHP {};
```
jkflying: {}

EKF/sideslip_fusion.cpp
```diff
@@ -218,7 +218,7 @@ void Ekf::fuseSideslip()
 		// apply covariance correction via P_new = (I -K*H)*P
 		// first calculate expression for KHP
 		// then calculate P - KHP
-		float KHP[_k_num_states][_k_num_states];
+		matrix::SquareMatrix<float, _k_num_states> KHP {};
```
jkflying: {}

EKF/vel_pos_fusion.cpp
```diff
@@ -146,14 +146,14 @@ void Ekf::fuseVelPosHeight(const float innov, const float innov_var, const int o
 
 	// calculate kalman gain K = PHS, where S = 1/innovation variance
 	for (int row = 0; row < _k_num_states; row++) {
-		Kfusion[row] = P[row][state_index] / innov_var;
+		Kfusion[row] = P(row,state_index) / innov_var;
 	}
 
-	float KHP[_k_num_states][_k_num_states];
+	matrix::SquareMatrix<float, _k_num_states> KHP {};
```
jkflying: {}

EKF/covariance.cpp
```diff
@@ -795,25 +780,12 @@ void Ekf::fixCovarianceErrors()
...
 		if (resetRequired) {
-			float delVelBiasVar[3];
-
-			// store all delta velocity bias variances
-			for (uint8_t stateIndex = 13; stateIndex <= 15; stateIndex++) {
-				delVelBiasVar[stateIndex - 13] = P[stateIndex][stateIndex];
-			}
-
-			// reset all delta velocity bias covariances
-			zeroCols(P, 13, 15);
```
jkflying: This was probably wrong

priseborough:
@kamilritz I've pushed the changes to the auto-code scripts to convert the syntax for the P[row][col] and nextP[row][col] to P(row,col) and nextP(row'col). Can you please use the output from the C_code_use_matrix_lib24.txt file in your changes.

dagar:
Nice, and the overall flash difference isn't too bad.

image

@TSC21 this should make https://github.com/PX4/Firmware/pull/11092 viable.

kamilritz:
> @kamilritz I've pushed the changes to the auto-code scripts to convert the syntax for the P[row][col] and nextP[row][col] to P(row,col) and nextP(row'col). Can you please use the output from the C_code_use_matrix_lib24.txt file in your changes.

@priseborough copied the code over locally and could not see any difference to commited version.

priseborough:
This was creating some additional operations due to the way the makeRowColSymmetrical operator was working compared to the legacy function it replaced. It was only a minor increase and I was able to remove a make symmetric operation in the covariance prediction.

priseborough:
Looking through the replay test log differences the differences in state and innovation sequences are minor but unexpected, and it is interesting that the log replayed using master starts using GPS and logging data at 27.71 sec whereas this PR does this a little earlier at 26.54 seconds. We should be getting EKF data before this and the original SITL log doesn't have this issue so I need to get that logging issue fixed so we can see what is happening between initial alignment and when GPS fusion commences to understand the reason for these differences.

kamilritz:
@priseborough I replayed the same code multiple times on the same log and the time shifting appears also in this case. There seems to be time shifts in between multiple runs inherent in the EKF log replay.

mhkabir:
@priseborough I think this is good to merge now

PR #679, opened 2019-12-09
start*Fusion and refactor reset velocity/position function

kamilritz:
A lot of the behavior of the EKF is steered by the control_status flags. Setting on of these fields to true or false is often not enough to change the mode of the EKF correctly. For example to start the heading fusion from the external vision data, the ev_yaw flag has to be set true, while the other heading flags have to be set true. Until now starting a certain type of fusion was managed in `control.cpp` via setting the flags.
This PR is adding separate functions for starting these fusions. which helps to reveal the intend of certain code lines to the reader.

Secondly, it is refactoring the `resetVelocity()` and `resetPosition()` functions. These function resetted the states according to the activated control_status flags. This was an issue for triggering a reset to a certain sensor source, such as vision velocity. E.g. Vision velocity was  always shadowed by an active GPS velocity.

Current issue: This does not fit on fmu v2.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -465,6 +465,7 @@ union filter_control_status_u {
 		uint32_t mag_aligned_in_flight   : 1; ///< 23 - true when the in-flight mag field alignment has been completed
 		uint32_t ev_vel      : 1; ///< 24 - true when local earth frame velocity data from external vision measurements are being fused
 		uint32_t synthetic_mag_z : 1; ///< 25 - true when we are using a synthesized measurement for the magnetometer Z component
+		uint32_t aux_vel     : 1; ///< 25 - true when auxiliary horizontal velocity data fusion is intended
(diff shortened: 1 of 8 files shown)


priseborough:
How did you intend to test these changes? I can go through the code, but that is not sufficient on its own given the scope of these changes.

priseborough:
This is a suggested list of cases for SITL testing:

1) Fail GPS (turn it off and back on after loss of navigation
2) Takeoff with declination set to 90 degrees to force loss of navigation
3) Step change the baro height by eg 20m
4) Turn off the baro for more than 10 second, then back on.
5) Optical flow test with flow data turned off, then back on after loss of navigation.
6) TBD - will complete list tomorrow

The log files could be used for subsequent replay testing of future changes.

Am happy to help with testing the reset behaviour.

PR #680, opened 2019-12-09
Support full vision variance 

kamilritz:
This PR add support to use all variance information in the visual_odometry message. It uses a 3d variance for position and velocity. Since we only fuse visual heading and not attitude, we do only support a one-dimensional uncertainty for the heading. 
Since external vision fusion in ECL is still fusing single measurement sequentially, we can not profit from the covariance values at the moment, so they are not added to the interface.

Secondly, the interface is expanded such that a 3D auxiliary velocity vector is passed instead of 2d vector. The vertical velocity is not fused at the moment. Since the control logic is not in place at the moment. This will be added, when we add finer fusion control granularity for all sensors.

See [here]() for the corresponding Firmware PR.

Unit tested cases:
- fusion start logic for external vision position, velocity and heading
- state's horizontal velocity uncertainty is decreasing faster in direction of smaller uncertainty in vision velocity measurement. Also in case where we have to align vision reference frame
- correct estimation of the alignment of the vision reference

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1398,14 +1398,16 @@ void Ekf::controlAuxVelFusion()
 		Vector2f aux_vel_innov_gate;
 		Vector3f aux_vel_obs_var;
 
-		_aux_vel_innov(0) = _state.vel(0) - _auxvel_sample_delayed.velNE(0);
-		_aux_vel_innov(1) = _state.vel(1) - _auxvel_sample_delayed.velNE(1);
+		_aux_vel_innov = _state.vel - _auxvel_sample_delayed.vel;
+		aux_vel_obs_var = _auxvel_sample_delayed.velVar;
 		aux_vel_innov_gate(0) = _params.auxvel_gate;
-		aux_vel_obs_var(0) = _auxvel_sample_delayed.velVarNE(0);
-		aux_vel_obs_var(1) = _auxvel_sample_delayed.velVarNE(1);
 
 		fuseHorizontalVelocity(_aux_vel_innov, aux_vel_innov_gate, aux_vel_obs_var,
 				_aux_vel_innov_var, _aux_vel_test_ratio);
+
+		// Can be enabled after bit for this is added to EKF_AID_MASK
+		// fuseVerticalVelocity(_aux_vel_innov, aux_vel_innov_gate, aux_vel_obs_var,
+		//		_aux_vel_innov_var, _aux_vel_test_ratio);
 
 	}
 }
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -238,7 +238,7 @@ void Ekf::controlExternalVisionFusion()
 				uncorrelateQuatStates();
 
 				// adjust the quaternion covariances estimated yaw error
-				increaseQuatYawErrVariance(sq(fmaxf(_ev_sample_delayed.angErr, 1.0e-2f)));
+				increaseQuatYawErrVariance(fmaxf(_ev_sample_delayed.angVar, sq(1.0e-2f)));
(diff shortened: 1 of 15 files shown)


EKF/control.cpp
```diff
@@ -361,9 +363,7 @@ void Ekf::controlExternalVisionFusion()
 			Vector3f vel_offset_earth = _R_to_earth * vel_offset_body;
 			vel_aligned -= vel_offset_earth;
 
-			_ev_vel_innov(0) = _state.vel(0) - vel_aligned(0);
-			_ev_vel_innov(1) = _state.vel(1) - vel_aligned(1);
-			_ev_vel_innov(2) = _state.vel(2) - vel_aligned(2);
+			_ev_vel_innov= _state.vel - vel_aligned;
```
jkflying:
```suggestion
			_ev_vel_innov = _state.vel - vel_aligned;
```

mhkabir:
This is awesome!

This came up sometime back for me, when I was working on fast flights down long tunnels with vision. Assuming isotropic measurement uncertainty is not a good assumption especially for visual-inertial odometry systems where the error strongly depends on direction of motion and feature distribution.

Another issue I ran into after implementing this was that the covariances too need to be rotated into the navigation frame (when using NED for GPS and an arbitrary vision frame) before being used. *I think this is missing in this PR, but could've missed it on mobile.*

mhkabir:
Also need to review that switching the auxvel interface to 3D doesn't have unexpected effects on the height when the external system doesn't provide a Z vel estimate (like the current landing target estimator implementation)

kamilritz:
> Another issue I ran into after implementing this was that the covariances too need to be rotated into the navigation frame (when using NED for GPS and an arbitrary vision frame) before being used. I think this is missing in this PR, but could've missed it on mobile.

@mhkabir  I missed that, thanks for catching that. It is now added. Please have a look again. I do not have vision system with proper variance information, so I can not really test this at the moment.

kamilritz:
> Also need to review that switching the auxvel interface to 3D doesn't have unexpected effects on the height when the external system doesn't provide a Z vel estimate (like the current landing target estimator implementation)

@mhkabir IMO this should be handled by the EKF2_AID_MASK parameter at some point. Such that you configure which components you want to fuse. At the moment the vertical component fusion is commented out.

EKF/control.cpp
```diff
@@ -305,7 +305,10 @@ void Ekf::controlExternalVisionFusion()
 					_ev_pos_innov(1) = _state.pos(1) - _hpos_pred_prev(1) - ev_delta_pos(1);
 
 					// observation 1-STD error, incremental pos observation is expected to have more uncertainty
-					ev_pos_obs_var(0) = ev_pos_obs_var(1) = sq(fmaxf(_ev_sample_delayed.posErr, 0.5f));
+					Matrix3f ev_pos_var = matrix::diag(_ev_sample_delayed.posVar);
```
mhkabir: Why not simply

`Vector3f ev_pos_var = _ev_rot_mat * _ev_sample_delayed.posVar`?

mhkabir: OK I guess it's nicer to have it as a matrix if we ever end up actually propagating the covariances.

mhkabir:
@bresch Squash please next time :cry:

bresch:
@mhkabir oops, sorry for that.; I'll check how the commits look next time.

kamilritz:
@mhkabir @bresch I would argue that it was the right thing to do here since the commits were separated in independent parts. The mistake that I made was combining full vision variance, auxiliiary velocity interface change and flash space saving in one PR. Sorry for that.

PR #681, opened 2019-12-11
Remove obsolete code in "initialiseFilter"

kamilritz:
This PR removes three obsolete code pieces during filter initialization. 

1) There was a `_ev_counter` counting incoming vision samples. It was never used.

2) An IF statement is checking the for the `ev_yaw` flag during initialization. The `ev_yaw` flag can not be true before entering the `controlFusionMode()` function, which happens only after successful initialization.

3) We allowed to initialize to vision height directly during initialization. This is not necessary, since the same switch to vision height will happen [here](https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L1116).

4) An IF statement is checking the for the `rng_hgt` flag during initialization. The `rng_hgt` flag can not be true before entering the `controlFusionMode()` function, which happens only after successful initialization. I checked with setting `EKF2_HGT_MODE=range_sensor`.

@dagar This safes 500B of flash. Please wait for pulling this into Firmware. We need the space for ECL changes.

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -267,11 +267,6 @@ bool Ekf::initialiseFilter()
 			_state.pos(2) = -math::max(_rng_filt_state * _R_rng_to_earth_2_2, _params.rng_gnd_clearance);
 			ECL_INFO_TIMESTAMPED("EKF using range finder height - commencing alignment");
 
-		} else if (_control_status.flags.ev_hgt) {
-			// if we are using external vision data for height, then the vertical position state needs to be reset
-			// because the initialisation position is not the zero datum
-			resetHeight();
-
 		}
 
 		// try to initialise the terrain estimator
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -259,16 +259,6 @@ bool Ekf::initialiseFilter()
 			increaseQuatYawErrVariance(sq(fmaxf(_params.mag_heading_noise, 1.0e-2f)));
 		}
 
-		if (_control_status.flags.rng_hgt) {
-			// if we are using the range finder as the primary source, then calculate the baro height at origin so  we can use baro as a backup
-			// so it can be used as a backup ad set the initial height using the range finder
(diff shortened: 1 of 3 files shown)


EKF/ekf.cpp
```diff
@@ -274,29 +254,11 @@ bool Ekf::initialiseFilter()
 		initialiseCovariance();
 
 		// update the yaw angle variance using the variance of the measurement
-		if (_control_status.flags.ev_yaw) {
-			// using error estimate from external vision data TODO: this is never true
-			increaseQuatYawErrVariance(sq(fmaxf(_ev_sample_delayed.angErr, 1.0e-2f)));
-		} else if (_params.mag_fusion_type <= MAG_FUSE_TYPE_3D) {
+		if (_params.mag_fusion_type <= MAG_FUSE_TYPE_3D) {
 			// using magnetic heading tuning parameter
 			increaseQuatYawErrVariance(sq(fmaxf(_params.mag_heading_noise, 1.0e-2f)));
```
bresch: I think you can also get rid of this one as `increaseQuatYawErrVariance(sq(fmaxf(_params.mag_heading_noise, 1.0e-2f)));` is already called in `resetMagHeading` https://github.com/PX4/ecl/blob/181303488cfc4d2ae34f61f0ad9a34b49a2501f7/EKF/ekf_helper.cpp#L735-L737
after the tilt has been aligned: https://github.com/PX4/ecl/blob/181303488cfc4d2ae34f61f0ad9a34b49a2501f7/EKF/control.cpp#L60

kamilritz: Good point. Probably this can be removed as well. But this could be a fundamental functional change. Lets merge this as is and I will open a separate PR for this.

kamilritz:
Removing the initialization to `ev_hgt` makes the `_primary_hgt_source` variable obsolete. It was only additionally used in the GPS checks. This was potentially dangerous, as the `_primary_hgt_source` variable was only set once during filter initialization. In this location it is replace with `_params.vdist_sensor_type` which is used in all the other location.

With this the PR stays non-functional

PR #683, opened 2019-12-12
Refactor ECL Test

kamilritz:
With this PR, the functionality to feed different sensor data at different rates into an ekf object is pulled out of the test_EKF_basics file into separate classes.

This will help to add future unit tests that require the Ekf object to get simulated data from the sensors.

Next step will be to add more sensors and expand the test cases.


test/SensorSimulator/Baro.cpp
```diff
@@ -0,0 +1,20 @@
+#include "Baro.h"
+
```
nicovanduijn: A `namespace sensor::baro{` might be nice

test/test_EKF_basics.cpp
```diff
@@ -34,128 +34,33 @@
...
-	bool _fuse_mag{true};
-	bool _fuse_gps{false};	// GPS measurements are expected to not come in from beginning
-
-	// GPS message
-	gps_message _gps_message{};
-
-	uint32_t _update_dt_us{};			// greatest common divider of all basic sensor periods
-	const uint32_t _init_duration_us{2000000};	// 2s	Duration of
+	Ekf* _ekf;
+	SensorSimulator* _sensor_simulator;
```
nicovanduijn: I don't understand why you're using pointers here. Can't you just create an instance of EKF and SensorSimulator? Then you don't have to manage the heap memory and worry about leaking it in case of a crash. I see the teardown, but you never know what happens in between

test/SensorSimulator/Sensor.h
```diff
@@ -0,0 +1,78 @@
...
+	// time in microseconds
+	uint32_t _update_period;
+	uint32_t _time_last_data_sent{0};
+
+	bool _is_running{false};
+
+	bool should_send(uint32_t time);
+
+	// Checks that the right amount time passed since last send data to fulfill rate
+	bool is_time_to_send(uint32_t time);
```
nicovanduijn: can be const

test/SensorSimulator/Sensor.h
```diff
@@ -0,0 +1,78 @@
...
+protected:
+
+	Ekf* _ekf;
+	// time in microseconds
+	uint32_t _update_period;
+	uint32_t _time_last_data_sent{0};
+
+	bool _is_running{false};
+
+	bool should_send(uint32_t time);
```
nicovanduijn: can be const

test/SensorSimulator/Sensor.h
```diff
@@ -0,0 +1,78 @@
...
+public:
+
+	Sensor(Ekf* ekf);
+	virtual ~Sensor();
+
+	void update(uint32_t time);
+
+	void setRate(uint32_t rate){ _update_period = uint32_t(1000000)/rate; }
+
+	bool isRunning(){ return _is_running; }
```
nicovanduijn: can be const

test/SensorSimulator/Baro.cpp
```diff
@@ -0,0 +1,20 @@
...
+Baro::Baro(Ekf* ekf):Sensor(ekf)
+{
+}
+
+Baro::~Baro()
+{
+}
+
+void Baro::send(uint32_t time)
```
nicovanduijn: here and in the other sensors: use the `override` keyword

test/SensorSimulator/Imu.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+void Imu::setData(Vector3f accel, Vector3f gyro)
+{
+	setAccelData(accel);
+	setGyroData(gyro);
+}
+
+void Imu::setAccelData(Vector3f accel)
+{
+	_accel_data = accel;
```
nicovanduijn: Are you sure Vector3f does a deep copy for assignment? (Probably, but it's good to be aware of what the copy assignment operation does for a specific object)

test/test_EKF_basics.cpp
```diff
@@ -231,12 +136,12 @@ TEST_F(EkfInitializationTest, gpsFusion)
 	// GIVEN: initialized EKF with default IMU, baro and mag input for 2s
 	// WHEN: setting GPS measurements for 11s, minimum GPS health time is set to 10 sec
 
-	_fuse_gps = true;
-	update_with_const_sensors(11000000,Vector3f{0.0f,0.0f,0.0f},Vector3f{0.0f,0.0f,-CONSTANTS_ONE_G}); // for further 3s
+	_sensor_simulator->startGps();
+	_sensor_simulator->run(11000000); // for further 3s
```
jkflying: fix comment

test/SensorSimulator/SensorSimulator.cpp
```diff
@@ -0,0 +1,81 @@
...
+	gps_data.vel_ned[1] = 0.0f;
+	gps_data.vel_ned[2] = 0.0f;
+	gps_data.vel_ned_valid = 1;
+	gps_data.nsats = 16;
+	gps_data.pdop = 0.0f;
+
+	return gps_data;
+}
+
+void SensorSimulator::run(uint32_t duration)
```
jkflying: Be specific for the units here. Call this eg. run_microseconds(), otherwise someone needs to read your code here to know how to use it.

it would be nice to also have a `run_seconds(int secs){run_microseconds(secs*1000000);}` to make it nicer to use in the general case.

test/SensorSimulator/SensorSimulator.cpp
```diff
@@ -0,0 +1,81 @@
...
+	uint32_t start_time = _time;
+
+	for(;_time < start_time + duration; _time+=1000)
+	{
+		_imu.update(_time);
+		_mag.update(_time);
+		_baro.update(_time);
+		_gps.update(_time);
+
+		_ekf->update();
```
jkflying: Just to check, EKF doesn't need to know that the time updated? Just timestamping the data going in is enough?

test/SensorSimulator/Imu.cpp
jkflying: Yep, it's a deep copy :+1:

test/SensorSimulator/SensorSimulator.cpp
kamilritz: yes, that is enough

kamilritz:
@nicovanduijn @jkflying  Thank you very much for the extensive review. This helped me a lot.

PR #686, opened 2019-12-16
Tilt Initialisation: Average Filter -> LowPass Filter

kamilritz:
The EKF is using accel measurements to initialize pitch and roll. Under the assumption that we are not moving, the accel measurement contains the gravity vector.

**Changes:**
To robustify against noise we previously average all the accel measurement over the period we wait for enough mag and baro measurements.This PR removes this averaging approach with a lowpass filter approach. The used filter is the usual 0.1/0.9 digital lowpass filter.

**Testing**:
I compared the time for successful tilt alignment in EKF replay to current master. There is no noticeable difference. The used log is recorded under soft motion. I also tested if it aligns under rough motion and it does. See log here:
https://review.px4.io/plot_app?log=3bf47850-4009-4793-9018-60fa10d82c7b

**Outlook:**
This will be needed for immediate EKF state resets. The averaging approach will introduce delay into a state reset, while a filtered accel measurement is immediately available.

**Question:**
Can we combine this with: https://github.com/kamilritz/ecl/blob/clean_initialiseFilter/EKF/covariance.cpp#L175

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -61,18 +61,18 @@ void Ekf::controlFusionModes()
 
 			// send alignment status message to the console
 			if (_control_status.flags.baro_hgt) {
-				ECL_INFO("EKF aligned, (pressure height, IMU buf: %i, OBS buf: %i)", (int)_imu_buffer_length, (int)_obs_buffer_length);
+				ECL_INFO("%llu: EKF aligned, (pressure height, IMU buf: %i, OBS buf: %i)", (unsigned long long)_imu_sample_delayed.time_us, (int)_imu_buffer_length, (int)_obs_buffer_length);
 
 			} else if (_control_status.flags.ev_hgt) {
-				ECL_INFO("EKF aligned, (EV height, IMU buf: %i, OBS buf: %i)", (int)_imu_buffer_length, (int)_obs_buffer_length);
+				ECL_INFO("%llu: EKF aligned, (EV height, IMU buf: %i, OBS buf: %i)", (unsigned long long)_imu_sample_delayed.time_us, (int)_imu_buffer_length, (int)_obs_buffer_length);
 
 			} else if (_control_status.flags.gps_hgt) {
-				ECL_INFO("EKF aligned, (GPS height, IMU buf: %i, OBS buf: %i)", (int)_imu_buffer_length, (int)_obs_buffer_length);
+				ECL_INFO("%llu: EKF aligned, (GPS height, IMU buf: %i, OBS buf: %i)", (unsigned long long)_imu_sample_delayed.time_us, (int)_imu_buffer_length, (int)_obs_buffer_length);
 
 			} else if (_control_status.flags.rng_hgt) {
-				ECL_INFO("EKF aligned, (range height, IMU buf: %i, OBS buf: %i)", (int)_imu_buffer_length, (int)_obs_buffer_length);
+				ECL_INFO("%llu: EKF aligned, (range height, IMU buf: %i, OBS buf: %i)", (unsigned long long)_imu_sample_delayed.time_us, (int)_imu_buffer_length, (int)_obs_buffer_length);
 			} else {
-				ECL_ERR("EKF aligned, (unknown height, IMU buf: %i, OBS buf: %i)", (int)_imu_buffer_length, (int)_obs_buffer_length);
+				ECL_ERR("%llu: EKF aligned, (unknown height, IMU buf: %i, OBS buf: %i)", (unsigned long long)_imu_sample_delayed.time_us, (int)_imu_buffer_length, (int)_obs_buffer_length);
 			}
 
 		}
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -114,6 +114,8 @@ bool Ekf::update()
 
 	// Only run the filter if IMU data in the buffer has been updated
 	if (_imu_updated) {
+		const imuSample &imu_init = _imu_buffer.get_newest();
+		_accel_lpf.update(imu_init.delta_vel);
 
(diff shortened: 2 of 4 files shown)


EKF/ekf.cpp
```diff
@@ -199,35 +209,10 @@ bool Ekf::initialiseFilter()
 		_gps_drift_velD = 0.0f;
 		_gps_alt_ref = 0.0f;
 
-		// Zero all of the states
-		_state.vel.setZero();
```
kamilritz: This code was a duplicate

EKF/ekf.cpp
```diff
@@ -258,6 +243,25 @@ bool Ekf::initialiseFilter()
...
 }
 
+bool Ekf::initialiseTilt()
+{
+	if (_accel_lpf.getState().norm() < 0.001f) {
+		return false;
+	}
+
+	// get initial roll and pitch estimate from delta velocity vector, assuming vehicle is static
+	Vector3f gravity_in_body = _accel_lpf.getState();
```
bresch:
```suggestion
	const Vector3f gravity_in_body = _accel_lpf.getState();
```

EKF/ekf.cpp
```diff
@@ -258,6 +243,25 @@ bool Ekf::initialiseFilter()
...
+	gravity_in_body.normalize();
+	float pitch = asinf(gravity_in_body(0));
```
bresch:
```suggestion
	const float pitch = asinf(gravity_in_body(0));
```

EKF/ekf.cpp
```diff
@@ -258,6 +243,25 @@ bool Ekf::initialiseFilter()
...
+	float roll = atan2f(-gravity_in_body(1), -gravity_in_body(2));
```
bresch:
```suggestion
	const float roll = atan2f(-gravity_in_body(1), -gravity_in_body(2));
```

EKF/ekf.cpp
```diff
@@ -258,6 +243,25 @@ bool Ekf::initialiseFilter()
...
+
+	Eulerf euler_init(roll, pitch, 0.0f);
```
bresch:
```suggestion
	const Eulerf euler_init(roll, pitch, 0.0f);
```

EKF/ekf.h
```diff
@@ -464,8 +466,8 @@ class Ekf : public EstimatorInterface
 	float _rng_filt_state{0.0f};		///< filtered height measurement (m)
 	uint32_t _mag_counter{0};		///< number of magnetometer samples read during initialisation
 	uint64_t _time_last_mag{0};		///< measurement time of last magnetomter sample (uSec)
-	AlphaFilterVector3f _mag_lpf;		///< filtered magnetometer measurement (Gauss)
-	Vector3f _delVel_sum;			///< summed delta velocity (m/sec)
+	AlphaFilterVector3f _mag_lpf;		///< filtered magnetometer measurement for instant reset(Gauss)
+	AlphaFilterVector3f _accel_lpf;		///< filtered accelerometer measurement for instant reset(Gauss
```
bresch:
```suggestion
	AlphaFilterVector3f _accel_lpf;		///< filtered accelerometer measurement for instant reset(Gauss)
```

EKF/ekf.cpp
kamilritz: We normalize in the next line. Therefore it can not be constant

EKF/ekf.cpp
kamilritz: nice

EKF/control.cpp
```diff
@@ -61,18 +61,18 @@ void Ekf::controlFusionModes()
 
 			// send alignment status message to the console
 			if (_control_status.flags.baro_hgt) {
-				ECL_INFO("EKF aligned, (pressure height, IMU buf: %i, OBS buf: %i)", (int)_imu_buffer_length, (int)_obs_buffer_length);
+				ECL_INFO("%llu: EKF aligned, (pressure height, IMU buf: %i, OBS buf: %i)", (unsigned long long)_imu_sample_delayed.time_us, (int)_imu_buffer_length, (int)_obs_buffer_length);
```
mhkabir: Isn't the ECL_INFO_TIMESTAMPED implemented for this exact case?

EKF/ekf.cpp
```diff
@@ -139,9 +141,17 @@ bool Ekf::initialiseFilter()
 {
 	// Keep accumulating measurements until we have a minimum of 10 samples for the required sensors
 
-	// Sum the IMU delta velocity measurements
+	// Filter accel for tilt initialization
 	const imuSample &imu_init = _imu_buffer.get_newest();
-	_delVel_sum += imu_init.delta_vel;
+	// Reset to imu sample on boot up
+	// Use mag_counter to reduce number of variables
+	if(_mag_counter == 0){
```
mhkabir: Not entirely happy about this reuse of mag_counter

kamilritz: Me neither, do you see any other solution?

EKF/control.cpp
kamilritz: Yes, but it is only working for statements without additional arguments. Otherwise I would added it already to this line.

EKF/ekf.cpp
bresch: Just use a boolean with a meaningful name that gets set to true only once.

EKF/control.cpp
mhkabir: You can add VA_ARGS to the macro and it will then work with arguments.

kamilritz: Yes, but the amount of argument values (also zero) can vary and not all the compiler running in CI support that. I tried hard.

PR #692, opened 2019-12-19
mag_fusion: re-introduce delay between mag reset and fusion

bresch:
This was dropped during the mag fusion refactor PR #662 

I directly added the check in `mag_fusion.cpp` since it is only used there. Is that fine for you @priseborough or this should absolutely not be done in case of 3D mag fusion selected by the user? (before, this check was only done when auto fusion was enabled but not in 3D fusion only)

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -473,7 +473,6 @@ private:
 
 	// Variables used to control activation of post takeoff functionality
 	float _last_on_ground_posD{0.0f};	///< last vertical position when the in_air status was false (m)
-	bool _flt_mag_align_converging{false};	///< true when the in-flight mag field post alignment convergence is being performd
 	uint64_t _flt_mag_align_start_time{0};	///< time that inflight magnetic field alignment started (uSec)
 	uint64_t _time_last_mov_3d_mag_suitable{0};	///< last system time that sufficient movement to use 3-axis magnetometer fusion was detected (uSec)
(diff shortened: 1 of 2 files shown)


bresch:
TODO: 
- [ ] SITL test

bresch:
@priseborough If I understand properly, the reason of that "update state only" is :
> Doing a hard reset of magnetic field states using a single sample can cause problems if the mag field sample is corrupted by noise (eg motor interference). Ideally multiple samples should be used, but this would require a significant code change to achieve so instead magnetic field state estimates are given time to form before they are used for vehicle yaw estimation. This is achieved by blocking the 3-axis fusion from accessing non magnetic field states and using heading fusion to constrain yaw drift for 10 seconds after a field reset. 621854c

However, we are not resetting the heading on a single sample anymore, but on a filtered value: see https://github.com/PX4/ecl/pull/662/commits/0b788cf5449377e4c535fa91eab2568cbf17f388
Is this extra protection, re-introduced by this PR still relevant? How can I reproduce the original issue in SITL?

priseborough:
@bresch We can also run into problems due to electromagnetic interference and hard iron offsets which are not improved by averaging and would be somewhat improved by this PR if the vehicle was rotated after the reset. Whether this change is still necessary, given the improvements to yaw reset covariance handling is not determined. However it was legacy behaviour that had been working without issue, so the onus of proof should be on its removal, not reinstatement IMO.

PR #693, opened 2019-12-19
ekf2: centralize mag covariance reset

bresch:
A complete and clean reset of the mag states covariances is now performed through the "resetMagCov"
function only. This avoid having slight differences of implementations across the code.

Log before this PR:
https://logs.px4.io/plot_app?log=f926bdca-989f-4f90-9483-0f3168660cd3
old

With this PR:
https://logs.px4.io/plot_app?log=ca6a418d-5da8-48ee-a636-e9d1bccd36b4
new

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -57,11 +57,7 @@ void Ekf::initialiseCovariance()
 
 	const float dt = FILTER_UPDATE_PERIOD_S;
 
-	// define the initial angle uncertainty as variances for a rotation vector
-	Vector3f rot_vec_var;
-	rot_vec_var.setAll(sq(_params.initial_tilt_err));
-
-	initialiseQuatCovariances(rot_vec_var);
+	resetQuatCov();
 
 	// velocity
 	P(4,4) = sq(fmaxf(_params.gps_vel_noise, 0.01f));
@@ -868,6 +864,12 @@ void Ekf::resetMagRelatedCovariances()
 
 void Ekf::resetQuatCov(){
 	zeroQuatCov();
+
+	// define the initial angle uncertainty as variances for a rotation vector
+	Vector3f rot_vec_var;
+	rot_vec_var.setAll(sq(_params.initial_tilt_err));
+
+	initialiseQuatCovariances(rot_vec_var);
 }
 
 void Ekf::zeroQuatCov()
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -97,14 +97,7 @@ void Ekf::initialiseCovariance()
 	// record IMU bias state covariance reset time - used to prevent resets being performed too often
 	_last_imu_bias_cov_reset_us = _imu_sample_delayed.time_us;
 
-	// earth frame and body frame magnetic field
-	// set to observation variance
-	for (uint8_t index = 16; index <= 21; index ++) {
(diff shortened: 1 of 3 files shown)


EKF/ekf_helper.cpp
```diff
@@ -496,17 +496,7 @@ bool Ekf::realignYawGPS()
...
 
-			// reset the corresponding rows and columns in the covariance matrix and set the variances on the magnetic field states to the measurement variance
-			clearMagCov();
-
-			if (_control_status.flags.mag_3D) {
-				for (uint8_t index = 16; index <= 21; index ++) {
-					P[index][index] = sq(_params.mag_noise);
-				}
-
-				// save covariance data for re-use when auto-switching between heading and 3-axis fusion
```
bresch: Note that this was now working because that part is only executed in mag 3D mode

kamilritz:
@bresch The mentioned PR went in and created some small conflicts here.

bresch:
@kamilritz I solved the conflicts and rebased on master.

EKF/covariance.cpp
```diff
@@ -95,14 +95,7 @@ void Ekf::initialiseCovariance()
...
 	_last_imu_bias_cov_reset_us = _imu_sample_delayed.time_us;
 
-	// earth frame and body frame magnetic field
-	// set to observation variance
-	for (uint8_t index = 16; index <= 21; index ++) {
-		P(index,index) = sq(_params.mag_noise);
-	}
-
-	// save covariance data for re-use when auto-switching between heading and 3-axis fusion
-	saveMagCovData();
```
kamilritz: We will stop saving the mag covariance data at this point in case "_control_status.flags.mag_3D=true".

EKF/covariance.cpp
```diff
@@ -867,16 +860,35 @@ void Ekf::fixCovarianceErrors(bool force_symmetry)
 
 void Ekf::resetMagRelatedCovariances()
 {
-	// set the quaternion covariance terms to zero
+	resetQuatCov();
+	resetMagCov();
+}
+
+void Ekf::resetQuatCov(){
```
kamilritz: This seems odd. I expect a function resetQuatCov to do more then just zeroing it. But that is what was done earlier. Maybe we should reset it to the initial value?

EKF/covariance.cpp
bresch: Yes, and this is perfectly fine because we automatically save the mag covariance when we switch out of 3d fusion; there is no need to save it when we are currently using it.

EKF/covariance.cpp
bresch: Maybe, yes. This is the reason I created this intermediate function, but since this PR is mostly "non-functional" I didn't want to change it here.

bresch: Actually I changed it here: https://github.com/PX4/ecl/pull/693/commits/5c6a135e2f53d602840a6354ea096561feb3a1ea
Using the initial uncertainty makes more sense than zero

bresch:
@priseborough 
>The quaternion covariance would be better set using the Ekf::initialiseQuatCovariances function.

I added that here: https://github.com/PX4/ecl/pull/693/commits/5c6a135e2f53d602840a6354ea096561feb3a1ea

PR #694, opened 2019-12-19
Refactor imu downsampling

kamilritz:
This pulls out the down sampling of the IMU into a separate class. This is nice as we can hide a few variables inside the class.

Additionally it increases code readability introducing two new functions: `computeVibrationMetrics()` and `checkIfVehicleAtRest()`.

This increased the flash space by ~800 Bytes. Therefore I removed duplicated initialization of variable and reduced the number of repeatedly stored strings.

Current added flash space: 344 Bytes

**Testing:**
- Through unit tests and SITL mission test
- Flight tested in [here](https://github.com/PX4/Firmware/pull/13870)

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -47,12 +47,12 @@
 bool Ekf::init(uint64_t timestamp)
 {
 	bool ret = initialise_interface(timestamp);
-	reset(timestamp);
+	reset();
 
 	return ret;
 }
 
-void Ekf::reset(uint64_t timestamp)
+void Ekf::reset()
 {
 	_state.vel.setZero();
 	_state.pos.setZero();
@@ -70,13 +70,6 @@ void Ekf::reset(uint64_t timestamp)
 	initialiseCovariance();
 
 	_delta_angle_corr.setZero();
-	_imu_down_sampled.delta_ang.setZero();
-	_imu_down_sampled.delta_vel.setZero();
-	_imu_down_sampled.delta_ang_dt = 0.0f;
-	_imu_down_sampled.delta_vel_dt = 0.0f;
-	_imu_down_sampled.time_us = timestamp;
-
-	_q_down_sampled.setIdentity();
 
 	_imu_updated = false;
 	_NED_origin_initialised = false;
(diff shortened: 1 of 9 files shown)


EKF/ekf.cpp
```diff
@@ -70,13 +70,6 @@ void Ekf::reset(uint64_t timestamp)
 	initialiseCovariance();
 
 	_delta_angle_corr.setZero();
-	_imu_down_sampled.delta_ang.setZero();
```
kamilritz: Initialization moved to class

EKF/ekf.h
```diff
@@ -438,8 +436,6 @@ class Ekf : public EstimatorInterface
 
 	// output predictor states
 	Vector3f _delta_angle_corr;	///< delta angle correction vector (rad)
-	imuSample _imu_down_sampled{};	///< down sampled imu data (sensor rate -> filter update rate)
```
kamilritz: Moved to class

EKF/estimator_interface.cpp
```diff
@@ -517,30 +545,15 @@ bool EstimatorInterface::initialise_interface(uint64_t timestamp)
...
-
-	_time_last_imu = 0;
-	_time_last_gps = 0;
-	_time_last_mag = 0;
-	_time_last_baro = 0;
-	_time_last_range = 0;
-	_time_last_airspeed = 0;
-	_time_last_optflow = 0;
 	_fault_status.value = 0;
-	_time_last_ext_vision = 0;
```
kamilritz: Already initialized through initialization brackets {} in the `estimator_interface.hpp`

priseborough:
If this is a non-functional change, then we should be able to do back to back comparisons of ecl replay output with and without these changes.  A SITL log can be used as input.   The results should be identical. There are some subtleties associated with the rotation prior to accumulation and order of calculations that could be missed just by only looking at a log with the PR included.

priseborough:
We should also check CPU load before and after on a F4 board.

priseborough:
I've run a replay log before and after the change 3 times each. The differences in the kinematic states appear to be only those due to small timing differences between replays.

priseborough:
I've checked on my F7 setup and there is no noticeable change in CPU load and a negligible increase in RAM before and after.

priseborough:
@kamilritz The code and testing look sOK from my end.  Can you please resolve conflicts. TXS

priseborough:
Needs PR https://github.com/PX4/Firmware/pull/13705 to build

PR #695, opened 2019-12-23
[WIP]Ekf2: centralize estimator alignement

bresch:
Move all the `resetMagHeading` calls in a central place that checks the mag field, proceed to the reset and clear/set the desired flags. This avoids copy/paste mistakes and offers a consistent and deterministic behavior.

The  initial alignment of the EKF is robustified as it also used this logic and prevents aligning when the mag field is disturbed.
During the test below, I exposed the vehicle to a strong magnetic disturbance until t=75s, the yaw then aligns.
2019-12-23_18-39-58_01_plot

Others:
- remove unused `_mag_use_inhibit_prev`

TODO:
- [ ] non-regression SITL tests
- [ ] unit test coverage

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -688,7 +688,7 @@ private:
 	bool isMeasuredMatchingGpsMagStrength(float mag_strength) const;
 	bool isMeasuredMatchingAverageMagStrength(float mag_strength) const;
 	static bool isMeasuredMatchingExpected(float measured, float expected, float gate);
-	void processMagResetFlags();
+	void processMagResetFlags(bool force = false);
 	bool isYawResetRequested() const;
 	void runPostAlignmentActions();
 	void runMagAndMagDeclFusions();
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -560,9 +560,6 @@ private:
 	// fuse single velocity and position measurement
 	void fuseVelPosHeight(const float innov, const float innov_var, const int obs_index);
 
-	// reset velocity states of the ekf
-	bool resetVelocity();
-
 	// fuse optical flow line of sight rate measurements
 	void fuseOptFlow();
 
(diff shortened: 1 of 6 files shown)


EKF/ekf.h
```diff
@@ -692,10 +692,14 @@ class Ekf : public EstimatorInterface
 	void checkMagInhibition();
 	bool shouldInhibitMag() const;
 	void checkMagFieldStrength();
+	bool isMagStrengthPlausible(float mag_strength) const;
 	bool isStrongMagneticDisturbance() const;
-	bool isMeasuredMatchingGpsMagStrength() const;
-	bool isMeasuredMatchingAverageMagStrength() const;
+	bool isMeasuredMatchingGpsMagStrength(float mag_strength) const;
+	bool isMeasuredMatchingAverageMagStrength(float mag_strength) const;
```
kamilritz: suggest `isMeasuredMagStrengthMatchingEarthsAverage()`

priseborough:
This PR is likely to unnecessarily frustrate users due to preventing them from taking off when the magnetic yaw interference is only at ground level.

It is not unusual for the mag field to be perturbed when the vehicle is on the ground, even when outdoors due to ferrous materials in adjacent vehicles, buildings and reinforced concrete. 

This is also related to the magnetometer consistency check in the commander. We need to define the errors we are trying to catch and what steps the user will be required to be able to fly.

When testing this we should ensure that:

- Yaw fusion is delayed until tilt estimates have stabilised.
- When not flying and movement is detected, periodically fuse the current INS heading to prevent unconstrained quaternion variance growth.
- When not flying and not moving, fuse last recorded INS yaw before movement stopped.
- The movement test is robust to the type of slower motion that can occur when launching from boats/ships. The result of a false negative for the  'is moving' test can be be a large error in the yaw gyro bias that would lead to loss of yaw accuracy and navigation after takeoff. This is a failure mode introduced by this PR for default param settings and needs to be balanced against the failure mode this PR addresses.
- Ensure we handle the case where the mag field stays perturbed. In that instance it will never start fusing GPS and the user will never be able take off.

bresch:
The code changed a lot and I would have to rework this PR. Closing for now.

PR #702, opened 2020-01-09
Adding airspeed sensor to sensor_simulator

kamilritz:
- adds a airspeed sensor to the sensor_simulator.
- add some file types to .gitignore
- changes time for the sensor simulator from uint32_t to uint64_t to match EKF's setter functions.
- adds a basic airspeed fusion test


test/sensor_simulator/airspeed.h
```diff
@@ -0,0 +1,64 @@
...
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ ****************************************************************************/
+
+/**
+ * Feeds Ekf with Mag data
```
RomanBapst: @kamilritz change to airspeed

test/sensor_simulator/sensor_simulator.cpp
```diff
@@ -30,6 +31,7 @@ void SensorSimulator::setSensorDataToDefault()
 	_flow.setRateHz(50);
 	_rng.setRateHz(30);
 	_vio.setRateHz(30);
+	_airspeed.setRateHz(30); // TODO: check this rate
```
RomanBapst: @kamilritz I think the rate is currently quite fast, like 100Hz.

RomanBapst: @dagar Is this [comment](https://github.com/PX4/Firmware/blob/master/src/drivers/differential_pressure/sdp3x/SDP3X.hpp#L66) correct?

test/sensor_simulator/airspeed.h
```diff
@@ -0,0 +1,64 @@
...
+class Airspeed: public Sensor
+{
+public:
+	Airspeed(std::shared_ptr<Ekf> ekf);
+	~Airspeed();
+
+	void setData(float true_airspeed, float eas2tas);
+
+private:
+	float _true_airspeed_data;
```
RomanBapst: @kamilritz Can you initialize these value to 0 here?

kamilritz: Good catch. I missed that also for the other sensors.

RomanBapst: @kamilritz Yes, would be good if you go through the others quickly

kamilritz: @RomanBapst Yes I added that to this PR

test/sensor_simulator/sensor_simulator.cpp
dagar: 100 Hz is more or less correct, although the SDP3X can be much faster. All the others are currently sampled at 100 Hz, but I'll do a pass to review them all shortly. https://github.com/PX4/Firmware/pull/13586

test/sensor_simulator/airspeed.h
kamilritz: @RomanBapst Yes I already added that to this PR.

PR #703, opened 2020-01-10
fixed output state variable name state_variance to state_variances

wangtaooo:
very nice


wangtaooo:
This is my first time to submit. now i am very happy.Thank you too.

PR #706, opened 2020-01-11
range_finder_checks: fix validity flag

bresch:
The result of the "range stuck" detection logic was overriding the current validity of the range finder data. This means that even if the quality of the data is 0, if it is not "stuck", the validity was set to true.

@RomanBapst I think we messed it up in https://github.com/PX4/ecl/pull/655

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -453,7 +453,7 @@ void Ekf::controlOpticalFlowFusion()
 		if (!_inhibit_flow_use && _control_status.flags.opt_flow) {
 			// inhibit use of optical flow if motion is unsuitable and we are not reliant on it for flight navigation
 			bool preflight_motion_not_ok = !_control_status.flags.in_air && ((_imu_sample_delayed.time_us - _time_good_motion_us) > (uint64_t)1E5);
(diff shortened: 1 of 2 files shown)


priseborough:
Do you have test evidence that this change fixes the bug?

bresch:
@priseborough Yes, I cant share the logs, but here you can see that the range finder is fused even if the reported quality is 0:
2020-01-15_02-58-39_01_plot
Replaying the same log with this fix shows that the bad samples aren't fused anymore:
2020-01-15_03-13-52_01_plot

bresch:
The second commit (https://github.com/PX4/ecl/pull/706/commits/f4f3c303142db10b3da1b97524de7e3d0ea69e23) fixes the issue that the optical flow fusion is in sync with the range finder data quality instead of the quality of the terrain estimate:
2020-01-15_03-19-25_01_plot

Same log replayed with this PR:
2020-01-15_03-18-42_01_plot

Also note that the terrain validity flag isn't always set to true anymore, because the range finder is only fused when the quality is "-1" and when no measurement is fused for a certain amount of time, the terrain validity goes to false (fixed by the first commit).

PR #707, opened 2020-01-13
ekf: add function to set the GPS antenna offset

CarlOlsson:
Comes together with https://github.com/PX4/Firmware/pull/13932
It works just setting the _params.gps_pos_body this way but since the value will change continuously maybe we should move gps_pos_body out of the _params struct.

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -284,6 +284,9 @@ public:
 	// set minimum continuous period without GPS fail required to mark a healthy GPS status
 	void set_min_required_gps_health_time(uint32_t time_us) { _min_gps_health_time_us = time_us; }
 
+	// set GPS sensor position
+	void set_GPS_pos_body(const Vector3f &gps_pos_body) { _params.gps_pos_body = gps_pos_body; }
+
 private:
 
 	static constexpr uint8_t _k_num_states{24};		///< number of EKF states


EKF/ekf.h
```diff
@@ -284,6 +284,9 @@ class Ekf : public EstimatorInterface
 	// set minimum continuous period without GPS fail required to mark a healthy GPS status
 	void set_min_required_gps_health_time(uint32_t time_us) { _min_gps_health_time_us = time_us; }
 
+	// set GPS sensor position
+	void set_GPS_pos_body(const Vector3f &gps_pos_body) { _params.gps_pos_body = gps_pos_body; }
```
dagar: This is going to be overwritten when there's a parameter update. 

https://github.com/PX4/Firmware/blob/4f362f583511b6326fda859934905e553f74eb03/src/modules/ekf2/ekf2_main.cpp#L612-L614

LorenzMeier:
@CarlOlsson Are you going to finish this or is it stale? It has pending review feedback.

CarlOlsson:
> This is going to be overwritten when there's a parameter update.

Yes, but it is set again every time we push a GPS measurement into ECL

> Are you going to finish this or is it stale? It has pending review feedback.

@dagar We have been flying this for a year but I don't have time now to clean it up, feel free to close it. I guess now days there are better ways to do it due to improved sensor architecture.

dagar:
@CarlOlsson we should be able to do this properly firmware side as soon as GPS device ids are in place. Then we'll move the GPS offsets out and EKF2 and handle more like sensor calibrations.

 - `SENS_GPSn_ID`
 - `SENS_GPSn_POS{X,Y,Z}`

CarlOlsson:
Discussed with @dagar closing for now until an improved architecture is in place

PR #708, opened 2020-01-13
Support vision velocity expressed in body frame too

kamilritz:
At the moment we expect the external vision's velocity to be expressed in world/local frame. This PR allows for the vision velocity also being expressed in body frame. For this an enum is added to the vision sample that stores the information about the used velocity frame.
VIO systems that output velocity in body frame do not rely on its orientation data to be correct or existing to transform the velocity frame to world coordinates. Instead we can directly transform the velocity to ekf's local frame by using the ekf's quaternion state.

**Testing:**
- Expanded unit tests to check if we can can fuse velocity in body and local frame
- Fly tests: [EV_VEL only](https://review.px4.io/plot_app?log=e43feac7-7f21-4bbf-8dc4-54ab0da1f002), [GPS & EV_VEL](https://review.px4.io/plot_app?log=646bc0f8-cf83-4418-bf72-9ed40bb7e502)

**Related Firmware PR:** https://github.com/PX4/Firmware/pull/13934

**Issue:**
Velocity variance was lower bounded by a parameter on the firmware side. This is not possible any more. We would need to get this parameter from the firmware side to apply it in ECL after we transform the velocity covariance to local frame.

FYI @nicovanduijn @bresch @jkflying

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -202,14 +202,14 @@ void Ekf::controlExternalVisionFusion()
 				if (_params.fusion_mode & MASK_USE_EVPOS && !_control_status.flags.ev_pos) {
 					_control_status.flags.ev_pos = true;
 					resetPosition();
-					ECL_INFO_TIMESTAMPED("commencing external vision position fusion");
+					ECL_INFO_TIMESTAMPED("commencing vision position fusion");
 				}
 
 				// turn on use of external vision measurements for velocity
 				if (_params.fusion_mode & MASK_USE_EVVEL && !_control_status.flags.ev_vel) {
 					_control_status.flags.ev_vel = true;
 					resetVelocity();
-					ECL_INFO_TIMESTAMPED("commencing external vision velocity fusion");
+					ECL_INFO_TIMESTAMPED("commencing vision velocity fusion");
 				}
 			}
 		}
@@ -235,7 +235,7 @@ void Ekf::controlExternalVisionFusion()
 				stopMagHdgFusion();
 				stopMag3DFusion();
 
-				ECL_INFO_TIMESTAMPED("commencing external vision yaw fusion");
+				ECL_INFO_TIMESTAMPED("commencing vision yaw fusion");
 			}
 		}
 
@@ -353,7 +353,7 @@ void Ekf::controlExternalVisionFusion()
 
 		// Turn off EV fusion mode if no data has been received
 		stopEvFusion();
-		ECL_INFO_TIMESTAMPED("External Vision Data Stopped");
+		ECL_INFO_TIMESTAMPED("Vision Data Stopped");
 
 	}
 }
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -202,14 +202,14 @@ void Ekf::controlExternalVisionFusion()
 				if (_params.fusion_mode & MASK_USE_EVPOS && !_control_status.flags.ev_pos) {
 					_control_status.flags.ev_pos = true;
 					resetPosition();
-					ECL_INFO_TIMESTAMPED("commencing vision position fusion");
+					ECL_INFO_TIMESTAMPED("starting vision pos fusion");
 				}
 
 				// turn on use of external vision measurements for velocity
 				if (_params.fusion_mode & MASK_USE_EVVEL && !_control_status.flags.ev_vel) {
 					_control_status.flags.ev_vel = true;
 					resetVelocity();
-					ECL_INFO_TIMESTAMPED("commencing vision velocity fusion");
+					ECL_INFO_TIMESTAMPED("starting vision vel fusion");
 				}
 			}
 		}
@@ -235,7 +235,7 @@ void Ekf::controlExternalVisionFusion()
 				stopMagHdgFusion();
 				stopMag3DFusion();
 
-				ECL_INFO_TIMESTAMPED("commencing vision yaw fusion");
+				ECL_INFO_TIMESTAMPED("starting vision yaw fusion");
 			}
 		}
 
@@ -353,7 +353,7 @@ void Ekf::controlExternalVisionFusion()
 
 		// Turn off EV fusion mode if no data has been received
 		stopEvFusion();
-		ECL_INFO_TIMESTAMPED("Vision Data Stopped");
+		ECL_INFO_TIMESTAMPED("vision data stopped");
 
 	}
 }
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -335,7 +335,7 @@ void Ekf::controlExternalVisionFusion()
 				}
 			}
 
-			ev_vel_obs_var = matrix::max(getVisionVelocityVarianceInEkfFrame(), sq(0.01f));
+			ev_vel_obs_var = matrix::max(getVisionVelocityVarianceInEkfFrame(), sq(0.05f));
 
 			ev_vel_innov_gates.setAll(fmaxf(_params.ev_vel_innov_gate, 1.0f));
 
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -324,22 +324,7 @@ void Ekf::controlExternalVisionFusion()
 			Vector3f ev_vel_obs_var;
 			Vector2f ev_vel_innov_gates;
 
-			Vector3f vel_aligned{_ev_sample_delayed.vel};
-			Matrix3f ev_vel_var = matrix::diag(_ev_sample_delayed.velVar);
-
-			// rotate measurement into correct earth frame if required
-			if (_params.fusion_mode & MASK_ROTATE_EV) {
-				vel_aligned = _R_ev_to_ekf * _ev_sample_delayed.vel;
-				ev_vel_var = _R_ev_to_ekf * ev_vel_var * _R_ev_to_ekf.transpose();
-			}
-
-			// correct velocity for offset relative to IMU
-			const Vector3f pos_offset_body = _params.ev_pos_body - _params.imu_pos_body;
-			const Vector3f vel_offset_body = _ang_rate_delayed_raw % pos_offset_body;
-			const Vector3f vel_offset_earth = _R_to_earth * vel_offset_body;
-			vel_aligned -= vel_offset_earth;
-
-			_ev_vel_innov = _state.vel - vel_aligned;
+			_ev_vel_innov = _state.vel - getVisionVelocityInEkfFrame();
 
 			// check if we have been deadreckoning too long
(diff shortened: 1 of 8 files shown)


jkflying:
> Velocity variance was lower bounded by a parameter on the firmware side. This is not possible any more. We would need to get this parameter from the firmware side to apply it in ECL after we transform the velocity covariance to local frame.

The real way to do this would be to clamp the norm. However for a covariance matrix this would require calculating SVD and clamping maximum component, which isn't realistic on embedded/realtime platforms. We could clamp maximum absolute value of any indiviual element, which would get us a limit of sqrt(2)*max, which is probably sufficient for this use case.

kamilritz:
> We can take the diagonal, figure out what we need to increase the scale by to hit the minimum for each element, then multiply rows and columns by sqrt of the scale vector. It will give us guarantee of minimum, and up to 1.7x the minimum.

Can you describe a bit more detailed how we should this? What do you mean by taking the diagonal?

jkflying:
Here's a sort of pythony pseudocode of the idea:
```python
def rescaleToAtLeast(covariance, min_variance):
  inv_d_sqrt = [max(1,sqrt(min_variance/abs(d))) for d in covariance.diag()]
  for i in range((covariance.cols()):
    for  j in range(covariance.rows()):
      covariance[i][j] *= inv_d_sqrt[i] * inv_d_sqrt[j]
```
This will scale the entire diagonal to at least min_variance, keeping the rows and columns proportionally weighted. If you want to account for off-diagonal elements, at the cost of a bit more computation you could instead calculate inv_d_sqrt something like:

```python
inv_d_sqrt = [max(1,sqrt(1/covariance.row(i).norm())) for i in range(covariance.rows())]
```

Make sure to test these formulas to make sure I haven't gotten something backwards :slightly_smiling_face:

kamilritz:
I now have played around with different options and I am not really convinced that this is the way to go. Specially if the covariance is rotate by 45 degrees. A covariance matrix like 
```
[ 0.1 0
  0     9]
```
becomes roughly
```
[4.5 -4.5
 -4.5 4.5/
``` 
.

From this without eigenvalue decomposition it is almost impossible to lower bound the original covariance matrix.

An alternative approach would be to forward the `EKF2_EVV_NOISE` and `EKF2_EV_NOISE_MD` parameters to ECL and apply it there. This would be effective, but the issue is that we would start apply limits in two different places.

mhkabir:
> We could also not apply an user specificied lower limit since in ECL there is a hardcoded lower limit enforced anyway. https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L383

jkflying:
For something this trivial (ie clamping) doing an eigenvalue decomposition isn't a suitable use of CPU time. I think we'll just need to trust the input from the EV system a bit more, and rely on the ECL clamping for extreme data rejection.

kamilritz:
NOTE: Last testflights after rebase showed strange behavior. But could also be the vehicle itself. Please fly test this again before merging.

EKF/ekf_helper.cpp
```diff
@@ -1513,6 +1510,47 @@ void Ekf::updateBaroHgtOffset()
...
 }
 
+Vector3f Ekf::getVisionVelocityInEkfFrame()
+{
+	// correct velocity for offset relative to IMU
+	const Vector3f pos_offset_body = _params.ev_pos_body - _params.imu_pos_body;
+	const Vector3f vel_offset_body = _ang_rate_delayed_raw % pos_offset_body;
+
+	// rotate measurement into correct earth frame if required
+	if (_ev_sample_delayed.vel_frame == BODY_FRAME_FRD)
```
jkflying, line 1520: Instead of if/else, could you rather use a `switch` on the enum for the different cases. This way if more get added in the future, it will get compilation errors until the extra cases are added, instead of silently failing.

EKF/ekf_helper.cpp
```diff
@@ -1513,6 +1510,47 @@ void Ekf::updateBaroHgtOffset()
...
+		return _ev_sample_delayed.vel - vel_offset_earth;
+	}
+}
+
+Vector3f Ekf::getVisionVelocityVarianceInEkfFrame()
+{
+	Matrix3f ev_vel_cov = _ev_sample_delayed.velCov;
+
+	// rotate measurement into correct earth frame if required
+	if (_ev_sample_delayed.vel_frame == BODY_FRAME_FRD)
```
jkflying, line 1540: same here

test/test_EKF_externalVision.cpp
```diff
@@ -204,3 +208,63 @@ TEST_F(EkfExternalVisionTest, visionAlignment)
...
+	const Matrix3f vel_cov_body(vel_cov_data);
+	const Vector3f vel_body(1.0f, 0.0f, 0.0f);
+	_sensor_simulator._vio.setVelocityCovariance(vel_cov_body);
+	_sensor_simulator._vio.setVelocity(vel_body);
+	_ekf_wrapper.enableExternalVisionVelocityFusion();
+	_sensor_simulator.startExternalVision();
+	_sensor_simulator.runSeconds(4);
+
+	const Vector3f velVar_new = _ekf->getVelocityVariance();
+	EXPECT_TRUE(velVar_new(1) > velVar_new(0));
```
jkflying, line 235: Rather than just saying "bigger", could you give some threshold/ratio you expect it to be bigger by, so this couldn't pass from random noise? Eg (not sure if these numbers are right, just a suggestion):
```suggestion
	EXPECT_TRUE(velVar_new(1) > 1.5f * velVar_new(0));
```

test/test_EKF_externalVision.cpp
```diff
@@ -204,3 +208,63 @@ TEST_F(EkfExternalVisionTest, visionAlignment)
...
+	const Matrix3f vel_cov_earth(vel_cov_data);
+	const Vector3f vel_earth(1.0f, 0.0f, 0.0f);
+	_sensor_simulator._vio.setVelocityCovariance(vel_cov_earth);
+	_sensor_simulator._vio.setVelocity(vel_earth);
+	_ekf_wrapper.enableExternalVisionVelocityFusion();
+	_sensor_simulator.startExternalVision();
+	_sensor_simulator.runSeconds(4);
+
+	const Vector3f velVar_new = _ekf->getVelocityVariance();
+	EXPECT_TRUE(velVar_new(0) > velVar_new(1));
```
jkflying, line 265: Same here

EKF/ekf_helper.cpp
kamilritz: Done

EKF/ekf_helper.cpp
kamilritz: done

test/test_EKF_externalVision.cpp
kamilritz: done

test/test_EKF_externalVision.cpp
kamilritz: done

kamilritz:
@jkflying  The lower bound on the vision velocity noise is removed from the ekf2_main file in the Firmware PR. This was necessary to pass on the full 3x3 covariance. The default lower bound was 0.1 m/s.
This change made the (now working) mavsdk integration tests fail on the iris_vision_velocity model. The noise for this model is set to 0.01 m/s. This was low enough for the estimator to come into trouble. 
There is an additional hardcoded lower bound for the vision observation noise inside the [control.cpp file](https://github.com/PX4/ecl/blob/master/EKF/control.cpp#L381). Increasing the lower bound here to 0.05 m/s is solving the issue.

jkflying:
That's annoying - I can certainly imagine use cases where we want that level of uncertainty. This is something to consider moving forward, especially if we are rethinking certain generated code sections.

@bresch I'm happy for this to be merged, if you would do the honours together with the Firmware PR?

EKF/ekf_helper.cpp
```diff
@@ -1491,6 +1488,49 @@ void Ekf::updateBaroHgtOffset()
...
+
+		case LOCAL_FRAME_FRD:
+			const Vector3f vel_offset_earth = _R_to_earth * vel_offset_body;
+			if(_params.fusion_mode & MASK_ROTATE_EV)
+			{
+				return _R_ev_to_ekf *_ev_sample_delayed.vel - vel_offset_earth;
+			}
+			return _ev_sample_delayed.vel - vel_offset_earth;
+	}
+	// switch statement guarantees to return
```
mhkabir, line 1510: Does it though? What if an invalid vel_frame is specified?

EKF/control.cpp
```diff
@@ -377,7 +362,7 @@ void Ekf::controlExternalVisionFusion()
 				}
 			}
 
-			ev_vel_obs_var = matrix::max(ev_vel_var.diag(), sq(0.01f));
+			ev_vel_obs_var = matrix::max(getVisionVelocityVarianceInEkfFrame(), sq(0.05f));
```
mhkabir, line 365: We mirror this change to the firmware side param lower bound. It is 0.01 right now, iirc.

EKF/ekf_helper.cpp
kamilritz: Yes. I checked. As long as there is no default case, the compiler will throw an error if there is a label missing or a label that does not belong to the enum.

EKF/control.cpp
kamilritz: Don't think so. At the moment it is [0.1 m/s](https://github.com/PX4/Firmware/blob/master/src/modules/ekf2/ekf2_params.c#L747). This is a value that gave good results with test on the Realsense T265. 
That this value here is lower is okay, since it is a fixed lower bound that should guarantee stability of the filter. The other is a tuning parameter.

mhkabir: I was referring to the lower bound in the metadata https://github.com/PX4/Firmware/blob/297e4dd5893128122109863ee79e30500e426553/src/modules/ekf2/ekf2_params.c#L743

kamilritz: oh, good point.

EKF/control.cpp
```diff
@@ -202,14 +202,14 @@ void Ekf::controlExternalVisionFusion()
 				if (_params.fusion_mode & MASK_USE_EVPOS && !_control_status.flags.ev_pos) {
 					_control_status.flags.ev_pos = true;
 					resetPosition();
-					ECL_INFO_TIMESTAMPED("commencing external vision position fusion");
+					ECL_INFO_TIMESTAMPED("commencing vision position fusion");
```
bresch, line 205: You could even shorten that to "Vision pos fusion started" or "starting vision pos fusion"

kamilritz: that is even better.

PR #709, opened 2020-01-13
Don't filter EV alignment

nicovanduijn:
This removes the IIR filter on the attitude offset between vision and EKF frame.
I don't see why this rotation would need to be low pass filtered. Especially if one of the two drifts, this leads to toilet-bowling.

Not sure if I did this right, so any review is appreciated.

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -1615,32 +1615,6 @@ void Ekf::calcExtVisRotMat()
 	// Calculate the quaternion delta that rotates from the EV to the EKF reference frame at the EKF fusion time horizon.
 	Quatf q_error = _state.quat_nominal * _ev_sample_delayed.quat.inversed();
 	q_error.normalize();
-
-	// convert to a delta angle and apply a spike and low pass filter
-	Vector3f rot_vec = q_error.to_axis_angle();
-
-	float rot_vec_norm = rot_vec.norm();
-
-	if (rot_vec_norm > 1e-6f) {
-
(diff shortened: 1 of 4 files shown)


EKF/ekf_helper.cpp
```diff
@@ -1615,32 +1615,6 @@ void Ekf::calcExtVisRotMat()
...
-
-		if (input_delta_len > 0.1f) {
-			rot_vec = _ev_rot_vec_filt + _input_delta_vec * (0.1f / input_delta_len);
-		}
-
-		// Apply a first order IIR low pass filter
-		const float omega_lpf_us = 0.2e-6f; // cutoff frequency in rad/uSec
-		float alpha = math::constrain(omega_lpf_us * (float)(_time_last_imu - _ev_rot_last_time_us), 0.0f, 1.0f);
-		_ev_rot_last_time_us = _time_last_imu;
-		_ev_rot_vec_filt = _ev_rot_vec_filt * (1.0f - alpha) + rot_vec * alpha;
```
kamilritz: _ev_rot_vec_filt should then be removed.

nicovanduijn:
Yeah, e.g. the log that triggered this change in the first place:
https://review.px4.io/plot_app?log=21099c0f-75f8-48b0-af7e-0071e0d50089

LorenzMeier:
I think you need to rebase and force-push.

bresch:
**Note for everyone**: I didn't merge that to ECL master but to a testing branch. We will open a PR to master if the tests show good results.

Issue #705, opened 2020-01-10
Stop mag heading fusion if yaw stays at +/-90 deg


kamilritz:
While I wrote a unit test, I noticed that when we set the orientation exactly to +/- 90 degrees that the heading uncertainty is increasing until it hits the limits. I followed it down to these [lines of code](https://github.com/PX4/ecl/blob/master/EKF/mag_fusion.cpp#L533-L537). Since in practice we will never stay exactly so close to +/- 90 degrees we will not always return before triggering the fusion. But if we do, we stop performing heading fusion.

@RomanBapst @priseborough FYI

mhkabir:
I wonder if this can also happen if you're using external vision which initializes to such an heading angle.

I still have't conclusively tracked down https://github.com/PX4/ecl/issues/698, and then there also was someone else reporting this: https://github.com/PX4/Firmware/issues/13116.

PR #710, opened 2020-01-14
Replace implementation of heading jacobian

kamilritz:
The previous computation of the heading jacobian had singularities at roll=0, pitch=0, yaw=+/-90 degrees. We therefore needed to stop the heading fusion in this orientation. This PR uses dual numbers to compute the jacobian with automatic differentiation. With this the singularity can be avoided.

**Testing:**
- unit testing shows no difference between the existing version and the proposed version except for previous singularity case.
- unit tests are failing at the moment to show how the proposed implementation is performing. Tests need to be removed before merging.
- SITL mission test: [log](https://review.px4.io/plot_app?log=1f97a874-6774-47ca-9c4d-deed7428962d)

**Issues:**
- This adds 700 Bytes of flash memory, since it needs to create multiple templated matrix functions with "Dual" as type.
- The automatic derivative approach seems to be 10 times slower. I profiled the test_EKF_jacobian.cpp file. The question is if this matters in relation to the rest of the load.
Screenshot from 2020-01-16 13-59-54

**Profiling:**
Previous
Screenshot from 2020-01-16 15-13-41

Proposed
Screenshot from 2020-01-16 15-13-29

Ekf::update()
Screenshot from 2020-01-16 15-18-09

https://github.com/PX4/ecl/issues/705

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -44,6 +44,8 @@
 
 #include "estimator_interface.h"
 
+typedef matrix::Dual<float, 4> Dual4f;
+
 class Ekf : public EstimatorInterface
 {
 public:
@@ -768,6 +770,10 @@ private:
 	void updateRangeDataContinuity();
 
 	bool isRangeDataContinuous() { return _dt_last_range_update_filt_us < 2e6f; }
+
+	Dual4f yaw321FromQuaterion(matrix::Quaternion<Dual4f> q) const;
+
+	Dual4f yaw312FromQuaterion(matrix::Quaternion<Dual4f> q) const;
 
 	// Increase the yaw error variance of the quaternions
 	// Argument is additional yaw variance in rad**2
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -44,8 +44,6 @@
 
 #include "estimator_interface.h"
 
-typedef matrix::Dual<float, 4> Dual4f;
-
 class Ekf : public EstimatorInterface
 {
 public:
@@ -771,9 +769,9 @@ private:
 
 	bool isRangeDataContinuous() { return _dt_last_range_update_filt_us < 2e6f; }
 
-	Dual4f yaw321FromQuaterion(matrix::Quaternion<Dual4f> q) const;
+	Dual4f yaw321FromQuaterion(const matrix::Quaternion<Dual4f>& q) const;
 
-	Dual4f yaw312FromQuaterion(matrix::Quaternion<Dual4f> q) const;
+	Dual4f yaw312FromQuaterion(const matrix::Quaternion<Dual4f>& q) const;
 
 	// Increase the yaw error variance of the quaternions
 	// Argument is additional yaw variance in rad**2
--- a/EKF/mag_fusion.cpp
+++ b/EKF/mag_fusion.cpp
@@ -44,6 +44,18 @@
 #include <ecl.h>
 #include <mathlib/mathlib.h>
 
+Dual4f Ekf::yaw321FromQuaterion(matrix::Quaternion<Dual4f> q) const
+{
+	return atan2(Dual4f(2) * (q(0) * q(3) + q(1) * q(2)),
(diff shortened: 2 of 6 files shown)


EKF/mag_fusion.cpp
```diff
@@ -44,6 +44,18 @@
 #include <ecl.h>
 #include <mathlib/mathlib.h>
 
+Dual4f Ekf::yaw321FromQuaterion(matrix::Quaternion<Dual4f> q) const
```
jkflying:
```suggestion
Dual4f Ekf::yaw321FromQuaterion(const matrix::Quaternion<Dual4f>& q) const
```

(and below too)

EKF/ekf.h
```diff
@@ -44,6 +44,8 @@
 
 #include "estimator_interface.h"
 
+typedef matrix::Dual<float, 4> Dual4f;
```
jkflying: I think there is a better place to put this - maybe `estimator_interface.h`

EKF/mag_fusion.cpp
```diff
@@ -44,6 +44,18 @@
...
+{
+	return atan2(Dual4f(2) * (q(0) * q(3) + q(1) * q(2)),
```
dagar:
```suggestion
	return atan2f(Dual4f(2) * (q(0) * q(3) + q(1) * q(2)),
```

On a board without double precision floating point (eg pixhawk 1, 2, 3, pixracer, etc) atan2 is an order of mangitude slower than atan2f.

``` Console
atan2(): 1000 events, 27911us elapsed, 27.91us avg, min 27us max 28us 0.284us rms
atan2f(): 1000 events, 2725us elapsed, 2.73us avg, min 2us max 3us 0.446us rms
```

http://ci.px4.io:8080/blue/organizations/jenkins/PX4_misc%2FFirmware-hardware/detail/master/1271/pipeline/788#step-800-log-361

jkflying: Good catch. @kamilritz this is probably the majority of the extra flash size you are seeing, too.

LorenzMeier:
Could you please rebase? If you have time: We should move CI to GH actions.

priseborough:
Is the automatic differentiation done internally in double precision? This PR definitely makes the code cleaner, but at what cost? The singularities it removes were not a problem in practice, but it would be good to eliminate them provided we don't give away too much in terms of clock cycles and flash.

jkflying:
> Is the automatic differentiation done internally in double precision?

No, it correctly uses whatever types it is passed. We need to check exactly what is happening here, but anything inside of Matrix that uses atan2 with floats should correctly call atan2f due to https://github.com/PX4/Matrix/blob/master/matrix/stdlib_imports.hpp

EKF/mag_fusion.cpp
kamilritz: Exactly

EKF/mag_fusion.cpp
kamilritz: Thanks

EKF/ekf.h
kamilritz: done

priseborough:
I could update the Matlab derivation script such that there is an alternative calculation with the singularity at a different orientation. The singularity is introduced by use of the atan function in the derivation of the observation Jacobian which for the 321 rotation sequence occurs when Tbn(1,1) approaches zero, eg 

https://github.com/PX4/ecl/blob/master/EKF/matlab/scripts/Inertial%20Nav%20EKF/GenerateNavFilterEquations.m#L394

For example defining the yaw in the 321 sequence derivation to be pi/2 - atan(Tbn(1,1)/Tbn(2,1)) moves the singularity from +- pi/2 to 0 and +- pi

I will submit a PR containing the updated derivation and code.

priseborough:
Closed now that singularity has been fixed in via alternative algebra see - https://github.com/PX4/ecl/pull/726

PR #717, opened 2020-01-17
Sensor Replay

kamilritz:
This PR adds the functionality to test the EKF with sensor data from ulog files. Therefore it includes following pieces:

- Python scripts that extract sensor data from ulog files and writes them to a csv file
- Extention to sensor_simulator that loads sensor data from csv file and replay the EKF with it.
- Simple logging of the EKF's state and covariance to csv file.
- First data set based on iris SITL flight
- CI action that checks if the logs of the test running with replay data is changing. This helps to see if there are functional changes.

**Attention: This will not simulate 1-to-1 an EKF replay as with `make px4_sitl none`, since the sequence of the measurements going into the buffer can differ!**

How to use it:

- create sensor_data.csv file from ulog file
```
cd test/sensor_simulator/
python3 createSensorDataFile.py <path/to/ulog> ../replay_data/<descriptive_name>.csv
```
- Setup the test file to use the EKF with the created sensor data by copy&paste an existing test case in `test/test_EKF_withReplayData.cpp` and adapt the paths to load the right sensor data and write it to the right place
```
_sensor_simulator.loadSensorDataFromFile("../../../test/replay_data/<descriptive_name>.csv");
_ekf_logger.setFilePath("../../../test/change_indication/<descriptive_name>.csv");
```
- You can feed the EKF with the data in the csv file, by running '_sensor_simulator.runReplaySeconds(duration_in_seconds)'. Be aware that replay sensor data will only be available when the corresponding sensor simulation are running. By default only imu, baro and mag sensor simulators are running. You can start a sensor simulation by calling `_sensor_simulator._<sensor>.start()`. Be also aware that you still have to setup the EKF yourself. This includes setting the bit mask (fusion_mode in common.h) according to what you intend to fuse. 
- In between `_sensor_simulator.runReplaySeconds(duration_in_seconds)` calls, you should write the state and covariances to the change_indication file by adding `_ekf_logger.writeStateToFile()`.

- Run the EKF with your data and all the other tests by running
```
make test
```
from the ecl directory

**Change indication** 
is the concept of running the EKF on different data-sets and compare the state of the EKF to a previous version. If a contributor makes a functional change that is run during the change_indication tests, this will produce a different output of the EKF's state. As the tests are run in CI, we will be able to check if the contributor forgot to run the checks themselves and add the new EKF's state outputs to the PR.

**Issue:** 
The change indication files as well as the sensor_data files will add a lot of lines to the repo. Does anyone see a good way to move this to external place?


dagar:
This looks nice, although the dependency on PX4/Firmware messages might be questionable. We try to keep the messages used for ekf2 replay reasonably stable, but they're still effectively internal data structures subject to change at any time.

It depends on the goal here, but one idea would be to use the ecl/EKF types for data storage, and then within PX4/Firmware carry a small script that reads a ulg (or csv) and dumps an ecl/ekf version of the data. I guess I'm basically describing the ekf2_main in script form...

priseborough:
Can you provide a log and brief list of steps so I can test this using a ulog.

priseborough:
I can generate my own logs for testing with some fallbacks and other reset behaviour.

kamilritz:
@priseborough  I added a guide on how to use it to the PR description

test/test_EKF_withReplayData.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "sensor_simulator/ekf_logger.h"
+
+class EkfReplayTest : public ::testing::Test {
+ public:
+	EkfReplayTest(): ::testing::Test(),
+	_ekf{std::make_shared<Ekf>()},
+	_sensor_simulator(_ekf),
+	_ekf_wrapper(_ekf),
+	_ekf_logger(_ekf) {};
+
```
priseborough: I would add a note here, eg:

// By default the IMU, Baro and Mag sensor simulators are already running sop no need to start them here

priseborough:
I was able to work through the instructions and generate reference data and demonstrate that functional changes (eg vertical position observation noise parameter) resulted in detectable differences.

PR #719, opened 2020-01-20
Updating optical flow interface

kamilritz:
Previously we had two different types that can store optical flow data: `flowSample` and `flow_message`. The only difference is that flow_message has no time_us field, since the time was passed separately to `setOpticalFlowDate(time, flow_message)`.
This PR removes the `flow_message` and changes the function to `setOpticalFlowData(flow_sample)`. 

Previously we also had some logic in place to only pop flowSamples from the buffer if the previous one was used. We waited for using the flow_sample until the midpoint of the flow sample was fallen behind the fusion horizon. I argue that we should compensate for half of the flow dt before sending it to the EKF. Therefore every time we pop a flow sample from the buffer it is actually time to use it. As it is done for all the other sensors.

**Firmware PR:** https://github.com/PX4/Firmware/pull/13988

**Testing:**
SITL log replay shows a small diff only visible if toggling between the logs

BEFORE:
Screenshot from 2020-01-20 14-17-04
[Log](https://review.px4.io/plot_app?log=8bcc5994-677f-4fdc-842e-554d303556ba)

AFTER:
Screenshot from 2020-01-20 14-17-39
[Log](https://review.px4.io/plot_app?log=aef16413-f24e-4840-b266-4cb8d6c8a2ea)

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -73,13 +73,6 @@ struct gps_message {
 	float pdop;		///< position dilution of precision
 };
 
-struct flow_message {
-	uint8_t quality;	///< Quality of Flow data
-	Vector2f flowdata;	///< Optical flow rates about the X and Y body axes (rad/sec)
-	Vector3f gyrodata;	///< Gyro rates about the XYZ body axes (rad/sec)
-	uint32_t dt;		///< integration time of flow samples (microseconds)
-};
-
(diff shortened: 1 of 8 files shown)


kamilritz:
Moved to PR#720

PR #720, opened 2020-01-20
[DO NOT MERGE] Update optical flow interface

kamilritz:
Previously we had two different types that can store optical flow data: `flowSample` and `flow_message`. The only difference is that flow_message has no time_us field, since the time was passed separately to `setOpticalFlowDate(time, flow_message)`.
This PR removes the `flow_message` and changes the function to `setOpticalFlowData(flow_sample)`. 

Previously we also had some logic in place to only pop flowSamples from the buffer if the previous one was used. We waited for using the flow_sample until the midpoint of the flow sample was fallen behind the fusion horizon. I argue that we should compensate for half of the flow dt before sending it to the EKF. Therefore every time we pop a flow sample from the buffer it is actually time to use it. As it is done for all the other sensors.

**Firmware PR:**  https://github.com/PX4/Firmware/pull/13988

**Testing:**
SITL log replay shows a small diff only visible if toggling between the logs

BEFORE:
Screenshot from 2020-01-20 14-17-04
[Log](https://review.px4.io/plot_app?log=8bcc5994-677f-4fdc-842e-554d303556ba)

AFTER:
Screenshot from 2020-01-20 14-17-39
[Log](https://review.px4.io/plot_app?log=aef16413-f24e-4840-b266-4cb8d6c8a2ea)

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -323,7 +323,6 @@ void EstimatorInterface::setRangeData(uint64_t time_usec, float data, int8_t qua
 	}
 }
 
-// TODO: Change pointer to constant reference
 void EstimatorInterface::setOpticalFlowData(const flowSample& flow)
 {
 	if (!_initialised || _flow_buffer_fail) {
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -324,7 +324,7 @@ void EstimatorInterface::setRangeData(uint64_t time_usec, float data, int8_t qua
 }
 
 // TODO: Change pointer to constant reference
-void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *flow)
+void EstimatorInterface::setOpticalFlowData(const flowSample& flow)
 {
 	if (!_initialised || _flow_buffer_fail) {
 		return;
@@ -342,10 +342,10 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
 	}
 
 	// limit data rate to prevent data being lost
-	if ((time_usec - _time_last_optflow) > _min_obs_interval_us) {
+	if ((flow.time_us - _time_last_optflow) > _min_obs_interval_us) {
 		// check if enough integration time and fail if integration time is less than 50%
 		// of min arrival interval because too much data is being lost
-		float delta_time = 1e-6f * (float)flow->dt; // in seconds
+		float delta_time = flow.dt; // in seconds
 		const float delta_time_min = 0.5e-6f * (float)_min_obs_interval_us;
 		bool delta_time_good = delta_time >= delta_time_min;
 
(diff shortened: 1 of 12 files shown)


test/sensor_simulator/flow.cpp
```diff
@@ -16,22 +16,24 @@ Flow::~Flow()
...
-flow_message Flow::dataAtRest()
+flowSample Flow::dataAtRest()
 {
-	flow_message _flow_at_rest;
-	_flow_at_rest.dt = 20000;
-	_flow_at_rest.flowdata = Vector2f{0.0f, 0.0f};
-	_flow_at_rest.gyrodata = Vector3f{0.0f, 0.0f, 0.0f};
+	flowSample _flow_at_rest;
+	_flow_at_rest.dt = 0.02f;
+	_flow_at_rest.flowRadXY = Vector2f{0.0f, 0.0f};
```
bresch: `flow_rad_xy`

test/sensor_simulator/flow.cpp
```diff
@@ -16,22 +16,24 @@ Flow::~Flow()
...
+	_flow_at_rest.gyroXYZ = Vector3f{0.0f, 0.0f, 0.0f};
```
bresch: `gyro_xyz`

EKF/estimator_interface.cpp
```diff
@@ -364,25 +363,22 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
...
 
-			optflow_sample_new.quality = flow->quality;
+			_time_last_optflow = flow.time_us;
 
-			// NOTE: the EKF uses the reverse sign convention to the flow sensor. EKF assumes positive LOS rate
-			// is produced by a RH rotation of the image about the sensor axis.
-			optflow_sample_new.gyroXYZ = - flow->gyrodata;
-			optflow_sample_new.flowRadXY = -flow->flowdata;
+			flowSample optflow_sample_new = flow;
+
```
bresch: where do you take care about the change of convention?

kamilritz: On the firmware side. If the client of ECL has a different convention, he should be responsible for adapting.

test/sensor_simulator/flow.cpp
kamilritz: Are you happy with flow_rad_xy or should the name be different?

bresch: I think flow_xy_rad is even better :D

kamilritz: Updated to `flow_xy_rad`

test/sensor_simulator/flow.cpp
kamilritz: Updated to `gyro_xyz`

EKF/estimator_interface.cpp
bresch: Makes sense. Can you link the corresponding firmware PR please?

kamilritz:
Changing the timestamp of the flow sample to the midpoint of the measurement periode, will likely cause trouble with the accumulation of delta_angles for flow sensors without built-in IMU.

kamilritz:
Moved to https://github.com/PX4/ecl/pull/733

PR #722, opened 2020-01-21
Auto-formatting

kamilritz:
This PR brings auto-formatting to ECL. It adds a few Makefile targets, which provide installation of astyle, checking the format and formatting the code. 

The formatting rules are copied from Firmware and are specified in `astyle/astylerc`. The files on which the auto-formatting is required, is specified in `astyle/format.sh`. At the moment it is only added for `EKF/RingBuffer.h` and `EKF/AlphaFilter.hpp`

A github action that checks if the format is according to the rules is also added.

### DIFF
--- a/.github/workflows/format_checks.yml
+++ b/.github/workflows/format_checks.yml
@@ -1,0 +1,12 @@
+name: Format Checks
+
+on: [push]
+
+jobs:
+  format_checks:
+    runs-on: ubuntu-latest
+    container: px4io/px4-dev-base-bionic:2020-01-13
+    steps:
+    - uses: actions/checkout@v1
+    - name: make check_format
+      run: make check_format
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 
 .cache/
 build/
+astyle/*/*
+astyle/*.txt
 
 */*.a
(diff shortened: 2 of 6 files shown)


mhkabir:
Any chance we could adopt clang-format here from the beginning, instead of buggy astyle?

kamilritz:
@mhkabir @jkflying I am open to use clang-format. I never really used it. If you can point me to a good reference, I might get motivated to change it. 
But I always thought that there is a lot of trouble with clang-format to get the right version for your OS.

jkflying:
From what I've seen with clang-format it is always possible to install the apt package on 'wrong' versions since they don't have dependencies.

dagar:
I think it's definitely the better tool, but might not be worth the effort unless someone is interested in generating an appropriate config.

The bigger problem I see is that we can't format the majority of ecl/EKF without making a mess of all the generated code.

kamilritz:
There are more and more files that do not contain auto-generated code. It would be good to enforce a style there.

kamilritz:
Which clang-format version would you suggest?

kamilritz:
@mhkabir @jkflying https://github.com/PX4/ecl/pull/727

PR #723, opened 2020-01-21
Updates to interface

kamilritz:
This PR contains several changes to the interface of ECL. The main goal is to standardize the interface. This is done by forcing all `set<Sensor>Data()` function to accept a `<Sensor>Sample`. Except from standardization this has the advantage that the pretty good documentation of the `<Sensor>Samples` function as part of the documentation of the interface.

Small remarks;
`gps_message` definition is adapted to profit from matrix:Vector
A legacy IMU interface from 2018 is also removed.

**Related Firmware PR:** https://github.com/PX4/Firmware/pull/13999

### DIFF
--- a/test/sensor_simulator/mag.cpp
+++ b/test/sensor_simulator/mag.cpp
@@ -15,9 +15,10 @@ Mag::~Mag()
 
 void Mag::send(uint64_t time)
 {
-	float mag[3];
-	_mag_data.copyTo(mag);
-	_ekf->setMagData(time,mag);
+	magSample mag_sample;
+	mag_sample.mag = _mag_data;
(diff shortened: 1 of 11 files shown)


test/sensor_simulator/mag.cpp
```diff
@@ -15,9 +15,10 @@ Mag::~Mag()
 
 void Mag::send(uint64_t time)
 {
-	float mag[3];
-	_mag_data.copyTo(mag);
-	_ekf->setMagData(time,mag);
+	magSample mag_sample;
```
bresch: You could do that on a single lone like you did with the baro

PR #726, opened 2020-01-22
EKF: Rework yaw fusion to remove singularities blocking fusion at +-90 deg yaw

priseborough:
Addresses issue identified by https://github.com/PX4/ecl/pull/710, but solves the problem by re-working the derivation to provide an alternate computational path that enable the singularity in the Jacobian calculation at +-90 deg yaw to be bypassed using a calculation that moves the singularity to 0 and +- 180 deg. The calculation with the largest denominator in the inverse calculation is used.

Testing involved slow 360 deg yaw about 6 orientations (Z down, X up, Y up, Y down, X down, Z up)

Screen Shot 2020-01-22 at 7 48 21 pm

Test log here: https://logs.px4.io/plot_app?log=5c59627b-be9a-4018-abbb-4cad0b4efc0a

Plot of yaw innovations from test log:

Screen Shot 2020-01-22 at 7 49 44 pm

The log was replayed using the same branch illustrating that the innovation sequence differences to the original test data were within expected limits

Screen Shot 2020-01-22 at 7 49 27 pm

The replay was then repeated using https://github.com/PX4/ecl/pull/710 with changes rebased on ecl master. The innovation sequence was visually identical.

Screen Shot 2020-01-22 at 7 49 27 pm

This PR produces identical results to https://github.com/PX4/ecl/pull/710 and is expected to do so with minimal effect on resource utilisation. 

TODO complete SITL profiling test.......

Update on profiling. I was unable to run SITL using callgrind/valgrind due to the 'brew install valgrind' not able to run on OSX versions newer than High Sierra.


priseborough:
> Do we need to update the [computation of the yaw](https://github.com/PX4/ecl/blob/148707951c70f82dbcdf20fcd53c918fc34d8c31/EKF/mag_fusion.cpp#L637) angle as well?

No, because it is using an atan2 function. The derivation was forced to use atan becausue the symbolic toolbox cannot evaluate a Jacobian for the atan2

PR #727, opened 2020-01-22
Auto Formatting with clang-format-6.0

kamilritz:
This PR brings auto-formatting with clang-format-6.0 to ECL. It adds a few Makefile targets, which provide installation of clang-format, checking the format and formatting the code.

The formatting rules are based on google style guide. Except for some indentation modifcation to stay closer to appearance of the Firmware. The following options were adapted:
```
AccessModifierOffset: -4
IndentWidth: 4
TabWidth: 4
UseTab: Always
```

The files on which the auto-formatting is required, is specified in tools/format.sh. At the moment it is only added for EKF/RingBuffer.h and EKF/AlphaFilter.hpp

A github action that checks if the format is according to the rules is also added.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -44,7 +44,7 @@ BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeColon
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit:     80
+ColumnLimit:     120
 CommentPragmas:  '^ IWYU pragma:'
 CompactNamespaces: false
 ConstructorInitializerAllOnOneLineOrOnePerLine: true
--- a/.clang-format
+++ b/.clang-format
@@ -1,7 +1,7 @@
 ---
 Language:        Cpp
 # BasedOnStyle:  Google
-AccessModifierOffset: -4 # Modified
+AccessModifierOffset: -8 # Modified
 AlignAfterOpenBracket: Align
 AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false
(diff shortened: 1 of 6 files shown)


.clang-format
```diff
@@ -0,0 +1,117 @@
...
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Attach
+BreakBeforeInheritanceComma: false
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     80
```
dagar:
```suggestion
ColumnLimit:     120
```

80 is more common, but we allow up to 120 in PX4.

.clang-format
```diff
@@ -0,0 +1,117 @@
...
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 2
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Auto
+TabWidth:        4 # Modified
```
dagar:
```suggestion
TabWidth:        8 # Modified
```

Not necessarily the "right" answer, but again if PX4/Firmware consistency is the goal.

.clang-format
kamilritz: I know. I propose that we start with 80. I will check how the nested code in control.cpp is dealing with this

kamilritz: I know. I propose that we start with 80. I will check how the nested code in control.cpp is dealing with this

.clang-format
kamilritz: Also true. I aimed for similar appearance with a bit more reading comfort. I know this can be very subjective.

priseborough: My preference would be commonality with PX4 on tabwidth to avoid issues with editor settings.

.clang-format
jkflying: ColumnLimit of 80 can often results in people using hard to read abbreviations instead of the class/function/variable naming which would be most clear. ColumnLimit of 120 IMO is a much better tradeoff.

PR #729, opened 2020-01-23
Robustify timestamp checks

### LINKED ISSUE #459: Timestamp comparison issue

kamilritz:
This PR solves the open issue on timestamp checks: fixes #459
The issue was previously addressed by lower-bounding the timestamp of incoming GPS and baro samples. This can potentially be dangerous as rather old data could be re-timestamped and still be used. By removing these lower bounds old data would enter the buffer as it is and would be rejected by the Ringbuffer [checks](https://github.com/PX4/ecl/blob/master/EKF/RingBuffer.h#L133).

The issue of comparing unsigned integers is solved by avoiding the subraction. e.g:
```
_time_last_imu > _time_last_gps + (uint64_t)10e6
```
There should be also no issue with overflows under the assumption that the interval is rather small

**Testing:**
- Added unit tests for starting and timing out major sensors
- SITL flight [log](https://review.px4.io/plot_app?log=06199d04-de39-48af-b8e4-b9afb6dd76ef)

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -86,10 +86,10 @@ void Ekf::controlFusionModes()
 
 	// check for intermittent data (before pop_first_older_than)
 	const baroSample &baro_init = _baro_buffer.get_newest();
-	_baro_hgt_faulty = !((_time_last_imu - baro_init.time_us) < 2 * BARO_MAX_INTERVAL);
+	_baro_hgt_faulty = !isRecent(baro_init.time_us, 2 * BARO_MAX_INTERVAL);
 
 	const gpsSample &gps_init = _gps_buffer.get_newest();
-	_gps_hgt_intermittent = !((_time_last_imu - gps_init.time_us) < 2 * GPS_MAX_INTERVAL);
+	_gps_hgt_intermittent = !isRecent(gps_init.time_us, 2 * GPS_MAX_INTERVAL);
 
 	// check for arrival of new sensor data at the fusion time horizon
 	_gps_data_ready = _gps_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_gps_sample_delayed);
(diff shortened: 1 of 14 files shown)


EKF/control.cpp
```diff
@@ -332,9 +332,10 @@ void Ekf::controlExternalVisionFusion()
 				ev_pos_obs_var(1) = fmaxf(ev_pos_var(1, 0), sq(0.01f));
 
 				// check if we have been deadreckoning too long
-				if ((_time_last_imu - _time_last_hor_pos_fuse) > _params.reset_timeout_max) {
-					// don't reset velocity if we have another source of aiding constraining it
-					if (((_time_last_imu - _time_last_of_fuse) > (uint64_t)1E6) && ((_time_last_imu - _time_last_hor_vel_fuse) > (uint64_t)1E6)) {
+				if (isTimedOut(_time_last_hor_pos_fuse, _params.reset_timeout_max)) {
+					// only velocity if we have no another source of aiding constraining it
```
RomanBapst:
```suggestion
					// reset only velocity if we have no another source of aiding constraining it
```

kamilritz:
Unit tests only. All the major fusion starts and timeouts are covered by unit tests.

RomanBapst:
@priseborough I reviewed this PR line by line and did not spot any problem.
Are you ok with the current testing evidence (unit tests & SITL flight log) or do you think we should do more?

EKF/ekf_helper.cpp
```diff
@@ -1748,6 +1750,15 @@ float Ekf::kahanSummation(float sum_previous, float input, float &accumulator) c
 	return t;
 }
 
+bool Ekf::isTimedOut(uint64_t last_sensor_timestamp, uint64_t timeout_period)
+{
+	return last_sensor_timestamp + timeout_period < _time_last_imu;
```
bkueng: Can you make sure this (and below) gets inlined by moving the definition to the declaration?

kamilritz: yes, done

EKF/ekf.h
```diff
@@ -785,6 +785,10 @@ class Ekf : public EstimatorInterface
 	// sensor measurement
 	float calculate_synthetic_mag_z_measurement(const Vector3f& mag_meas, const Vector3f& mag_earth_predicted);
 
+	bool isTimedOut(uint64_t last_sensor_timestamp, uint64_t timeout_period);
```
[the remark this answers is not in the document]
kamilritz: Thanks

PR #730, opened 2020-01-23
Expanding auto-formatting

kamilritz:
Expanding auto-formatting to folders `mathlib` and `validation` and files `EKF/vel_pos_fusion.cpp` and `imu_down_sampler.*pp`

Style changes: 
- Increase penalty on breaking line after an assignment
- Add opening brackets for functions on new line.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -76,7 +76,7 @@ IndentWidth:     8 # Modified
 IndentWrappedFunctionNames: false
 JavaScriptQuotes: Leave
 JavaScriptWrapImports: true
-KeepEmptyLinesAtTheStartOfBlocks: false
+KeepEmptyLinesAtTheStartOfBlocks: true # Modified
 MacroBlockBegin: ''
 MacroBlockEnd:   ''
 MaxEmptyLinesToKeep: 1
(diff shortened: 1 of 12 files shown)


.clang-format
```diff
@@ -37,7 +37,7 @@ BraceWrapping:
   SplitEmptyRecord: true
   SplitEmptyNamespace: true
 BreakBeforeBinaryOperators: None
-BreakBeforeBraces: Attach
+BreakBeforeBraces: Custom # Modified
```
jkflying: This seems really inconsistent now. Is there a reason you made it custom?

kamilritz: Yes. To allow breaking before braces after functions. These breaks are widely used in the current code bode.

EKF/RingBuffer.h
```diff
@@ -86,12 +88,14 @@ class RingBuffer {
 		return true;
 	}
 
-	void unallocate() {
+	void unallocate()
+	{
```
dagar: This one actually stands out as being different than PX4.

dagar: From the google style guide.

 - "The open curly brace is always on the end of the last line of the function declaration, not the start of the next line."
 - "The close curly brace is either on the last line by itself or on the same line as the open curly brac"

kamilritz: Ok, you got me. But let me add an empty line after the function header.

PR #733, opened 2020-01-23
Update flow interface

kamilritz:


EKF/control.cpp
```diff
@@ -133,8 +133,9 @@ void Ekf::controlFusionModes()
 	}
 
 	// We don't fuse flow data immediately because we have to wait for the mid integration point to fall behind the fusion time horizon.
-	// This means we stop looking for new data until the old data has been fused.
-	if (!_flow_data_ready) {
+	// This means we stop looking for new data until the old data has been fused, unless we are not fusing optical flow,
+	// in this case we need to empty the buffer
+	if (!_flow_data_ready || !_control_status.flags.opt_flow) {
```
RomanBapst: @kamilritz I don't understand this anymore. When we add the flow sample to the ringbuffer, why don't we make sure that the timestamp of that sample is selected such that it is located in the middle of the integration period. Then you can just pop that sample from the buffer and fuse it right away.

kamilritz: For flow sensor without IMU, we need to accumulate the delta_angle by integrating the stand alone imu. When would put them in the buffer with the middle point timestamp, the integration would be harder. We want to move the integration to the flow driver level. After such a change we can push them in the buffer with middle point timestamp.

RomanBapst: @kamilritz Fair enough. I was hoping to be able to get rid of that logic in control.cpp but for now let's leave it.

.github/workflows/firmware_build_test.yml
```diff
@@ -16,7 +16,11 @@ jobs:
         repository: PX4/Firmware
         ref: master
         submodules: recurvise
-    - run: git submodule update --init --recursive
+    - name: Checkout matching branch on PX4/Firmware if possible
```
RomanBapst: @kamilritz How is this related to this PR?

EKF/estimator_interface.cpp
```diff
@@ -377,25 +376,22 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
...
-			optflow_sample_new.quality = flow->quality;
+			_time_last_optflow = flow.time_us;
 
-			// NOTE: the EKF uses the reverse sign convention to the flow sensor. EKF assumes positive LOS rate
-			// is produced by a RH rotation of the image about the sensor axis.
-			optflow_sample_new.gyroXYZ = - flow->gyrodata;
-			optflow_sample_new.flowRadXY = -flow->flowdata;
+			flowSample optflow_sample_new = flow;
+
+			// compensate time-stamp for the trailing edge of the flow data integration period
```
RomanBapst: @kamilritz This comment does not make sense to me. We compensate for measurement delay and for the average filter update period.

.github/workflows/firmware_build_test.yml
kamilritz: This PR is breaking the interface to Firmware. If you provide a branch with the same name on upstream, it will use it in the Firmware Build tests. That is the reason why all the checks are passing.

EKF/estimator_interface.cpp
kamilritz: Also not to me. It was there before. We can get rid of it.

RomanBapst:
For testing evidence, please see the related[ Firmware PR](https://github.com/PX4/Firmware/pull/13988)

PR #734, opened 2020-01-24
Ekf: use helper functions to centralize the checks of horizontal aiding

bresch:
I had enough of reading those:
`_control_status.flags.opt_flow && !(_control_status.flags.gps || _control_status.flags.ev_pos || _control_status.flags.ev_vel)` `_control_status.flags.opt_flow && !_control_status.flags.gps && !_control_status.flags.ev_pos && !_control_status.flags.ev_vel)`

The functions are:
- isOnlyActiveSourceOfHorizontalAiding(flag) : if the flag passed is set but not the other ones
- isOtherSourceOfHorizontalAiding(flag) : if an other aiding is active than the one passed
   for example: if `opt_flow` is true, all the following calls will return `true` regardless of their own state:
   - isOtherSourceOfHorizontalAiding(_control_status.flags.gps)
   - isOtherSourceOfHorizontalAiding(_control_status.flags.ev_vel)
   - isOtherSourceOfHorizontalAiding(_control_status.flags.ev_pos)
- isHorizontalAidingActive() : if one or more flags are true

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -375,7 +375,7 @@ void EstimatorInterface::setOpticalFlowData(uint64_t time_usec, flow_message *fl
 			delta_time = delta_time_min;
 		}
 
-		const bool relying_on_flow = !_control_status.flags.gps && !_control_status.flags.ev_pos && !_control_status.flags.ev_vel;
+		const bool relying_on_flow = !isOtherSourceOfHorizontalAidingThan(_control_status.flags.opt_flow);
 
 		const bool flow_quality_good = (flow->quality >= _params.flow_qual_min);
 
@@ -574,6 +574,30 @@ bool EstimatorInterface::local_position_is_valid()
 {
 	// return true if we are not doing unconstrained free inertial navigation
 	return !_deadreckon_time_exceeded;
+}
+
+bool EstimatorInterface::isOnlyActiveSourceOfHorizontalAiding(const bool aiding_flag) const
+{
+	return aiding_flag && !isOtherSourceOfHorizontalAidingThan(aiding_flag);
+}
+
+bool EstimatorInterface::isOtherSourceOfHorizontalAidingThan(const bool aiding_flag) const
+{
+	const int nb_sources = getNumberOfActiveHorizontalAidingSources();
+	return aiding_flag ? nb_sources > 1 : nb_sources > 0;
+}
+
+int EstimatorInterface::getNumberOfActiveHorizontalAidingSources() const
+{
+	return int(_control_status.flags.gps)
+	+ int(_control_status.flags.opt_flow)
+	+ int(_control_status.flags.ev_pos)
+	+ int(_control_status.flags.ev_vel);
+}
+
+bool EstimatorInterface::isHorizontalAidingActive() const
+{
+	return getNumberOfActiveHorizontalAidingSources() > 0;
 }
 
 void EstimatorInterface::printBufferAllocationFailed(const char * buffer_name)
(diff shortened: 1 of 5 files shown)


EKF/estimator_interface.cpp
```diff
@@ -576,6 +576,30 @@ bool EstimatorInterface::local_position_is_valid()
...
+	return aiding_flag && !isOtherSourceOfHorizontalAiding(aiding_flag);
+}
+
+bool EstimatorInterface::isOtherSourceOfHorizontalAiding(const bool aiding_flag) const
+{
+	const int nb_sources = getNumberOfActiveHorizontalAidingSources();
+	return aiding_flag ? nb_sources > 1 : nb_sources > 0;
+}
+
+int EstimatorInterface::getNumberOfActiveHorizontalAidingSources() const
```
kamilritz: Auxiliary Velocity would also be an horizontal aiding source, tough it is not actively used at the moment.

EKF/estimator_interface.h
```diff
@@ -252,6 +252,23 @@ class EstimatorInterface
 	// return true if the global position estimate is valid
 	virtual bool global_position_is_valid() = 0;
 
+	// Return true if this horizontal aiding source is active and if it is the only one
```
kamilritz:
```suggestion
	// Return true if passed horizontal aiding source is the only active one
```

kamilritz: But the function name would also be explanatory enough

kamilritz: Or you could also list the viable flags.

EKF/estimator_interface.h
```diff
@@ -252,6 +252,23 @@ class EstimatorInterface
...
+	bool isOnlyActiveSourceOfHorizontalAiding(bool aiding_flag) const;
+
+	/*
+	 * Check if there are any other active source of horizontal aiding
+	 * Warning: does not tell if the selected source is
+	 * active, use isOnlyActiveSourceOfHorizontalAiding() for this
+	 * @param aiding_flag a flag in _control_status.flags
+	 * @return true if an other source than aiding_flag is active
+	 */
+	bool isOtherSourceOfHorizontalAiding(bool aiding_flag) const;
```
kamilritz:
```suggestion
	bool isOtherSourceOfHorizontalAidingActiveThen(bool aiding_flag) const;
```

EKF/estimator_interface.cpp
bresch: Yes, the good thing is that we'll just have to add it here instead of in 20 places

EKF/estimator_interface.cpp
```diff
@@ -576,6 +576,30 @@ bool EstimatorInterface::local_position_is_valid()
...
+
+bool EstimatorInterface::isOtherSourceOfHorizontalAidingThan(const bool aiding_flag) const
+{
+	const int nb_sources = getNumberOfActiveHorizontalAidingSources();
+	return aiding_flag ? nb_sources > 1 : nb_sources > 0;
+}
+
+int EstimatorInterface::getNumberOfActiveHorizontalAidingSources() const
+{
+	return (int)_control_status.flags.gps
```
mhkabir:
```suggestion
	return int(_control_status.flags.gps)
```

mhkabir: (same for below)

bresch:
To be sure that the logic works, I made a STIL test where I disable GPS in air while having optical flow and an other one without optical flow: 
I had in the first case "GPS data stopped, using only EV or OF" and in the second case "GPS data stopped", which means that the check still works properly.

PR #737, opened 2020-01-25
EKF control minor cleanup (use switch statement and fix code style)

dagar:
I was going to take a look at aiding range aid for vision height mode, but thought I'd get this out of the way first.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -932,8 +932,12 @@ void Ekf::controlHeightFusion()
 	checkRangeAidSuitability();
 	_range_aid_mode_selected = (_params.range_aid == 1) && isRangeAidSuitable();
 
-	if (_params.vdist_sensor_type == VDIST_SENSOR_BARO) {
+	switch (_params.vdist_sensor_type) {
+	default:
+		ECL_ERR("Invalid height mode: %d", _params.vdist_sensor_type);
 
+	// FALLTHROUGH
+	case VDIST_SENSOR_BARO:
 		if (_range_aid_mode_selected && _range_data_ready && _rng_hgt_valid) {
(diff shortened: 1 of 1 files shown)


EKF/control.cpp
```diff
@@ -931,8 +931,11 @@ void Ekf::controlHeightFusion()
 	checkRangeAidSuitability();
 	_range_aid_mode_selected = (_params.range_aid == 1) && isRangeAidSuitable();
 
-	if (_params.vdist_sensor_type == VDIST_SENSOR_BARO) {
-
+	switch (_params.vdist_sensor_type) {
+	default:
```
kamilritz: What is the default case added for? Why the fallthrough?

EKF/control.cpp
```diff
@@ -976,10 +979,9 @@ void Ekf::controlHeightFusion()
 				_hgt_sensor_offset = _gps_sample_delayed.hgt - _gps_alt_ref + _state.pos(2);
 			}
 		}
-	}
+		break;
 
-	// set the height data source to range if requested
-	if ((_params.vdist_sensor_type == VDIST_SENSOR_RANGE) && _rng_hgt_valid) {
```
kamilritz: The check for the `_rng_hgt_valid` is missing now

EKF/control.cpp
dagar: I thought it couldn't hurt if someone entered a bogus HGT_MODE.

EKF/control.cpp
dagar: Oops, nice catch. Fixed.

I also changed it to _range_data_ready && _rng_hgt_valid for consistency.

PR #738, opened 2020-01-28
Hot flow fusion fix

kamilritz:
Fix to bug introduced in https://github.com/PX4/ecl/pull/729


RomanBapst:
@kamilritz How did we miss this? Well, I guess we are just humans.

PR #739, opened 2020-01-30
Fix bug causing incorrect covariance initialisation

priseborough:
This issue was found by @RomanBapst during a git bisect search to find the cause of a growing Z delta velocity bias estimate and vertical position and velocity innovations

Screen Shot 2020-01-30 at 5 52 45 pm

Screen Shot 2020-01-30 at 5 53 55 pm

that had appeared with a hardware setup when the firmware was updated. This search determined that the issue was introduced by https://github.com/PX4/ecl/pull/689/commits/ea3f95a56486a274cc71b549176de8364eea40d5 in a PR that was supposed to be a non functional change.

Because the covariance initialisation is dependent on the quaternion states, the Ekf::initialiseCovariance() function must be called after the states are aligned, not before.

It is also unnecessary to call Ekf::initialiseCovariance() in the Ekf::reset() function because this function sets _filter_initialised to false which causes the Ekf::initialiseFilter() function to be called until the state vector  (and with this change the covariance matrix ) has been reset.

The default value for delta velocity process noise has also been increased to fix poor stability in delta velocity bias estimation with seen in a log.

After these changes, we get the following results with replay of the same log:

Screen Shot 2020-01-30 at 5 56 50 pm

Screen Shot 2020-01-30 at 5 56 25 pm

Without the process noise change, the initial bias estimation is poor but fixes when flight commences, eg:

Screen Shot 2020-01-30 at 5 58 38 pm

An upstream PR is required to change the default value used for this parameter.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -223,7 +223,7 @@ struct parameters {
 
 	// process noise
 	float gyro_bias_p_noise{1.0e-3f};	///< process noise for IMU rate gyro bias prediction (rad/sec**2)
-	float accel_bias_p_noise{6.0e-3f};	///< process noise for IMU accelerometer bias prediction (m/sec**3)
+	float accel_bias_p_noise{1.0e-2f};	///< process noise for IMU accelerometer bias prediction (m/sec**3)
 	float mage_p_noise{1.0e-3f};		///< process noise for earth magnetic field prediction (Gauss/sec)
 	float magb_p_noise{1.0e-4f};		///< process noise for body magnetic field prediction (Gauss/sec)
 	float wind_vel_p_noise{1.0e-1f};	///< process noise for wind velocity prediction (m/sec**2)
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -67,8 +67,6 @@ void Ekf::reset()
 	_output_new.pos.setZero();
 	_output_new.quat_nominal.setIdentity();
 
-	initialiseCovariance();
-
 	_delta_angle_corr.setZero();
 
 	_imu_updated = false;
@@ -195,7 +193,6 @@ bool Ekf::initialiseFilter()
 		// calculate the initial magnetic field and yaw alignment
 		_control_status.flags.yaw_align = resetMagHeading(_mag_lpf.getState(), false, false);
 
-
 		// update the yaw angle variance using the variance of the measurement
 		if (_params.mag_fusion_type <= MAG_FUSE_TYPE_3D) {
 			// using magnetic heading tuning parameter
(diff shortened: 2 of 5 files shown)


EKF/ekf.cpp
```diff
@@ -216,6 +213,9 @@ bool Ekf::initialiseFilter()
 		// reset the output predictor state history to match the EKF initial values
 		alignOutputFilter();
 
+		// initialise the state covariance matrix now we have starting values for all lthe states
```
LorenzMeier: typo ("lthe states")

EKF/ekf.cpp
```diff
@@ -216,6 +213,9 @@ bool Ekf::initialiseFilter()
...
+		initialiseCovariance();
```
RomanBapst: @priseborough If we call this function here, aren't we then overriding the variances set by increaseQuatYawErrVariance() [here](https://github.com/PX4/ecl/pull/739/files#diff-0871d0235139c63531138e2793b76de7L202) ?

RomanBapst: @priseborough Would it make sense to have increaseQuatYawErrVariance() as part of initialiseCovariance(). Then we are always sure it gets called at the right moment.

EKF/ekf.cpp
priseborough: See https://github.com/PX4/ecl/pull/739/commits/0d12f1bdde6431dc32c2681049f8bffd36b9be21

EKF/ekf.cpp
priseborough: See https://github.com/PX4/ecl/pull/739/commits/0d12f1bdde6431dc32c2681049f8bffd36b9be21

Safer now to move it before the yaw variance is modified.

priseborough:
Only tested on replay.  Need to test with tailsitter SITL.

EKF/common.h
```diff
@@ -223,7 +223,7 @@ struct parameters {
 
 	// process noise
 	float gyro_bias_p_noise{1.0e-3f};	///< process noise for IMU rate gyro bias prediction (rad/sec**2)
-	float accel_bias_p_noise{6.0e-3f};	///< process noise for IMU accelerometer bias prediction (m/sec**3)
+	float accel_bias_p_noise{1.0e-2f};	///< process noise for IMU accelerometer bias prediction (m/sec**3)
```
bresch: The default parameter in PX4 for the accel bias process noise is 3e-3 ([EKF2_ACC_B_NOISE](https://dev.px4.io/master/en/advanced/parameter_reference.html#EKF2_ACC_B_NOISE)). Here, you then multiply the process noise by a factor of 3 to fix the initial stability issue of the bias. 
Isn't the initial divergence a sign of something going wrong? I agree that increasing the process noise accelerate convergence before takeoff, but the bias shouldn't go initially in the wrong direction.

priseborough: Agree, but we are doing those covariance calculations in single precision and stability issues when the variances for those states become very small has been something we have encountered before.

I have some ideas about how we can further investigate this which we can discuss at the dev call. These include instrumenting all the fusion methods active during replay of this log and identifying all those that drive the bias state positive. Also we need to fix the bug that stops replay logging before 26 seconds.

Issue #740, opened 2020-01-30
Make the test target run on MacOs


kamilritz:
Currently, the the EKF can be build on MacOS, but it is not possible to build the tests (`make test`)

PR #741, opened 2020-02-01
Make test with Clang

kamilritz:
This allows for the test also run with the Clang compiler.

@MaEtUgR You have added [this](https://github.com/PX4/ecl/pull/741/files#diff-59007f958bf3519ed464d95c410f5d1eL44) on the Firmware side. Do you know if this is still necessary? This caused issues with the Clang compiler.

I noticed that if the same code is compiled with Clang compiler, the change indication output differs from the one produced when compiled with GCC.

#740

### DIFF
--- a/test/gtest.cmake
+++ b/test/gtest.cmake
@@ -41,10 +41,3 @@ endif()
 
 # Add googletest, defines gtest and gtest_main targets
 add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/googletest-src ${CMAKE_CURRENT_BINARY_DIR}/googletest-build EXCLUDE_FROM_ALL)
-
-# Remove visibility.h from the compile flags for gtest because of poisoned exit()
-get_target_property(GTEST_COMPILE_FLAGS gtest COMPILE_OPTIONS)
-list(REMOVE_ITEM GTEST_COMPILE_FLAGS "-include")
-list(REMOVE_ITEM GTEST_COMPILE_FLAGS "visibility.h")
-set_target_properties(gtest PROPERTIES COMPILE_OPTIONS "${GTEST_COMPILE_FLAGS}")
-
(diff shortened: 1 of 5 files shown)


test/test_EKF_airspeed.cpp
```diff
@@ -96,7 +96,7 @@ TEST_F(EkfAirspeedTest, testWindVelocityEstimation)
 	EXPECT_TRUE(matrix::isEqual(vel, simulated_velocity_earth));
  	const Vector3f vel_wind_expected = simulated_velocity_earth - R_to_earth_sim * (Vector3f(airspeed_body(0), airspeed_body(1), 0.0f));
 	EXPECT_TRUE(matrix::isEqual(vel_wind_earth, Vector2f(vel_wind_expected.slice<2,1>(0,0))));
-	EXPECT_FLOAT_EQ(height_before_pressure_correction, 0.0f);
+	EXPECT_NEAR(height_before_pressure_correction, 0.0f, 1e-5f);
```
MaEtUgR:
```suggestion
	EXPECT_FLOAT_EQ(height_before_pressure_correction, 0.0f);
```
?

test/gtest.cmake
```diff
@@ -41,10 +41,3 @@ endif()
 
 # Add googletest, defines gtest and gtest_main targets
 add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/googletest-src ${CMAKE_CURRENT_BINARY_DIR}/googletest-build EXCLUDE_FROM_ALL)
-
-# Remove visibility.h from the compile flags for gtest because of poisoned exit()
```
MaEtUgR: This was only ncessary in the PX4/Firmware since the cmake setup there globally includes a custom visibility.h which poisons some symbols like `exit()` which are necessary to properly compile gtest. I'd presume it's not necessary for the ecl because I hope the cmake structure is smarter here.

test/test_EKF_airspeed.cpp
kamilritz: The float equal option will not work as the height can have a value that is xy*e^-6.

test/gtest.cmake
kamilritz: Cool, I think we should then be good, removing this in ECL.

kamilritz:
@MaEtUgR That is a good point. This is not really intentional. We should change that at some point, so that the tests are listed separately.

kamilritz:
@bresch

PR #743, opened 2020-02-10
ekf: add vehicle at rest to control status flags

CarlOlsson:


dagar:
Seems fine to me. Anyone else have an opinion about these flags? @bresch @RomanBapst

mhkabir:
I wonder if we should move this kind of motion detection out of the EKF and into land detector ("vehicle model").

PR #746, opened 2020-02-12
EKF: Revert "Make covariance matrices of type matrix::SquareMatrix"

priseborough:
Fixes https://github.com/PX4/ecl/issues/745

Log used for replay testing here: https://logs.px4.io/plot_app?log=e7623a61-20fc-4179-b048-38abe9ac6c6e

Log from replay using master at commit b399698: https://logs.px4.io/plot_app?log=66a8dfba-c1c2-4ea2-81fc-bbc5836d2c5b

Summary Plot
Screen Shot 2020-02-12 at 11 03 07 am

Log from replay of this PR using PX4/Firmware branch https://github.com/priseborough/Firmware/tree/debug_using_ecl_master: https://logs.px4.io/plot_app?log=6daac421-cc73-4937-82b0-f33ba077c20e

Summary Plot
Screen Shot 2020-02-12 at 11 12 43 am

There is a change the external interface access to the covariance data - see https://github.com/PX4/ecl/compare/master...priseborough:debug_using_ecl_master?expand=1#diff-0be2b3391d4586f3ecc561669eee2529. This could probably be avoided. Suggestions are welcome.

The matrix::SquareMatrix changes can then be reintroduced starting with non-functional changes and using the sensor replay capability added by https://github.com/PX4/ecl/pull/717 to verify output.

PR #747, opened 2020-02-12
EKF: Revert "Make covariance matrices of type matrix::SquareMatrix"

bresch:
replaces https://github.com/PX4/ecl/pull/746


jkflying:
Reassuring to see that the 'change indicator' test is working as expected. I guess this would have been caught if the change indicator was in earlier.

bresch:
@jkflying Yes, and it will be really useful when we'll have to re-apply the patch

bresch:
Made a flight test and everything seems fine:
https://logs.px4.io/plot_app?log=ecb397a9-815d-4600-8b6d-80b7724a6f29
newplot(2)

jkflying:
Just an idea, if you make a covariances_diagonal() compatible with the old API, then you won't need a firmware PR. Especially since this is hopefully a temporary fix.

bresch:
replaced by https://github.com/PX4/ecl/pull/749

PR #751, opened 2020-02-13
EKF: Add emergency yaw recovery using EKF-GSF estimator (backport)

priseborough:
This change request adds an alternative method of estimating yaw that only relies on IMU gyro, IMU acceleration and GPS horizontal velocity data. It uses a multiple-hypothesis bank of 3-state Extended Kalman Filters whose outputs are combined using a Gaussian Sum Filter. This method was selected after evaluation of several alternatives in Matlab - see https://github.com/priseborough/3_state_filter/tree/flightLogReplay-wip

The change is being made against both stable (this PR) and master (PR TBD) to facilitate wider testing.

- A bank of up to 8 complementary filters using only IMU and airspeed in FW flight if available is used to generate a yaw rate, yaw angle and a front and right acceleration.
- These are then used by a bank of 3-state EKF's that estimate North and East velocity and yaw angle. These are corrected using the GPS horizontal velocity measurements.
- The output from the bank of  EKF's are then combined using a Gaussian Sum Filter (GSF) which estimates a weighted combination of the output from the bank of EKF yaw estimators
- The EKF's and GSF do not start running until the vehicle is airborne. The bank of AHRS solutions run all the time.

If large innovations and loss of navigation is detected in the 30 second period after takeoff, the main navigation EKF resets it's yaw angle to the output from the Gaussian Sum Filter
It is intended to provide a means of recovering yaw and hence navigation following takeoff with a bad magnetometer or magnetic environment rather than rely on the PX4 Commander placing the vehicle into a non-position control mode which for flight in AUTO modes, usually results in damage to the vehicle.

It has undergone a significant amount of testing with a multi-rotor platform and off-line testing and tuning using replay. TODO upload logs and provide links. Loss of navigation was instigated by setting EKF2_MAG_DECL to -90 degrees (100 degrees from the local declination) and EKF2_DECL_TYPE = 0 to force the EKF to use the parameter value. It was tested with the following PX4/Firmware changes to facilitate data logging and some experimental changes to Commander to test the feasibility of a  Commander selected reset. This was not successful and the internal EKF reset was used instead,

https://github.com/priseborough/Firmware/tree/pr-yawEKFGSF-stable

The following figures shows how the bank of yaw estimators converges immediately after takeoff and how the weighting calculated by the Gaussian Sum Filter stabilises. The rapid rise in innovations as the flyaway commences, followed by the reset of the EKF yaw to the GSF-EKF estimate and the fall in innovations after the yaw reset can be seen. After the reset, magnetometer use is inhibited.

Test 1
Screen Shot 2020-02-13 at 9 45 00 pm

Test 2
Screen Shot 2020-02-13 at 10 23 48 pm

Test 3
Screen Shot 2020-02-13 at 10 27 24 pm

Test 4
Screen Shot 2020-02-13 at 10 28 21 pm

Test 5
Screen Shot 2020-02-13 at 10 33 12 pm

Test 6 - this flight included significant post reset manoeuvring in manual mode
Screen Shot 2020-02-13 at 10 34 44 pm

Test 7 - Tailsitter with forward flight transition. This flight data was generated using a correct compass declination, so no reset was performed. It demonstrates that the yaw estimates converge and  closely agree with the main nav filter during forward flight with sustained coordinated turns.
Screen Shot 2020-02-13 at 10 55 38 pm

TODO make changes on the PX4/Firmware side to synchronise the Commander failsafes so that the yaw reset activates before the failsafe lands the vehicle. 

Acknowledgements:

Rudaba Khan https://github.com/Rudaba for most of the GSF, IMM and Particle Filter Matlab code
@CarlOlsson for testing, review and code improvements.

Logs from multi-rotor flight testing used for verification and post flight replay tuning:
https://logs.px4.io/plot_app?log=285fad2c-5273-43ad-a32e-c797a5688c08
https://logs.px4.io/plot_app?log=c657a366-16b5-4a7c-aa2c-75d1ff852c23
https://logs.px4.io/plot_app?log=3ee14866-8415-4c0a-a010-dadf5821e39f
https://logs.px4.io/plot_app?log=1c2c64e2-b66a-4684-bf2e-eebbe24dd5d2
https://logs.px4.io/plot_app?log=645d7b21-3dfa-4366-b75b-fb43faf1fae4

### DIFF
--- a/EKF/EKFGSF_yaw_estimator.cpp
+++ b/EKF/EKFGSF_yaw_estimator.cpp
@@ -427,10 +427,12 @@ void Ekf::stateUpdateEKFGSF(const uint8_t model_index)
 
 	} else {
 		// using a 321 Tait-Bryan rotation to define yaw state
-		// take roll pitch yaw from AHRS prediction
-		Eulerf euler321(_ahrs_ekf_gsf[model_index].R);
+		// take roll and pitch from AHRS prediction
+		Eulerf euler321;
+		euler321(0) = atan2f(_ahrs_ekf_gsf[model_index].R(2,1), _ahrs_ekf_gsf[model_index].R(2,2));
+		euler321(1) = -asinf(_ahrs_ekf_gsf[model_index].R(2,0));
 
-		// replace the yaw angle using the EKF state estimate
+		// take yaw angle from the 3-state EKF updated state estimate
 		euler321(2) = _ekf_gsf[model_index].X[2];
 
 		// update the rotation matrix used by the AHRS prediction algorithm
@@ -630,9 +632,7 @@ float Ekf::gaussianDensityEKFGSF(const uint8_t model_index) const
 	// transpose(innovation) * inv(S) * innovation
 	float normDist = tempVec[0] * _ekf_gsf[model_index].innov[0] + tempVec[1] * _ekf_gsf[model_index].innov[1];
 
-	normDist = expf(-0.5f * normDist);
-	normDist *= sqrtf(t4)/ M_TWOPI_F;
-	return normDist;
+	return M_TWOPI_INV * sqrtf(t4) * expf(-0.5f * normDist);
 }
 
 bool Ekf::getDataEKFGSF(float *yaw_composite, float *yaw_variance, float yaw[N_MODELS_EKFGSF], float innov_VN[N_MODELS_EKFGSF], float innov_VE[N_MODELS_EKFGSF], float weight[N_MODELS_EKFGSF]) const
--- a/EKF/EKFGSF_yaw_estimator.cpp
+++ b/EKF/EKFGSF_yaw_estimator.cpp
@@ -838,7 +838,8 @@ Matrix3f Ekf::updateRotMatEKFGSF(Matrix3f &R, Vector3f &g)
 	for (uint8_t r = 0; r < 3; r++) {
 		rowLengthSq = ret(r,0) * ret(r,0) + ret(r,1) * ret(r,1) + ret(r,2) * ret(r,2);
 		if (rowLengthSq > FLT_EPSILON) {
-			const float rowLengthInv = 1.0f / sqrtf(rowLengthSq);
+			// Use linear approximation for inverse sqrt taking advantage of the row length being close to 1.0
+			const float rowLengthInv = 1.5f - 0.5f * rowLengthSq;
 			ret(r,0) *= rowLengthInv;
 			ret(r,1) *= rowLengthInv;
 			ret(r,2) *= rowLengthInv;
--- a/EKF/EKFGSF_yaw_estimator.cpp
+++ b/EKF/EKFGSF_yaw_estimator.cpp
@@ -744,9 +744,7 @@ bool Ekf::resetYawToEKFGSF()
 
 }
 
-// request the EKF reset the yaw to the estimate from the internal EKF-GSF filter
-// argument should be incremented only when a new reset is required
-void Ekf::request_ekfgsf_yaw_reset(uint8_t counter)
+void Ekf::requestEmergencyNavReset(uint8_t counter)
 {
 	if (counter > _yaw_extreset_counter) {
 		_yaw_extreset_counter = counter;
--- a/EKF/EKFGSF_yaw_estimator.cpp
+++ b/EKF/EKFGSF_yaw_estimator.cpp
@@ -583,7 +583,7 @@ void Ekf::runEKFGSF()
 	// models with larger innovations are weighted less
 	_ekf_gsf_yaw_variance = 0.0f;
 	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
-		float yaw_delta = _ekf_gsf[model_index].X[2] - X_GSF[2];
+		float yaw_delta = wrap_pi(_ekf_gsf[model_index].X[2] - X_GSF[2]);
 		_ekf_gsf_yaw_variance += _ekf_gsf[model_index].W * (_ekf_gsf[model_index].P[2][2] + sq(yaw_delta));
 	}
 }
--- a/EKF/EKFGSF_yaw_estimator.cpp
+++ b/EKF/EKFGSF_yaw_estimator.cpp
@@ -515,20 +515,8 @@ void Ekf::runEKFGSF()
 		return;
 	}
 
-	// calculate common values used by the AHRS prediction models
-	_ahrs_accel_norm = _ahrs_accel.norm();
-	_ahrs_turn_comp_enabled = _control_status.flags.fixed_wing && _params.EKFGSF_tas_default > FLT_EPSILON;
-	if (_ahrs_accel_norm > CONSTANTS_ONE_G) {
-		if (_ahrs_turn_comp_enabled && _ahrs_accel_norm <= 2.0f * CONSTANTS_ONE_G) {
-			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f - (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
-		} else if (_ahrs_accel_norm <= 1.5f * CONSTANTS_ONE_G) {
-			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f - 2.0f * (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
-		}
-	} else if (_ahrs_accel_norm > 0.5f * CONSTANTS_ONE_G) {
-		_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f + 2.0f * (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
-	}
-
 	// AHRS prediction cycle for each model - this always runs
+	calcAccelGainEKFGSF();
 	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
 		statePredictEKFGSF(model_index);
 	}
@@ -824,4 +812,22 @@ Dcmf Ekf::taitBryan312ToRotMat(Vector3f &rot312)
 		R(2, 1) = s1;
 
 		return R;
+}
+
+void Ekf::calcAccelGainEKFGSF()
+{
+	// calculate acceleration fusion  gain and other common values used by the AHRS complementary filter models
+	_ahrs_accel_norm = _ahrs_accel.norm();
+	_ahrs_turn_comp_enabled = _control_status.flags.fixed_wing && _params.EKFGSF_tas_default > FLT_EPSILON;
+	if (_ahrs_accel_norm > CONSTANTS_ONE_G) {
+		if (_ahrs_turn_comp_enabled && _ahrs_accel_norm <= 2.0f * CONSTANTS_ONE_G) {
+			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f - (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
+		} else if (_ahrs_accel_norm <= 1.5f * CONSTANTS_ONE_G) {
+			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f - 2.0f * (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
+		}
+	} else if (_ahrs_accel_norm > 0.5f * CONSTANTS_ONE_G) {
+		_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f + 2.0f * (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
+	} else {
+		_ahrs_accel_fusion_gain = 0.0f;
+	}
 }
--- a/EKF/EKFGSF_yaw_estimator.cpp
+++ b/EKF/EKFGSF_yaw_estimator.cpp
@@ -409,23 +409,18 @@ void Ekf::stateUpdateEKFGSF(const uint8_t model_index)
 	}
 	float yawDelta = _ekf_gsf[model_index].X[2] - oldYaw;
 
-	// calculate the delta yaw rotation matrix
-	Matrix3f yawRotMat;
+	// apply the change in yaw angle to the AHRS
+	// take advantage of sparseness in the yaw rotation matrix
 	const float cosYaw = cosf(yawDelta);
 	const float sinYaw = sinf(yawDelta);
-	yawRotMat(0,0) = cosYaw;
-	yawRotMat(0,1) = -sinYaw;
-	yawRotMat(0,2) = 0.0f;
-	yawRotMat(1,0) = sinYaw;
-	yawRotMat(1,1) = cosYaw;
-	yawRotMat(1,2) = 0.0f;
-	yawRotMat(2,0) = 0.0f;
-	yawRotMat(2,1) = 0.0f;
-	yawRotMat(2,2) = 1.0f;
+	Matrix3f  R = _ahrs_ekf_gsf[model_index].R;
+	_ahrs_ekf_gsf[model_index].R(0,0) = R(0,0) * cosYaw - R(1,0) * sinYaw;
+	_ahrs_ekf_gsf[model_index].R(0,1) = R(0,1) * cosYaw - R(1,1) * sinYaw;
+	_ahrs_ekf_gsf[model_index].R(0,2) = R(0,2) * cosYaw - R(1,2) * sinYaw;
+	_ahrs_ekf_gsf[model_index].R(1,0) = R(0,0) * sinYaw + R(1,0) * cosYaw;
+	_ahrs_ekf_gsf[model_index].R(1,1) = R(0,1) * sinYaw + R(1,1) * cosYaw;
+	_ahrs_ekf_gsf[model_index].R(1,2) = R(0,2) * sinYaw + R(1,2) * cosYaw;
 
-	// apply to the AHRS
-	// TODO take advantage of sparseness in yawRotMat
-	_ahrs_ekf_gsf[model_index].R = yawRotMat * _ahrs_ekf_gsf[model_index].R;
 }
 
 void Ekf::initialiseEKFGSF()
--- a/EKF/EKFGSF_yaw_estimator.cpp
+++ b/EKF/EKFGSF_yaw_estimator.cpp
@@ -816,17 +816,25 @@ Dcmf Ekf::taitBryan312ToRotMat(Vector3f &rot312)
 
 void Ekf::calcAccelGainEKFGSF()
 {
-	// calculate acceleration fusion  gain and other common values used by the AHRS complementary filter models
+	// calculate common values used by the AHRS complementary filter models
 	_ahrs_accel_norm = _ahrs_accel.norm();
 	_ahrs_turn_comp_enabled = _control_status.flags.fixed_wing && _params.EKFGSF_tas_default > FLT_EPSILON;
-	if (_ahrs_accel_norm > CONSTANTS_ONE_G) {
-		if (_ahrs_turn_comp_enabled && _ahrs_accel_norm <= 2.0f * CONSTANTS_ONE_G) {
-			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f - (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
-		} else if (_ahrs_accel_norm <= 1.5f * CONSTANTS_ONE_G) {
-			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f - 2.0f * (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
+
+	// Calculate the acceleration fusion gain using a continuous function that is unity at 1g and zero
+	// at the min and mas g value. Allow for more acceleration when flying as a fixed wing vehicle using centripetal
+	// acceleration correction as higher and more sustained g will be experienced.
+	// Use a quadratic finstead of linear unction to prevent vibration around 1g reducing the tilt correction effectiveness.
+	float accel_g = _ahrs_accel_norm / CONSTANTS_ONE_G;
+	if (accel_g > 1.0f) {
+		if (_ahrs_turn_comp_enabled && accel_g < 2.0f) {
+			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(2.0f - accel_g);
+		} else if (accel_g < 1.5f) {
+			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(3.0f - 2.0f * accel_g);
+		} else {
+			_ahrs_accel_fusion_gain = 0.0f;
 		}
-	} else if (_ahrs_accel_norm > 0.5f * CONSTANTS_ONE_G) {
-		_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f + 2.0f * (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
+	} else if (accel_g > 0.5f) {
+		_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(2.0f * accel_g - 1.0f);
 	} else {
 		_ahrs_accel_fusion_gain = 0.0f;
 	}
--- a/EKF/EKFGSF_yaw_estimator.cpp
+++ b/EKF/EKFGSF_yaw_estimator.cpp
@@ -76,8 +76,10 @@ void Ekf::quatPredictEKFGSF(const uint8_t model_index)
 	// Perform angular rate correction using accel data and reduce correction as accel magnitude moves away from 1 g (reduces drift when vehicle picked up and moved).
 	// During fixed wing flight, compensate for centripetal acceleration assuming coordinated turns and X axis forward
 	Vector3f correction = {};
-	Vector3f accel = _ahrs_accel;
-	if (_ahrs_accel_norm > 0.5f * CONSTANTS_ONE_G && (_ahrs_accel_norm < 1.5f * CONSTANTS_ONE_G || _ahrs_turn_comp_enabled)) {
+	if (_ahrs_accel_fusion_gain > 0.0f) {
+
+		Vector3f accel = _ahrs_accel;
+
 		if (_ahrs_turn_comp_enabled) {
 			// turn rate is component of gyro rate about vertical (down) axis
 			const float turn_rate = _ahrs_ekf_gsf[model_index].R(2,0) * _ang_rate_delayed_raw(0)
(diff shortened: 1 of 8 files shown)


priseborough:
Thankyou @CarlOlsson  for spotting this bug:

If for example a solution of -170 and +170 degrees was to be summed with a weight of 0.5 each, the resultant would currently be (-170 + 170)/2 = 0 deg instead of the 180 deg required.

I will modify the way the yaw solutions are combined so that each yaw solution is converted to a NE plane unit vector. Each vector is then multiplied by its weighting and the vectors are then summed. The yaw angle of the summed vector is the weighted value required.

A patch for this will be pushed tomorrow.

jkflying:
The other way I've seen this solved is by averaging like this:
```
mean = a + relative_angle(b - a) / 2
```
The `relative_angle()` function just wraps between +- 180.

dagar:
Architecturally do you actually want this to reside within ecl/EKF or is for convenience utilizing the existing infrastructure (the data you need is already on hand, ekf2 replay) plus simplicity of yaw reset to EKF GSF?

priseborough:
@dagar There  are two reasons for placing it in the EKF, firstly the access to the time aligned IMU and GPS data using the exisiting buffers and secondly being able to perform the fallback reset using data at the same time horizon.

priseborough:
Replay test output for https://github.com/PX4/ecl/pull/751/commits/9e20f9d09c7d67893e9636a9fa90588cc10f75e8 which changes how the weighted average composite yaw is obtained from the individual filter states.

Screen Shot 2020-02-14 at 6 50 36 am

EKF/ekf.cpp
```diff
@@ -88,6 +88,8 @@ bool Ekf::init(uint64_t timestamp)
 
 	_dt_ekf_avg = FILTER_UPDATE_PERIOD_S;
 
+	_ang_rate_delayed_raw.zero();
```
kamilritz: This is not necessary, as the matrix::Vector will be initialized with zeros anyway. Also for other variables we left out an explicit assignment to zero in the `init()` function.

EKF/ekf.cpp
```diff
@@ -124,6 +126,9 @@ bool Ekf::update()
 		runTerrainEstimator();
 
 		updated = true;
+
```
kamilritz: I would prefer the `updated = true` at the end of the block.

EKF/estimator_interface.h
```diff
@@ -265,8 +265,8 @@ class EstimatorInterface
 	// return true if the global position estimate is valid
 	virtual bool global_position_is_valid() = 0;
 
-	// return true if the EKF is dead reckoning the position using inertial data only
-	bool inertial_dead_reckoning() {return _is_dead_reckoning;}
+	// return true if the EKF has exceeded the maximum time for dead reckoning the position using inertial data only
+	bool inertial_dead_reckoning() {return _deadreckon_time_exceeded;}
```
kamilritz: This is quite a functional change. This seems not related to the rest of the PR. Could be good to have this as a separate PR.

EKF/ekf.h
```diff
@@ -264,6 +264,24 @@ class Ekf : public EstimatorInterface
...
+				float delVel[3], // delta velocity from IMU (m/s)
+				float *delVelDt, // delta velocity integration time (sec)
+				float vel[3], // NED velocity measurement (m/s)
+				float *velErr, // 1-sigma velocity measurement uncertainty (m/s)
+				bool *fuse_vel, // true when velociy measurement has been updated
+				float quat[4]); // reference quaternion from EKF
+
+	// request the EKF reset the yaw to the estimate from the internal EKF-GSF filter
+	// argment should be incremented only when a new reset is required
+	void request_ekfgsf_yaw_reset(uint8_t counter);
```
kamilritz: The preferred case style for functions is snakeCase

EKF/ekf.cpp
```diff
@@ -359,6 +364,10 @@ void Ekf::predictState()
 	// filter and limit input between -50% and +100% of nominal value
 	input = math::constrain(input, 0.5f * FILTER_UPDATE_PERIOD_S, 2.0f * FILTER_UPDATE_PERIOD_S);
 	_dt_ekf_avg = 0.99f * _dt_ekf_avg + 0.01f * input;
+
+	// some calculations elsewhere in code require a raw angular rate vector so calculate here to avoid duplication
+	_ang_rate_delayed_raw = _imu_sample_delayed.delta_ang / fmaxf(_imu_sample_delayed.delta_ang_dt, FILTER_UPDATE_PERIOD_S / 4);
```
kamilritz: Is there a reason for lower bounding the dt? Why `FILTER_UPDATE_PERIOD_S / 4`?

EKF/control.cpp
```diff
@@ -674,7 +673,31 @@ void Ekf::controlGpsFusion()
...
+			if (reset_yaw_to_EKFGSF) {
+				// Attempt to recover using an alternative algorithm for estimating yaw
+				if (_ekf_gsf_vel_fuse_started && (_imu_sample_delayed.time_us - _emergency_yaw_reset_time) > 5000000) {
+					resetYawToEKFGSF();
+					_emergency_yaw_reset_time = _imu_sample_delayed.time_us;
+					_do_emergency_yaw_reset = false;
+				}
+
+				// Reset the timeout counters
+				_time_last_pos_fuse = _time_last_imu;
```
kamilritz: Should we reset the counters even if we did not enter the above if statement and did not call `resetYawToEKFGSF()`

kamilritz: Should `_time_last_delpos_fuse = _time_last_imu;` and `_time_last_of_fuse = _time_last_imu;` also be added to the `do_reset`-case?

EKF/EKFGSF_yaw_estimator.cpp
```diff
@@ -0,0 +1,799 @@
...
+	_output_new.quat_nominal = _state_reset_status.quat_change * _output_new.quat_nominal;
+
+	// capture the reset event
+	_state_reset_status.quat_counter++;
+
+	// reset velocity and position states to GPS - if yaw is fixed then the filter should start to operate correctly
+	resetVelocity();
+	resetPosition();
+
+	ECL_INFO_TIMESTAMPED("EKF emergency yaw reset");
```
kamilritz: The "EKF" in the beginning of the string is not really necessary, as the prefix "EKF" will be printed anyways. And it adds flash space. Same below.

priseborough:
See https://github.com/PX4/ecl/pull/754#issuecomment-586540282 for more recent testing of the same code rebased on master.

priseborough:
@kamilritz It has been submitted against both stable and master on request of a project sponsor who are working from the last stable release and were concerned about the stability of the master branch given the amount of change in the past few months. Changes arising from review will be made against both PR's.

Thank you for your review comments. I will go through them in due course. 

I am happy to have this broken into a separate class, but may need some assistance with doing that.

EKF/control.cpp
```diff
@@ -674,7 +673,31 @@ void Ekf::controlGpsFusion()
...
-			if (do_reset) {
+			// A reset to the EKF-GSF estimate can be performed after a recent takeoff which will enable
+			// recovery from a bad magnetometer or field estimate.
+			// This special case reset can also be requested externally.
+			// The minimum time interval between resets to the EKF-GSF estimate must be limited to
+			// allow the EKF-GSF time to improve its estimate if the first reset was not successful.
+			const bool stopped_following_gps_velocity = (_time_last_imu - _time_last_vel_fuse) > _params.EKFGSF_reset_delay &&
+							   (_time_last_vel_fuse > _time_last_on_ground_us);
+			const bool recent_takeoff = (_control_status.flags.in_air && (_imu_sample_delayed.time_us - _time_last_on_ground_us) < 30000000);
+			const bool reset_yaw_to_EKFGSF = (do_reset || _do_emergency_yaw_reset || stopped_following_gps_velocity) && recent_takeoff && ((_imu_sample_delayed.time_us - _emergency_yaw_reset_time) > 5000000);
```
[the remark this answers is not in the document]
CarlOlsson: `do_reset` is only true if both the GPS velocity and position measurements are failing the innovations checks. `stopped_following_gps_velocity` was added in order to trigger the reset as soon as the velocity measurements were rejected for one second, this generally happens several seconds before the position innovation checks are failing

EKF/EKFGSF_yaw_estimator.cpp
```diff
@@ -0,0 +1,799 @@
...
+	}
+}
+
+// converts Tait-Bryan 312 sequence of rotations from frame 1 to frame 2
+// to the corresponding rotation matrix that rotates from frame 2 to frame 1
+// rot312(0) - First rotation is a RH rotation about the Z axis (rad)
+// rot312(1) - Second rotation is a RH rotation about the X axis (rad)
+// rot312(2) - Third rotation is a RH rotation about the Y axis (rad)
+// See http://www.atacolorado.com/eulersequences.doc
+Dcmf Ekf::taitBryan312ToRotMat(Vector3f &rot312)
```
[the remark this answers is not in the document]
CarlOlsson: Also the tilt align zero yaw code can be put in a function and used for the main ekf to save some flash

CarlOlsson:
This adds around 4 percentage points of CPU load on a v4 processor. Some ideas to reduce it
- Every filter in the bank estimates the same gyro bias -> only do it once
- Every filter in the bank estimates the same tilt angle, only the yaw differs- > investigate if it is possible to reduce the number of computations

jkflying:
Another idea for CPU consumption reduction, from the graphs it appears that (at least in this case) all except 1 of the filters converged to the same solution. This means to me that you could probably run 3 (or 5) instead of 8. You would have to be a bit smarter than simply averaging, eg. more like a kernel density/vote mechanism, but it should drop the CPU a lot.

CarlOlsson:
> You would have to be a bit smarter than simply averaging, eg. more like a kernel density/vote mechanism, but it should drop the CPU a lot.

@jkflying Right now the composite yaw angle is not just the average but the weighted average. The weight of each filter is the likelihood calculated by sampling a two dimensional zero mean Gaussian, where the covariance matrix is the filter innovation variance and the sample point is the innovation. This is why the output of the filter bank converges to the correct yaw angle much faster than the individual filters converge to the same angle. Is something like this what you had in mind?

EKF/ekf.cpp
priseborough: Zero is bad obviously, so  /4 was chosen to give plenty of margin for frame jitter.

EKF/control.cpp
```diff
@@ -674,7 +673,31 @@ void Ekf::controlGpsFusion()
 			// We haven't had an absolute position fix for a longer time so need to do something
 			do_reset = do_reset || ((_time_last_imu - _time_last_pos_fuse) > (2 * _params.reset_timeout_max));
 
-			if (do_reset) {
+			// A reset to the EKF-GSF estimate can be performed after a recent takeoff which will enable
+			// recovery from a bad magnetometer or field estimate.
+			// This special case reset can also be requested externally.
+			// The minimum time interval between resets to the EKF-GSF estimate must be limited to
+			// allow the EKF-GSF time to improve its estimate if the first reset was not successful.
+			const bool stopped_following_gps_velocity = (_time_last_imu - _time_last_vel_fuse) > _params.EKFGSF_reset_delay &&
```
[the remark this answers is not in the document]
priseborough: Will do.

EKF/control.cpp
```diff
@@ -674,7 +673,31 @@ void Ekf::controlGpsFusion()
...
+							   (_time_last_vel_fuse > _time_last_on_ground_us);
+			const bool recent_takeoff = (_control_status.flags.in_air && (_imu_sample_delayed.time_us - _time_last_on_ground_us) < 30000000);
+			const bool reset_yaw_to_EKFGSF = (do_reset || _do_emergency_yaw_reset || stopped_following_gps_velocity) && recent_takeoff && ((_imu_sample_delayed.time_us - _emergency_yaw_reset_time) > 5000000);
+
+			if (reset_yaw_to_EKFGSF) {
+				// Attempt to recover using an alternative algorithm for estimating yaw
+				if (_ekf_gsf_vel_fuse_started && (_imu_sample_delayed.time_us - _emergency_yaw_reset_time) > 5000000) {
+					resetYawToEKFGSF();
+					_emergency_yaw_reset_time = _imu_sample_delayed.time_us;
```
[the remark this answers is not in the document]
priseborough: Habit. In the past the EKF code used the fusion time  horizon (delayed time) for internal fusion logic. I will review usage to ensure it is consistent.

EKF/control.cpp
```diff
@@ -674,7 +673,31 @@ void Ekf::controlGpsFusion()
...
+			// The minimum time interval between resets to the EKF-GSF estimate must be limited to
+			// allow the EKF-GSF time to improve its estimate if the first reset was not successful.
+			const bool stopped_following_gps_velocity = (_time_last_imu - _time_last_vel_fuse) > _params.EKFGSF_reset_delay &&
+							   (_time_last_vel_fuse > _time_last_on_ground_us);
+			const bool recent_takeoff = (_control_status.flags.in_air && (_imu_sample_delayed.time_us - _time_last_on_ground_us) < 30000000);
+			const bool reset_yaw_to_EKFGSF = (do_reset || _do_emergency_yaw_reset || stopped_following_gps_velocity) && recent_takeoff && ((_imu_sample_delayed.time_us - _emergency_yaw_reset_time) > 5000000);
+
+			if (reset_yaw_to_EKFGSF) {
+				// Attempt to recover using an alternative algorithm for estimating yaw
+				if (_ekf_gsf_vel_fuse_started && (_imu_sample_delayed.time_us - _emergency_yaw_reset_time) > 5000000) {
```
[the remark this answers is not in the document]
priseborough: Will fix.

EKF/control.cpp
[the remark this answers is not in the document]
priseborough: A loss of navigation after takeoff is an emergency. However the 30 second activation limit can be removed when we get more confidence in the long term stability of the algorithm. Preliminary test result are encouraging, but we haven't done extended FW flights or hovering flights in one position yet.

EKF/EKFGSF_yaw_estimator.cpp
priseborough: Will combine messages and fix.

EKF/control.cpp
priseborough: Will check.

EKF/ekf.h
priseborough: Will fix.

EKF/estimator_interface.h
priseborough: Will remove. With improvements to the internal reset we are no longer relying on the commander to trigger a reset externally.

EKF/control.cpp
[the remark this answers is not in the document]
priseborough: Will check.

EKF/ekf.cpp
kamilritz: Is it possible for delta_ang_dt to become 0?

EKF/control.cpp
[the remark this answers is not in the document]
kamilritz: But could there be a situation where we need to do the reset in flight? In this case the variable `recent_takeoff` will not allow for the EKFGSF reset.

jkflying:
@CarlOlsson It's fairly similar, depending on how strongly the GSF de-weights the outlier(s). Doing a replay with the bank limited to 3 would be a good test to see if the GSF is really doing what you expect with marginal data. I'd expect the 2 that agree to drive the main estimate, and cause the outlier to be completely rejected.

CarlOlsson:
Easy logfile = high horizontal acceleration directly after takeoff
8 filters (top) vs 3 filters (bottom)
image

The output from the filter banks are almost identical
image

Hard logfile = no horizontal acceleration until t=320s
image
image

The output from the filter banks are almost identical. Based on just these two logfiles, no reduction in performance is observed by reducing the number of filters to 3

jkflying:
That looks really promising. The case you didn't manage to hit there was, with 3 filters, if one of them initialized at the same angle as the one that didn't converge in the initial 8-bank replay. If you can replicate that, and the GSF can handle the case of 1 outlier + 2 inliers, then I'd say voting isn't necessary at small bank sizes :+1:

CarlOlsson:
@jkflying Here you go
image
image

jkflying:
Awesome. I'm happy to say it's good with 3 filters. So that cuts the CPU by roughly 60%.

EKF/control.cpp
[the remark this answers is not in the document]
priseborough: I think keeping the concept of a defined 'post takeoff' time period makes sense in the context of setting loss of navigation parameters when we are trying to detect loss of navigation caused by a bad yaw estimate. This is because crash logs submitted in the past have shown a loss of navigation immediately after takeoff due to a bad yaw estimate is a common cause and is exacerbated by proximity to ground obstacles during that period of operation. Recovering rapidly from this type of error was the primary motivator for developing this algorithm.

Usually if the EKF takes off with a good yaw angle, it is able to reject the magnetometer or learn the new biases and field values for errors that develop later in flight, but not always. There are yaw induced failure scenarios that can occur later in flight that we may also be able to detect abd recover from using this estimator.

Splitting the detection into a period immediately following takeoff with a more rapid time response and a subsequent period with a slower time response and lower false detection rate will be investigated.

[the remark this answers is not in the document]
priseborough: The exisiting missing aiding check is too slow to catch a yaw error induced fly-away in time. Also if we are using optical flow, we still want to reset the yaw if post takeoff movement indicates it is bad, even if the optical flow is preventing loss of navigation.

EKF/control.cpp
priseborough: Timer reset has been addressed by https://github.com/PX4/ecl/pull/751/commits/7ff3418f7aff5a3543aef2ce87ca6a4fe81a9e36

EKF/ekf.h
priseborough: I've renamed that function to better reflect when it should be used and what it does. See https://github.com/PX4/ecl/pull/751/commits/87b23448807b3f36eb8f053909f683af6be6e9e3

EKF/estimator_interface.h
priseborough: See https://github.com/PX4/ecl/pull/751/commits/8606d981dd3c1d34e6432cd6036af8e45d8650a5

EKF/control.cpp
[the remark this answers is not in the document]
priseborough: See https://github.com/PX4/ecl/pull/751/commits/91548dae66e6dc01e57fc8c5e6229b2a47f6c414

EKF/control.cpp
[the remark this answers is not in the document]
priseborough: Se https://github.com/PX4/ecl/pull/751/commits/91548dae66e6dc01e57fc8c5e6229b2a47f6c414

EKF/control.cpp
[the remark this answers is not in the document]
priseborough: Se https://github.com/PX4/ecl/pull/751/commits/91548dae66e6dc01e57fc8c5e6229b2a47f6c414

priseborough:
I'm re-run some of my replay data using 3 vs 5 vs 8 filters and compared how quickly the estimate yaw variance falls below the maximum use threshold of 0.04 (0.2^2) that we currently set. There was no significant difference in the convergence time or the quality of the subsequent yaw estimate based on the main filter innovation sequences post reset. See

Screen Shot 2020-02-19 at 10 52 37 am

Using a larger number of filters does reduce the yaw uncertainty that each filter is initialised to which in theory would provide more immunity to GPS velocity error spikes immediately after filter alignment.

EKF/control.cpp
```diff
@@ -674,7 +673,31 @@ void Ekf::controlGpsFusion()
...
 
-			if (do_reset) {
+			// A reset to the EKF-GSF estimate can be performed after a recent takeoff which will enable
+			// recovery from a bad magnetometer or field estimate.
+			// This special case reset can also be requested externally.
+			// The minimum time interval between resets to the EKF-GSF estimate must be limited to
+			// allow the EKF-GSF time to improve its estimate if the first reset was not successful.
+			const bool stopped_following_gps_velocity = _time_last_imu > (_params.EKFGSF_reset_delay + _time_last_vel_fuse) &&
+							   (_time_last_vel_fuse > _time_last_on_ground_us);
+			const bool recent_takeoff = (_control_status.flags.in_air && (_imu_sample_delayed.time_us - _time_last_on_ground_us) < 30000000);
```
kamilritz: There is one subtraction of two unsigned integers left here.

EKF/control.cpp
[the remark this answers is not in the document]
kamilritz: Ok, I understand now. The different response time is a good point

EKF/control.cpp
kamilritz: Thanks. Should _time_last_delpos_fuse = _time_last_imu; and _time_last_of_fuse = _time_last_imu; also be added to the `do_reset`-case below the GSF case ?

EKF/EKFGSF_yaw_estimator.cpp
```diff
@@ -0,0 +1,827 @@
...
+			_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f - 2.0f * (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
+		}
+	} else if (_ahrs_accel_norm > 0.5f * CONSTANTS_ONE_G) {
+		_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.0f + 2.0f * (_ahrs_accel_norm - CONSTANTS_ONE_G)/CONSTANTS_ONE_G);
+	}
+
+	// AHRS prediction cycle for each model - this always runs
+	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
+		statePredictEKFGSF(model_index);
+	}
```
bresch:
```c++
	_ahrs_accel_norm = _ahrs_accel.norm();
	_ahrs_turn_comp_enabled = _control_status.flags.fixed_wing && _params.EKFGSF_tas_default > FLT_EPSILON;
	computeAhrsFusionGain()

	// AHRS prediction cycle for each model - this always runs
	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
		statePredictEKFGSF(model_index);
	}
```
```c++
void Ekf::computeAhrsFusionGain()
{
   	// calculate common values used by the AHRS prediction models
   	// see https://www.desmos.com/calculator/c8hztzyrni
	float attenuation = 2.f;
	if (_ahrs_turn_comp_enabled 
	    && _ahrs_accel_norm > CONSTANTS_ONE_G
	    && _ahrs_accel_norm <= 2.f * CONSTANTS_ONE_G) {
		attenuation = 1.f;
	}

	const float delta_accel = _ahrs_accel_norm - CONSTANTS_ONE_G;
	const float accel_ratio_abs = math::min(fabsf(delta_accel), 1.f / attenuation) / CONSTANTS_ONE_G;
	_ahrs_accel_fusion_gain = _params.EKFGSF_tilt_gain * sq(1.f - attenuation * accel_ratio_abs);
}

Comment (bresch) [reply EKF/EKFGSF_yaw_estimator.cpp]: Comparison with the other implementation : https://www.desmos.com/calculator/ycph8sadhx
Also note that in the previous implementation, when `_ahrs_accel_norm < 0.5 g` or ` _ahrs_accel_norm > 1.5 g`, the `_ahrs_accel_fusion_gain` isn't updated, so a spike can be fused with a high gain.

Comment (priseborough) [reply EKF/EKFGSF_yaw_estimator.cpp]: We need to be able to do tilt correction with 45 degree banked turns + some gust margin when operating as a fixed wing. These settings are working OK for replay of the logs we have, but we may end up adjusting the logic as we get more data.

File: EKF/EKFGSF_yaw_estimator.cpp [hunk 15]
```diff
@@ -0,0 +1,827 @@
...
+		const float upper_accel_limit = CONSTANTS_ONE_G * 1.1f;
+		const float lower_accel_limit = CONSTANTS_ONE_G * 0.9f;
+		const bool ok_to_align = ((accel_norm_sq > lower_accel_limit * lower_accel_limit &&
+			  accel_norm_sq < upper_accel_limit * upper_accel_limit));
+		if (ok_to_align) {
+			initialiseEKFGSF();
+			alignQuatEKFGSF();
+			_ahrs_ekf_gsf_tilt_aligned = true;
+		}
+		return;
```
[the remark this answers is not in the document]
priseborough: Agree that that is the preferred solution, however I had a go at doing that at the time of writing this code and ran into compiler errors that I didn't know how to resolve, so have structured it the way you see here so that the algorithm can be properly tested.

If someone wants to provide assistance with converting this to a separate class, then that would be good.

EKF/control.cpp
priseborough: See https://github.com/PX4/ecl/pull/751/commits/a27635e18420763481439e0a0b14e28dd9377198

priseborough:
> Can't we use cheap `wrap_pi` functions instead of expensive `cos`, `sin` and `atan2`?

I've been unable to find/devise an algorithm that will achieve that goal.

EKF/EKFGSF_yaw_estimator.cpp
bresch: @priseborough To show you that the asymmetric curve is also handled: https://www.desmos.com/calculator/kq95hay45h
DeepinScreenshot_select-area_20200220135708

jkflying:
> Can't we use cheap wrap_pi functions instead of expensive cos, sin and atan2?

It's actually a 2D problem because we are taking into account the weight of each as magnitude. So we can't simplify to 1D.

EKF/EKFGSF_yaw_estimator.cpp
priseborough: I've fixed the bug that can leave the gain at the previous value and have moved the gain calculation into a function. See https://github.com/PX4/ecl/pull/751/commits/340df697ed2e8ba3d9f2aa289dc1dd3d4d462a2e

bresch:
>    Can't we use cheap wrap_pi functions instead of expensive cos, sin and atan2?

> It's actually a 2D problem because we are taking into account the weight of each as magnitude. So we can't simplify to 1D.
   > edit: We could simplify if we only had 2 measurements. Otherwise some kind of combinatorial algorithm might work, but I don't think that would be faster than N x sin&cos + 1 x atan2

I think that I found a way to do the weighted wrapping, I'll make a PR later.

priseborough:
@bresch that will be good but don't prioritise it too high as it only saves N sinf + N cosf + 1 atan2f per GPS update (5 Hz). The area to focus on are the conversions between rotation matrix, Tait-Bryan rotation sequence  and quaternions that occur every GPS update. There is also the conversion from quaternion to rotation matrix that is performed every IMU update (125 Hz) cycle. 

I have some ideas about how we can improve efficiency of these calculations which will be explored next week.

priseborough:
The costs of these conversions varies - see below, but there is too much unnecessary conversion to and from quaternions so I experimented with  eliminating them and replaced the propagation of attitude with a direct approach using the rotation matrix. This hasn't been benchmarked yet, but should provide an improvement. 

quaternion -> rotation matrix = 31 floats (floats being *,+,-)
rotation matrix -> quaternion = 13 floats + 1 sqrt + 1 divide
rotation matrix -> Tait-Bryan 321 = 2 x atan2 + 1 x asin
Tait-Bryan 321 -> rotation matrix = 3 x sine + 3 x cosine + 24 floats
Tait-Bryan 321 -> quaternion = 6 x divide + 20 x floats + 3 x sine + 3 x cosine

P.S. The change has been reverted because there was a difference in the weightings, see before:

replay_before

and after:

replay_after

priseborough:
The inverse sqrt operation in the rotation matrix renormalisation can be replaced with a linear approximation with no detectable change in behaviour due to the row length being close to 1.

See replay test result below

Screen Shot 2020-02-23 at 11 05 38 am

priseborough:
Removal of quaternion operations has been effective - see https://github.com/PX4/ecl/pull/754#issuecomment-590034825

priseborough:
After the last set of commit's I am now out of ideas for efficiency improvements (excluding reducing the number of models used).

Now would be a good time to get some help with moving it into a class.

priseborough:
The algorithm development and testing is now complete so this PR has now served its purpose. Work will now continue with the PR against master here https://github.com/PX4/ecl/pull/754 which will include work to move the yaw estimator into a separate class.

PR #752, opened 2020-02-13
do not zero gps altitude reference in init function

RomanBapst:
- if gps checks pass before the filter initialized this will set the gps
reference altitude to zero and thus global altitude will not be correct.
I found that this was actually the case in SITL because there the gps checks pass really quickly.

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -184,7 +184,6 @@ bool Ekf::initialiseFilter()
 
 		// reset variables that are shared with post alignment GPS checks
 		_gps_pos_deriv_filt(2) = 0.0f;
-		_gps_alt_ref = 0.0f;
 
 		if(!initialiseTilt()){
 			return false;


EKF/ekf.cpp
```diff
@@ -184,7 +184,6 @@ bool Ekf::initialiseFilter()
 
 		// reset variables that are shared with post alignment GPS checks
 		_gps_pos_deriv_filt(2) = 0.0f;
-		_gps_alt_ref = 0.0f;
```
bresch: Shouldn't we put it in the reset function above?

RomanBapst: @bresch I'm not sure, I'd need to look at that. but you need to make sure that you re-run the gps checks after that so that the origin is set again.

RomanBapst: @bresch Do we support setting new origins nowdays? Maybe I missed that part.

bresch: No, we don't support new origins, it's just to reset all the variables in the reset function. It also contains `_NED_origin_initialised = false;` so it will have to pass the checks and set it again.
Not sure if there is any corner case that gets created by not resetting it to zero but I have the feeling that it's safer.

bresch:
Ping @RomanBapst

RomanBapst:
Replace by #840

PR #753, opened 2020-02-13
ekf: clean uncorrelateQuatStates function

bresch:
As the name can be ambiguous, it gets renamed "uncorrelateQuatFromOtherStates".

Also replace the loops storing the values and reapplying them by simply
zeroing two slices (this also saves 130B of flash).

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -457,7 +457,7 @@ bool Ekf::realignYawGPS()
 			// calculate new filter quaternion states using corrected yaw angle
 			_state.quat_nominal = Quatf(euler321);
 			_R_to_earth = Dcmf(_state.quat_nominal);
-			uncorrelateQuatStates();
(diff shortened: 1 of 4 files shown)


EKF/ekf_helper.cpp
```diff
@@ -1368,29 +1368,10 @@ void Ekf::fuse(float *K, float innovation)
...
-	// zero rows and columns
-	P.uncorrelateCovariance<4>(0);
-
-	// restore 4x4 elements
-	for (row = 0; row < 4; row++) {
-		for (col = 0; col < 4; col++) {
-			P(row, col) = variances[row][col];
-		}
-	}
+	P.slice<_k_num_states - 4, 4>(4, 0) = 0.f;
```
jkflying: This was a little confusing until I remembered the quaternion is the first group :sweat_smile: Otherwise you would have needed to write above _and_ below, for rows and columns.

PR #760, opened 2020-02-24
EKF: dont print wrong reset message on baro timeout

CarlOlsson:
Without this change the EKF will print
```
EKF baro hgt timeout - reset to GPS
EKF gps hgt timeout - reset to GPS
```

Directly after each other if the baro times out

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -824,10 +824,9 @@ void Ekf::controlHeightSensorTimeouts()
 				ECL_WARN_TIMESTAMPED("baro hgt timeout - reset to baro");
 
 			}
-		}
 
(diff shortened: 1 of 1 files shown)


CarlOlsson:
Would be nice if someone could merge this at some point

bresch:
Also, baro can reset to GPS, but EV_hgt, GPS and range finder can only reset to baro or themselves. I think they should all reset to any of the other ones, given a defined priority. 
Furthermore, rng and ev_hgt will reset to baro even if faulty (only GPS timeout checks for baro health...) and declare it non-faulty.
I'll add an issue to clean up this "failsafe" fail-over logic.

PR #761, opened 2020-02-25
EKF: dont check gps_check_fail_status is velocity reset

CarlOlsson:
This PR removes a bug that was introduced [here](https://github.com/PX4/ecl/commit/4511b9ff5e0a75c6eb4912b813a62f23ea2cd707) 

In a velocity reset we only used the GPS measurement if `_gps_check_fail_status.value` was equal to zero. The value of this flag is independent of `EKF2_GPS_CHECK` so checks can fail even if they are not configured to have any effect. E.g. on PX4 v1.10.1 the check will always fail if the vehicle is using two GPS modules since the gdop field is set to infinity. (On master gdop is replaced by pdop which is set correctly since https://github.com/PX4/Firmware/pull/13515/files)

My interpretation is that the GPS checks configured to run by `EKF2_GPS_CHECK` and defined by the corresponding threshold parameters are *preflight* checks and should have no influence other than that.

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -45,7 +45,7 @@
 #include <mathlib/mathlib.h>
 #include <cstdlib>
 
-// Reset the velocity states. If we have a recent and valid
+// Reset the velocity states. If we have a recent
 // gps measurement then use for velocity initialisation
 bool Ekf::resetVelocity()
 {
@@ -53,7 +53,7 @@ bool Ekf::resetVelocity()
 	Vector3f vel_before_reset = _state.vel;
 
 	// reset EKF states
-	if (_control_status.flags.gps && _gps_check_fail_status.value==0) {
+	if (_control_status.flags.gps) {
 		ECL_INFO_TIMESTAMPED("reset velocity to GPS");
 		// this reset is only called if we have new gps data at the fusion time horizon
 		_state.vel = _gps_sample_delayed.vel;
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -45,7 +45,7 @@
 #include <mathlib/mathlib.h>
 #include <cstdlib>
 
-// Reset the velocity states. If we have a recent
+// Reset the velocity states. If we have a recent and valid
 // gps measurement then use for velocity initialisation
 bool Ekf::resetVelocity()
 {
@@ -53,7 +53,7 @@ bool Ekf::resetVelocity()
 	Vector3f vel_before_reset = _state.vel;
 
 	// reset EKF states
-	if (_control_status.flags.gps) {
+	if (_control_status.flags.gps && isTimedOut(_last_gps_fail_us, (uint64_t)_min_gps_health_time_us)) {
 		ECL_INFO_TIMESTAMPED("reset velocity to GPS");
 		// this reset is only called if we have new gps data at the fusion time horizon
 		_state.vel = _gps_sample_delayed.vel;


kamilritz:
This was added then with the intent of stopping resets to bad GPS velocities after flying into GPS denied environments, where we still have other sources of aiding.

CarlOlsson:
> This was added then with the intent of stopping resets to bad GPS velocities after flying into GPS denied environments, where we still have other sources of aiding.

Yeah, but the EKF will still continuously fuse the GPS from the next sample so the gain is marginal if any. So I would propose to merge this and handle the case of GPS denied environments differently, probably more sophisticated logic is needed there.

kamilritz:
> Yeah, but the EKF will still continuously fuse the GPS from the next sample so the gain is marginal if any. So I would propose to merge this and handle the case of GPS denied environments differently, probably more sophisticated logic is needed there.

In my case the difference was not marginal. I was flying into a tunnel. While hovering in place, the EKF reset to the falsely non-zero GPS velocity, causing it to the fly into the wall.

For me this seems to be a more general discussion about what the GPS checks are. I understand that historically this was only thought as a preflight check, but IMHO this should also be used in-flight. If not, I suggest to stop performing the checks also after takeoff.

For now I would rather change it to `&& (_gps_check_fail_status.value==0 && "EKF_GPS_CHECK active"` as an intermediate solution.

CarlOlsson:
> In my case the difference was not marginal. I was flying into a tunnel. While hovering in place, the EKF reset to the falsely non-zero GPS velocity, causing it to the fly into the wall.

Sorry to hear that. But wouldn't it still have flown into the wall since the gps data would have been fused into the EKF directly afterwards? Or was the data so bad that the innovation checks would have been failing.

> For me this seems to be a more general discussion about what the GPS checks are. I understand that historically this was only thought as a preflight check, but IMHO this should also be used in-flight. If not, I suggest to stop performing the checks also after takeoff.

I agree, but that will require more work since we need to add different thresholds for when to pass preflight checks and when to allow data into the EKF. And also add configuration options to set what to check in preflight and what checks need to pass for the data to be fused.

kamilritz:
> But wouldn't it still have flown into the wall since the gps data would have been fused into the EKF directly afterwards? Or was the data so bad that the innovation checks would have been failing.

Not in my case, since I had a VIO sensor proving valid odometry information. After a reset the EKF would still have been vulnerable to bad GPS data. Probably it would have fused it for a short amount of time as the variance was increased during the reset. But while converging to the VIO solution it would not be affected that much by bad GPS data. 
In comparison while reseting to the unrealistically high velocity, the delta velocity biases would also change and the filter would need much longer to converge towards the more valid VIO data, if it even does. More likely it will reject the VIO data.

To the best of my knowledge, I was testing this additional check on the replay log of the crash in the tunnel and it stopped it from diverging after the reset.

Would you be okay with the intermediate solution I have proposed?

CarlOlsson:
Okey thanks for the explanation!

> Would you be okay with the intermediate solution I have proposed?

Yes, you mean something like this?
```
if (_control_status.flags.gps && _time_last_imu - _last_gps_fail_us > (uint64_t)_min_gps_health_time_us) {
```

kamilritz:
Yes. But without the subtraction of two unsigned integers. xD

CarlOlsson:
hehe, okey I will update the PR and test it in replay

kamilritz:
Thanks a lot. I am sorry that I caused troubles with that additional check.

CarlOlsson:
> Thanks a lot. I am sorry that I caused troubles with that additional check.

Oh no problem! It's cool to see more contributors to ECL, with more diversity in the projects used by ECL these things are inevitable to happen. Good that we have replay so we can catch it in simulations :)

CarlOlsson:
v1.10.1 vs PR
image

Issue #667, opened 2019-11-08
Rewrite EKF symbolic derivations in sympy


CarlOlsson:
Just realized that I can't run `GenerateNavFilterEquations.m` since I don't have access to the MATLAB symbolic toolbox anymore.
I guess I am not alone so we should rewrite it in sympy
@RomanBapst I remember you mentioning that you already started?

mhkabir:
This would be awesome.

RomanBapst:
@CarlOlsson Yep, I can share what I've started.

bresch:
@RomanBapst Can you share what you currently have?

dagar:
Done!

PR #762, opened 2020-02-27
WIP: added python script with ekf derivation

RomanBapst:
### Motivation
Some time ago I decided to implement the 24 state EKF filter equations in python for the following reasons:

- I do not have a Matlab license anymore
- I wanted to be able to derive the equations quickly and not wait for 3 hours every time I change something
- I wanted to use this in combination with an editor which allows me to interact with the equations once they were generated

### What's the state of this
So far I have implemented:

- definition of states and state propagation
- derivation of covariance matrix prediction and simplification (most interesting one)
- derivation of 3D mag fusion
- derivation of vel/pos fusion

I used simpy's [cse](https://docs.sympy.org/latest/modules/rewriting.html#common-subexpression-detection-and-collection) for doing common subexpression detection.
My initial observation with this comparing to the brute force algorithm implemented in Matlab is:

- it's very fast, simplifying the covariance matrix update takes between 10 and 20 seconds
- it produces more intermediate variables than the matlab version
- it produces code with less mathematical operations (+, -, *)

### Verification of the equations
In order to test the correctness of the equations I numerically evaluated the predicted covariance matrix using random numbers. I then summed each row of the covariance matrix and compared the values between the two derivations (c++ code generated by matlab and the python code).
The values look exactly the same as seen in the figure below.
So I'm currently assuming that the derivation is mathematically correct and there are no big mistakes in the code anymore. This was just some initial experiment though, this needs to be done properly moving forward.
image

Update:
Compared the executing time of the covariance matrix update between this derivation here and the one from matlab. I ran both versions for 10000 times in a loop side by side and evaluated the performance using calgrind. The resulting graph can be seen below:
image

I also noticed some slight differences in the numerical result of the predicted covariance matrix between the two derivations. The figure below shows two matrices:
1) The difference between the predicted matrices, e.g. `P_next_matlab - P_next_python` multiplied by a factor of 1e6 in order to make the error more visible.
2) The second matrix shows the same error but relative  to `P_next - P` in percent. In other words, "how big are those differences in magnitude compared to the delta you add to `P` in order to arrive at `P_next` "
I wanted to figure out if the small differences seen in 1) are negligible, and it seems like they are.
However, it would still be nice to figure out where those differences come from.
image

**Update**
Converted all floats to double in both derivations and the errors are gone!

### Editor
I was using [spyder](https://www.spyder-ide.org/) which allows you to interact with the script.

fixes https://github.com/PX4/ecl/issues/667

### DIFF
--- a/EKF/python/ekf_derivation/code_gen.py
+++ b/EKF/python/ekf_derivation/code_gen.py
@@ -12,6 +12,9 @@ class CodeGenerator:
     def __init__(self, file_name):
         self.file_name = file_name
         self.file = open(self.file_name, 'w', buffering=0)
+
+    def print_string(self, string):
+        self.file.write("// " + string + "\n")
         
     def get_ccode(self, expression):
         return ccode(expression, type_aliases={real:float32})
@@ -20,14 +23,16 @@ class CodeGenerator:
         write_string = ""
         for item in subexpressions:
             write_string = write_string + "float " + str(item[0]) + " = " + self.get_ccode(item[1]) + ";\n"
-        
+
         write_string = write_string + "\n\n"
         self.file.write(write_string)
         
     def write_matrix(self, matrix, identifier, is_symmetric=False):
         write_string = ""
-        
-        if matrix.shape[0] == 1 or matrix.shape[1] == 1:
+
+        if matrix.shape[0] * matrix.shape[1] == 1:
+            write_string = write_string + identifier + " = " + self.get_ccode(matrix[0]) + ";\n"
+        elif matrix.shape[0] == 1 or matrix.shape[1] == 1:
             for i in range(0,len(matrix)):
                 write_string = write_string + identifier + "(" + str(i) + ") = " + self.get_ccode(matrix[i]) + ";\n"
         else:
--- a/EKF/python/ekf_derivation/code_gen.py
+++ b/EKF/python/ekf_derivation/code_gen.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+"""
+Created on Sat Mar 14 12:47:24 2020
+
+@author: roman
+"""
+from sympy import ccode
+from sympy.codegen.ast import float32, real
+
+class CodeGenerator:
+    def __init__(self, file_name):
+        self.file_name = file_name
+        self.file = open(self.file_name, 'w', buffering=0)
+        
+    def get_ccode(self, expression):
+        return ccode(expression, type_aliases={real:float32})
+        
+    def write_subexpressions(self,subexpressions):
+        write_string = ""
+        for item in subexpressions:
+            write_string = write_string + "float " + str(item[0]) + " = " + self.get_ccode(item[1]) + ";\n"
+        
+        write_string = write_string + "\n\n"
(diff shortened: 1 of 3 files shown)


jkflying:
Awesome! Could you put the two outputs (I guess for the covariance, since that is most performance critical) into functions in C++ files, and run them in callgrind? Then we can see what the actual instruction count differences are.

RomanBapst:
@jkflying Yes, I'm already on this.

RomanBapst:
Update: I cleaned up the code a bit and pushed everything to this branch. If you run main.py you will get autogenerated cpp files in the same directory.

I also tested the covariance prediction on a Pixhawk cube and got the following results:

- perf showed that the ekf2 average execution time dropped slightly but not significantly (850us vs 811us avg) on an F4
- saved about 3.1K of flash on an F4

EKF/python/ekf_derivation/code_gen.py
```diff
@@ -0,0 +1,49 @@
...
+            write_string = write_string + identifier + " = " + self.get_ccode(matrix[0]) + ";\n"
+        elif matrix.shape[0] == 1 or matrix.shape[1] == 1:
+            for i in range(0,len(matrix)):
+                write_string = write_string + identifier + "(" + str(i) + ") = " + self.get_ccode(matrix[i]) + ";\n"
+        else:
+            #write_string = "float " + identifier + "[" + str(matrix.shape[0]) + "][" + str(matrix.shape[1]) + "] = {};\n"
+            for j in range(0, matrix.shape[1]):
+                for i in range(0, matrix.shape[0]):
+                    if j >= i or not is_symmetric:
+                        write_string = write_string + identifier + "(" + str(i) + "," + str(j) + ") = " + self.get_ccode(matrix[i,j]) + ";\n"
```
jkflying, line 43: These all need square brackets now, also above
```suggestion
                        write_string = write_string + identifier + "[" + str(i) + "," + str(j) + "] = " + self.get_ccode(matrix[i,j]) + ";\n"
```

dusan19:
> Update
> Converted all floats to double in both derivations and the errors are gone!

The results you got when using doubles (same with both matlab and python equations), are they closer to matlab float results or to python float results?
Assuming the differences come from float precision limitations and that results obtained with doubles are the ground truth, this comparison would tell us if the python equations make the numerical situation better or worse than matlab equations. Hopefully it would just be a verification that python generated equations using floats are closer to ground truth (doubles).

priseborough:
Can you do something about the powf( arg,2) instances in the generated code. If not we can do a find and replace with sq(arg) in the editor.

bresch:
@RomanBapst Could you please follow up on this?

dusan19:
I managed to create the log from boot where the estimator crashes every 15 min for 2hrs, and replayed it with the changes from this PR.
here are the logs of the original log replay, and the replay with this PR, respectively.
The logs show that the python generated equations do not solve the problem, the estimator still crashes, but the crashing is less often, first time it happened after more than an hour from boot

Original:

orig_replay

With python generated equations:

python_cov

priseborough:
This concept has been developed further by https://github.com/PX4/ecl/pull/870 and subsequent PR's

Issue #759, opened 2020-02-24
Earth rate compensation


Slimercorp:
Here there is bug.  https://github.com/PX4/ecl/blob/71d4d22ae4db0c89d96890ce37b5e56fd4f4cfb6/EKF/ekf.cpp#L249 
Result is 2 * Earth rate in corrected_delta_ang. 
Must be
`corrected_delta_ang -= _R_to_earth.transpose() * _earth_rate_NED * _imu_sample_delayed.delta_ang_dt;`

bresch:
https://github.com/PX4/ecl/commit/ffea65e1a4168d7ec20084dc7934be7b5ac13a94

priseborough:
Must have been brain fade. It should be 

corrected_delta_ang -= _R_to_earth.transpose() * _earth_rate_NED * _imu_sample_delayed.delta_ang_dt;

PR #763, opened 2020-02-27
EKF: Fix sign error in earth rotation rate correction

priseborough:
Addresses issue https://github.com/PX4/ecl/issues/759
fixes https://github.com/PX4/ecl/issues/759


priseborough:
Thank you https://github.com/Slimercorp for spotting this

bresch:
@priseborough  The change indicator is supposed to fail here, could you push a new version of the csv generated by `make test` please?

kamilritz:
> The change indicator is supposed to fail here, could you push a new version of the csv generated by make test please?

@bresch Cool, that shows that the change indicator is catching small changes.

PR #754, opened 2020-02-13
EKF: Add Emergency yaw recovery using EKF-GSF estimator

priseborough:
This is functionally the same as https://github.com/PX4/ecl/pull/751 and any changes identified during review and testing of https://github.com/PX4/ecl/pull/751 will be cherry-picked into this branch.

Testing used https://github.com/priseborough/Firmware/tree/pr-yawEKFGSF PX4/Firmware version.

Flight testing of this version to follow, but here is replay output using https://logs.px4.io/plot_app?log=d1e5d8e4-5c44-4350-b355-ee3acb2503b8 where a navigation failure was induced by taking off with the declination set to an error of 100 degrees. This can be achieved by setting EKF2_MAG_DECL to a large value at least 90 degrees away from the declination at the test location and EKF2_DECL_TYPE = 0 to force the EKF to use the parameter value. 

If you do testing make sure you have a non-position control mode like altitude or manual/stabilized available.

Screen Shot 2020-02-14 at 7 32 14 am

PR #764, opened 2020-02-28
EKF: Add Emergency yaw recovery using EKF-GSF estimator

priseborough:
Replaces https://github.com/PX4/ecl/pull/754

The estimator has been moved into its own class. Some other minor efficiency improvements have been made.

Replay of a flyaway test log shows the same performance .

Before (with N_MODELS_EKFGSF set to 5)
before

After:
after

This was tested with the following firmware changes. Note - an externally requested yaw reset has not been tested and may add unnecessary complexity so the commander changes and the ekf2_main call to Ekf::requestEmergencyNavReset is probably unnecessary.

https://github.com/priseborough/Firmware/tree/pr-yawEKFGSF-refactor

### DIFF
--- a/EKF/EKFGSF_yaw.cpp
+++ b/EKF/EKFGSF_yaw.cpp
@@ -112,8 +112,8 @@ void EKFGSF_yaw::update(const Vector3f del_ang, // IMU delta angle rotation vect
 	// equal to the weighting value before it is summed.
 	Vector2f yaw_vector = {};
 	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
-		yaw_vector(0) += _ekf_gsf[model_index].W * cosf(_ekf_gsf[model_index].X[2]);
-		yaw_vector(1) += _ekf_gsf[model_index].W * sinf(_ekf_gsf[model_index].X[2]);
+		yaw_vector(0) += _ekf_gsf[model_index].W * cosf(_ekf_gsf[model_index].X(2));
+		yaw_vector(1) += _ekf_gsf[model_index].W * sinf(_ekf_gsf[model_index].X(2));
 	}
 	_gsf_yaw = atan2f(yaw_vector(1),yaw_vector(0));
 
@@ -121,7 +121,7 @@ void EKFGSF_yaw::update(const Vector3f del_ang, // IMU delta angle rotation vect
 	// models with larger innovations are weighted less
 	_gsf_yaw_variance = 0.0f;
 	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
-		float yaw_delta = wrap_pi(_ekf_gsf[model_index].X[2] - _gsf_yaw);
+		float yaw_delta = wrap_pi(_ekf_gsf[model_index].X(2) - _gsf_yaw);
 		_gsf_yaw_variance += _ekf_gsf[model_index].W * (_ekf_gsf[model_index].P(2,2) + yaw_delta * yaw_delta);
 	}
 
@@ -239,7 +239,7 @@ void EKFGSF_yaw::ahrsAlignYaw()
 			Eulerf euler_init(_ahrs_ekf_gsf[model_index].R);
 
 			// set the yaw angle
-			euler_init(2) = wrap_pi(_ekf_gsf[model_index].X[2]);
+			euler_init(2) = wrap_pi(_ekf_gsf[model_index].X(2));
 
 			// update the rotation matrix
 			_ahrs_ekf_gsf[model_index].R = Dcmf(euler_init);
@@ -247,7 +247,7 @@ void EKFGSF_yaw::ahrsAlignYaw()
 		} else {
 			// Calculate the 312 Tait-Bryan rotation sequence that rotates from earth to body frame
 			Vector3f rot312;
-			rot312(0) = wrap_pi(_ekf_gsf[model_index].X[2]); // first rotation (yaw) taken from EKF model state
+			rot312(0) = wrap_pi(_ekf_gsf[model_index].X(2)); // first rotation (yaw) taken from EKF model state
 			rot312(1) = asinf(_ahrs_ekf_gsf[model_index].R(2, 1)); // second rotation (roll)
 			rot312(2) = atan2f(-_ahrs_ekf_gsf[model_index].R(2, 0), _ahrs_ekf_gsf[model_index].R(2, 2));  // third rotation (pitch)
 
@@ -272,20 +272,20 @@ void EKFGSF_yaw::predictEKF(const uint8_t model_index)
 	// Calculate the yaw state using a projection onto the horizontal that avoids gimbal lock
 	if (fabsf(_ahrs_ekf_gsf[model_index].R(2, 0)) < fabsf(_ahrs_ekf_gsf[model_index].R(2, 1))) {
 		// use 321 Tait-Bryan rotation to define yaw state
-		_ekf_gsf[model_index].X[2] = atan2f(_ahrs_ekf_gsf[model_index].R(1, 0), _ahrs_ekf_gsf[model_index].R(0, 0));
+		_ekf_gsf[model_index].X(2) = atan2f(_ahrs_ekf_gsf[model_index].R(1, 0), _ahrs_ekf_gsf[model_index].R(0, 0));
 	} else {
 		// use 312 Tait-Bryan rotation to define yaw state
-		_ekf_gsf[model_index].X[2] = atan2f(-_ahrs_ekf_gsf[model_index].R(0, 1), _ahrs_ekf_gsf[model_index].R(1, 1)); // first rotation (yaw)
+		_ekf_gsf[model_index].X(2) = atan2f(-_ahrs_ekf_gsf[model_index].R(0, 1), _ahrs_ekf_gsf[model_index].R(1, 1)); // first rotation (yaw)
 	}
 
 	// calculate delta velocity in a horizontal front-right frame
 	const Vector3f del_vel_NED = _ahrs_ekf_gsf[model_index].R * _delta_vel;
-	const float dvx =   del_vel_NED(0) * cosf(_ekf_gsf[model_index].X[2]) + del_vel_NED(1) * sinf(_ekf_gsf[model_index].X[2]);
-	const float dvy = - del_vel_NED(0) * sinf(_ekf_gsf[model_index].X[2]) + del_vel_NED(1) * cosf(_ekf_gsf[model_index].X[2]);
+	const float dvx =   del_vel_NED(0) * cosf(_ekf_gsf[model_index].X(2)) + del_vel_NED(1) * sinf(_ekf_gsf[model_index].X(2));
+	const float dvy = - del_vel_NED(0) * sinf(_ekf_gsf[model_index].X(2)) + del_vel_NED(1) * cosf(_ekf_gsf[model_index].X(2));
 
 	// sum delta velocities in earth frame:
-	_ekf_gsf[model_index].X[0] += del_vel_NED(0);
-	_ekf_gsf[model_index].X[1] += del_vel_NED(1);
+	_ekf_gsf[model_index].X(0) += del_vel_NED(0);
+	_ekf_gsf[model_index].X(1) += del_vel_NED(1);
 
 	// predict covariance - autocode from https://github.com/priseborough/3_state_filter/blob/flightLogReplay-wip/calcPupdate.txt
 
@@ -308,8 +308,8 @@ void EKFGSF_yaw::predictEKF(const uint8_t model_index)
 	float dazVar = _gyro_noise * _delta_ang_dt; // variance of yaw delta angle - rad^2
 	dazVar *= dazVar;
 
-	const float t2 = sinf(_ekf_gsf[model_index].X[2]);
-	const float t3 = cosf(_ekf_gsf[model_index].X[2]);
+	const float t2 = sinf(_ekf_gsf[model_index].X(2));
+	const float t3 = cosf(_ekf_gsf[model_index].X(2));
 	const float t4 = dvy*t3;
 	const float t5 = dvx*t2;
 	const float t6 = t4+t5;
@@ -347,8 +347,8 @@ void EKFGSF_yaw::updateEKF(const uint8_t model_index)
 	velObsVar *= velObsVar;
 
 	// calculate velocity observation innovations
-	_ekf_gsf[model_index].innov(0) = _ekf_gsf[model_index].X[0] - _vel_NE(0);
-	_ekf_gsf[model_index].innov(1) = _ekf_gsf[model_index].X[1] - _vel_NE(1);
+	_ekf_gsf[model_index].innov(0) = _ekf_gsf[model_index].X(0) - _vel_NE(0);
+	_ekf_gsf[model_index].innov(1) = _ekf_gsf[model_index].X(1) - _vel_NE(1);
 
 	// copy covariance matrix to temporary variables
 	const float P00 = _ekf_gsf[model_index].P(0,0);
@@ -408,14 +408,14 @@ void EKFGSF_yaw::updateEKF(const uint8_t model_index)
 	}
 	const float t8 = P11+velObsVar;
 	const float t10 = P00+velObsVar;
-	float K[3][2];
 
- 	K[0][0] = -P01*P10*t7+P00*t7*t8;
-	K[0][1] = -P00*P01*t7+P01*t7*t10;
-	K[1][0] = -P10*P11*t7+P10*t7*t8;
-	K[1][1] = -P01*P10*t7+P11*t7*t10;
-	K[2][0] = -P10*P21*t7+P20*t7*t8;
-	K[2][1] = -P01*P20*t7+P21*t7*t10;
+	matrix::Matrix<float, 3, 2> K;
+ 	K(0,0) = -P01*P10*t7+P00*t7*t8;
+	K(0,1) = -P00*P01*t7+P01*t7*t10;
+	K(1,0) = -P10*P11*t7+P10*t7*t8;
+	K(1,1) = -P01*P10*t7+P11*t7*t10;
+	K(2,0) = -P10*P21*t7+P20*t7*t8;
+	K(2,1) = -P01*P20*t7+P21*t7*t10;
 
 	const float t11 = P00*P01*t7;
 	const float t15 = P01*t7*t10;
@@ -467,14 +467,10 @@ void EKFGSF_yaw::updateEKF(const uint8_t model_index)
 	// force symmetry
 	_ekf_gsf[model_index].P.makeBlockSymmetric<3>(0);
 
-	float oldYaw = _ekf_gsf[model_index].X[2];
-	for (uint8_t obs_index = 0; obs_index < 2; obs_index++) {
-		// apply the state corrections including the compression scale factor
-		for (unsigned row = 0; row < 3; row++) {
-			_ekf_gsf[model_index].X[row] -= K[row][obs_index] * _ekf_gsf[model_index].innov(obs_index) * innov_comp_scale_factor;
-		}
-	}
-	float yawDelta = _ekf_gsf[model_index].X[2] - oldYaw;
+	// Correct the state vector and capture the change in yaw angle
+	float oldYaw = _ekf_gsf[model_index].X(2);
+	_ekf_gsf[model_index].X -= (K * _ekf_gsf[model_index].innov) * innov_comp_scale_factor;
+	float yawDelta = _ekf_gsf[model_index].X(2) - oldYaw;
 
 	// apply the change in yaw angle to the AHRS
 	// take advantage of sparseness in the yaw rotation matrix
@@ -500,14 +496,14 @@ void EKFGSF_yaw::initialiseEKFGSF()
 	const float yaw_increment = 2.0f * M_PI_F / (float)N_MODELS_EKFGSF;
 	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index++) {
 		// evenly space initial yaw estimates in the region between +-Pi
-		_ekf_gsf[model_index].X[2] = -M_PI_F + (0.5f * yaw_increment) + ((float)model_index * yaw_increment);
+		_ekf_gsf[model_index].X(2) = -M_PI_F + (0.5f * yaw_increment) + ((float)model_index * yaw_increment);
 
 		// All filter models start with the same weight
 		_ekf_gsf[model_index].W = 1.0f / (float)N_MODELS_EKFGSF;
 
 		// take velocity states and corresponding variance from last meaurement
-		_ekf_gsf[model_index].X[0] = _vel_NE(0);
-		_ekf_gsf[model_index].X[1] = _vel_NE(1);
+		_ekf_gsf[model_index].X(0) = _vel_NE(0);
+		_ekf_gsf[model_index].X(1) = _vel_NE(1);
 		_ekf_gsf[model_index].P(0,0) = sq(_vel_accuracy);
 		_ekf_gsf[model_index].P(1,1) = _ekf_gsf[model_index].P(0,0);
 
@@ -555,7 +551,7 @@ bool EKFGSF_yaw::getLogData(float *yaw_composite, float *yaw_variance, float yaw
 		*yaw_composite = _gsf_yaw;
 		*yaw_variance = _gsf_yaw_variance;
 		for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index++) {
-			yaw[model_index] = _ekf_gsf[model_index].X[2];
+			yaw[model_index] = _ekf_gsf[model_index].X(2);
 			innov_VN[model_index] = _ekf_gsf[model_index].innov(0);
 			innov_VE[model_index] = _ekf_gsf[model_index].innov(1);
 			weight[model_index] = _ekf_gsf[model_index].W;
--- a/EKF/EKFGSF_yaw.cpp
+++ b/EKF/EKFGSF_yaw.cpp
@@ -649,7 +649,7 @@ bool EKFGSF_yaw::getYawData(float *yaw, float *yaw_variance)
 	return false;
 }
 
-void EKFGSF_yaw::pushVelData(Vector2f velocity, float accuracy)
+void EKFGSF_yaw::setVelocity(Vector2f velocity, float accuracy)
 {
 	_vel_NE = velocity;
 	_vel_accuracy = accuracy;
--- a/EKF/EKFGSF_yaw.cpp
+++ b/EKF/EKFGSF_yaw.cpp
@@ -347,7 +347,7 @@ void EKFGSF_yaw::updateEKF(const uint8_t model_index)
 	_ekf_gsf[model_index].innov(0) = _ekf_gsf[model_index].X(0) - _vel_NE(0);
 	_ekf_gsf[model_index].innov(1) = _ekf_gsf[model_index].X(1) - _vel_NE(1);
 
-	// Use temproary variables for covariance elements to reduce verbosity of auto-code expressions
+	// Use temporary variables for covariance elements to reduce verbosity of auto-code expressions
 	const float &P00 = _ekf_gsf[model_index].P(0,0);
 	const float &P01 = _ekf_gsf[model_index].P(0,1);
 	const float &P02 = _ekf_gsf[model_index].P(0,2);
@@ -359,32 +359,24 @@ void EKFGSF_yaw::updateEKF(const uint8_t model_index)
 	const float &P22 = _ekf_gsf[model_index].P(2,2);
 
 	// calculate innovation variance
-	_ekf_gsf[model_index].S[0][0] = P00 + velObsVar;
-	_ekf_gsf[model_index].S[1][1] = P11 + velObsVar;
-	_ekf_gsf[model_index].S[0][1] = P01;
-	_ekf_gsf[model_index].S[1][0] = P10;
+	_ekf_gsf[model_index].S(0,0) = P00 + velObsVar;
+	_ekf_gsf[model_index].S(1,1) = P11 + velObsVar;
+	_ekf_gsf[model_index].S(0,1) = P01;
+	_ekf_gsf[model_index].S(1,0) = P10;
+
+	// Update the inverse of the innovation covariance matrix S_inverse
+	updateInnovCovMatInv(model_index);
 
 	// Perform a chi-square innovation consistency test and calculate a compression scale factor that limits the magnitude of innovations to 5-sigma
-	float S_det_inv = (_ekf_gsf[model_index].S[0][0]*_ekf_gsf[model_index].S[1][1] - _ekf_gsf[model_index].S[0][1]*_ekf_gsf[model_index].S[1][0]);
 	float innov_comp_scale_factor = 1.0f;
-	if (fabsf(S_det_inv) > 1E-6f) {
-		// Calculate elements for innovation covariance inverse matrix assuming symmetry
-		S_det_inv = 1.0f / S_det_inv;
-		const float S_inv_NN = _ekf_gsf[model_index].S[1][1] * S_det_inv;
-		const float S_inv_EE = _ekf_gsf[model_index].S[0][0] * S_det_inv;
-		const float S_inv_NE = _ekf_gsf[model_index].S[0][1] * S_det_inv;
 
-		// The following expression was derived symbolically from test ratio = transpose(innovation) * inverse(innovation variance) * innovation = [1x2] * [2,2] * [2,1] = [1,1]
-		const float test_ratio = _ekf_gsf[model_index].innov(0)*(_ekf_gsf[model_index].innov(0)*S_inv_NN + _ekf_gsf[model_index].innov(1)*S_inv_NE) + _ekf_gsf[model_index].innov(1)*(_ekf_gsf[model_index].innov(0)*S_inv_NE + _ekf_gsf[model_index].innov(1)*S_inv_EE);
+	// test ratio = transpose(innovation) * inverse(innovation variance) * innovation = [1x2] * [2,2] * [2,1] = [1,1]
+	const float test_ratio = _ekf_gsf[model_index].innov * (_ekf_gsf[model_index].S_inverse * _ekf_gsf[model_index].innov);
 
-		// If the test ratio is greater than 25 (5 Sigma) then reduce the length of the innovation vector to clip it at 5-Sigma
-		// This protects from large measurement spikes
-		if (test_ratio > 25.0f) {
-			innov_comp_scale_factor = sqrtf(25.0f / test_ratio);
-		}
-	} else {
-		// skip this fusion step because calculation is badly conditioned
-		return;
+	// If the test ratio is greater than 25 (5 Sigma) then reduce the length of the innovation vector to clip it at 5-Sigma
+	// This protects from large measurement spikes
+	if (test_ratio > 25.0f) {
+		innov_comp_scale_factor = sqrtf(25.0f / test_ratio);
 	}
 
 	// calculate Kalman gain K  nd covariance matrix P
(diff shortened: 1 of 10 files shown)


EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1816,137 @@ void Ekf::stopFlowFusion()
...
+
+			_R_to_earth(0, 0) = c0 * c2 - s0 * s1 * s2;
+			_R_to_earth(1, 1) = c0 * c1;
+			_R_to_earth(2, 2) = c2 * c1;
+			_R_to_earth(0, 1) = -c1 * s0;
+			_R_to_earth(0, 2) = s2 * c0 + c2 * s1 * s0;
+			_R_to_earth(1, 0) = c2 * s0 + s2 * s1 * c0;
+			_R_to_earth(1, 2) = s0 * s2 - s1 * c0 * c2;
+			_R_to_earth(2, 0) = -s2 * c1;
+			_R_to_earth(2, 1) = s1;
```
bresch: This looks the same as 
- https://github.com/PX4/ecl/blob/3fa5f501ae7edc4a793f0973e156706afcc8b419/EKF/mag_fusion.cpp#L650-L665
- https://github.com/PX4/ecl/blob/3fa5f501ae7edc4a793f0973e156706afcc8b419/EKF/gps_yaw_fusion.cpp#L344-L360
- https://github.com/PX4/ecl/blob/3fa5f501ae7edc4a793f0973e156706afcc8b419/EKF/ekf_helper.cpp#L600-L627

Can we package that 312 to DCM into a helper function to avoid a fourth duplicate?

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1816,137 @@ void Ekf::stopFlowFusion()
...
+
+		// update quaternion states
+		_state.quat_nominal = quat_after_reset;
+		uncorrelateQuatFromOtherStates();
+
+		// record the state change
+		_state_reset_status.quat_change = q_error;
+
+		// update transformation matrix from body to world frame using the current estimate
+		_R_to_earth = Dcmf(_state.quat_nominal);
```
bresch: This look the same as:
- https://github.com/PX4/ecl/blob/3fa5f501ae7edc4a793f0973e156706afcc8b419/EKF/ekf_helper.cpp#L688-L697
- https://github.com/PX4/ecl/blob/3fa5f501ae7edc4a793f0973e156706afcc8b419/EKF/gps_yaw_fusion.cpp#L366-L388
But not with the same exact order; is that by design? Can't we unify most of the "yaw reset" logic? Why are we computing 
```c++
float scalar; 
  
 if (q_error(0) >= 0.0f) { 
 	scalar = -2.0f; 
  
 } else { 
 	scalar = 2.0f; 
 } 
  
 delta_ang_error(0) = scalar * q_error(1); 
 delta_ang_error(1) = scalar * q_error(2); 
 delta_ang_error(2) = scalar * q_error(3); 
```
In `gps_yaw_fusion` but not in the other ones?

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1816,137 @@ void Ekf::stopFlowFusion()
...
+
+		// reset the rotation from the EV to EKF frame of reference if it is being used
+		if ((_params.fusion_mode & MASK_ROTATE_EV) && !_control_status.flags.ev_yaw) {
+			calcExtVisRotMat();
```
bresch: Why are we doing that here? It's not done in any of the other "yaw reset" functions.
It's also done at each loop, before each EV fusion : https://github.com/PX4/ecl/blob/3fa5f501ae7edc4a793f0973e156706afcc8b419/EKF/control.cpp#L186-L189 so it would just do it a 2nd time, right?

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1816,137 @@ void Ekf::stopFlowFusion()
...
+		// apply the change in attitude quaternion to our newest quaternion estimate
+		// which was already taken out from the output buffer
+		_output_new.quat_nominal = _state_reset_status.quat_change * _output_new.quat_nominal;
+
+		// capture the reset event
+		_state_reset_status.quat_counter++;
+
+		// reset velocity and position states to GPS - if yaw is fixed then the filter should start to operate correctly
+		resetVelocity();
+		resetPosition();
```
bresch: In `realignYawGPS` we're also requesting a vel/pos reset but through a flag: https://github.com/PX4/ecl/blob/3fa5f501ae7edc4a793f0973e156706afcc8b419/EKF/ekf_helper.cpp#L463 , why not here?

EKF/ekf_helper.cpp
priseborough: Done

EKF/ekf_helper.cpp
priseborough: I think that's a result of developing against stable and then rebasing the code against master. Will remove.

EKF/ekf_helper.cpp
priseborough: This is at the end of the processing chain for an EKF fusion time horizon update and i was worried about a race condition with other processes that could call a position or velocity timeout. We can experiment with doing that n replay and see if it matters.

priseborough: This code also ensures the reset will happen on this frame rather than waiting for the next GPS message which could be up to 200msec later.

EKF/ekf_helper.cpp
priseborough: Unification is something I would like to do, but would rather tackle it in a separate PR. The code we have for the current reset to EKF-GSF yaw is working reliably based on testing to date, so I would be reluctant to change it until we fully understand the consequences of doing so.

priseborough: The idea behind the delta_ang_error check was to reset the states and covariances only if the yaw change was significant and was intended to the used for the types of resets we do every flight, such as the in-air reset after takeoff when the vehicle has cleared groudn based magnetic interference.

In the case of an emergency yaw reset we work on the assumption that a hard reset is necessary so no need to do that check.

priseborough:
Flight testing was repeated with a total of 9 takeoffs with multi-rotor. This included takeoffs with immediate post takeoff switch to auto mode. Th mission plan loaded with a representative survey type mission plan loaded to determine how the main EKF and EKF-GSF would manage subsequent flight with longer non-manoeuvring segments without use of the magnetometer. The flight vehicle was observed to start to flyaway with a visible stop when the reset was commanded. It then continued the mission. Yaw estimation during survey legs was stable.

https://logs.px4.io/plot_app?log=bdbd4248-28de-4ce1-8a57-bbf5fb55b6e4

Yaw for the whole flight:
Screen Shot 2020-02-29 at 1 21 44 pm

Yaw showing the reset event:
Screen Shot 2020-02-29 at 1 22 26 pm

GPS velocity innovations for the whole flight:
Screen Shot 2020-02-29 at 1 23 53 pm

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1790,141 @@ void Ekf::stopFlowFusion()
...
+{
+	// don't allow reet using the EKF-GSF estimate until the filter has started fusing velocity
+	// data and the yaw estimate has converged
+	float new_yaw, new_yaw_variance;
+	if (yawEstimator.getYawData(&new_yaw, &new_yaw_variance) && new_yaw_variance < sq(_params.EKFGSF_yaw_err_max)) {
+		// save a copy of the quaternion state for later use in calculating the amount of reset change
+		Quatf quat_before_reset = _state.quat_nominal;
+
+		// update transformation matrix from body to world frame using the current estimate
+		_R_to_earth = Dcmf(_state.quat_nominal);
```
kamilritz: Do we need to update the rotation matrix? Is it not synced with the quaternion?

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	// force symmetry
+	makeCovSymEKF(model_index);
+
+	float oldYaw = _ekf_gsf[model_index].X[2];
+	for (uint8_t obs_index = 0; obs_index < 2; obs_index++) {
+		// apply the state corrections including the compression scale factor
+		for (unsigned row = 0; row < 3; row++) {
+			_ekf_gsf[model_index].X[row] -= K[row][obs_index] * _ekf_gsf[model_index].innov[obs_index] * innov_comp_scale_factor;
+		}
+	}
```
kamilritz: I think this can be written as a matrix-vector multiplication

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+			innov_VN[model_index] = _ekf_gsf[model_index].innov[0];
+			innov_VE[model_index] = _ekf_gsf[model_index].innov[1];
+			weight[model_index] = _ekf_gsf[model_index].W;
+		}
+		return true;
+	}
+	return false;
+}
+
+void EKFGSF_yaw::makeCovSymEKF(const uint8_t model_index)
```
kamilritz: If you make P a `matrix:SquareMatrix(float, 3) you can make it symmetric by calling `P.makeBlockSymmetric<3>(0)`. Then would not need this function.

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+			_ahrs_accel_fusion_gain = 0.0f;
+		}
+	} else if (accel_g > 0.5f) {
+		_ahrs_accel_fusion_gain = _tilt_gain * sq(2.0f * accel_g - 1.0f);
+	} else {
+		_ahrs_accel_fusion_gain = 0.0f;
+	}
+}
+
+Matrix3f EKFGSF_yaw::updateRotMat(Matrix3f &R, Vector3f &g)
```
kamilritz: What is happening here? The function name is not very descriptive.

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	// Renormalise rows
+	float rowLengthSq;
+	for (uint8_t r = 0; r < 3; r++) {
+		rowLengthSq = ret(r,0) * ret(r,0) + ret(r,1) * ret(r,1) + ret(r,2) * ret(r,2);
+		if (rowLengthSq > FLT_EPSILON) {
+			// Use linear approximation for inverse sqrt taking advantage of the row length being close to 1.0
+			const float rowLengthInv = 1.5f - 0.5f * rowLengthSq;
+			ret(r,0) *= rowLengthInv;
+			ret(r,1) *= rowLengthInv;
+			ret(r,2) *= rowLengthInv;
```
kamilritz:
```suggestion
			ret.row(r) *= rowLengthInv;
```

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+		}
+        }
+
+	return ret;
+}
+
+bool EKFGSF_yaw::getYawData(float *yaw, float *yaw_variance)
```
kamilritz: Personally, I would split this up into three function: `float getYaw()`, `float getYawVariance()` and `bool isVelocityFusionStarted()`

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	ret(1,1) += R(1,2) * g(0) - R(1,0) * g(2);
+	ret(1,2) += R(1,0) * g(1) - R(1,1) * g(0),
+        ret(2,0) += R(2,1) * g(2) - R(2,2) * g(1);
+	ret(2,1) += R(2,2) * g(0) - R(2,0) * g(2);
+	ret(2,2) += R(2,0) * g(1) - R(2,1) * g(0);
+
+	// Renormalise rows
+	float rowLengthSq;
+	for (uint8_t r = 0; r < 3; r++) {
+		rowLengthSq = ret(r,0) * ret(r,0) + ret(r,1) * ret(r,1) + ret(r,2) * ret(r,2);
```
kamilritz:
```suggestion
		rowLengthSq = ret.row(r).norm_squared();
```

EKF/ekf_helper.cpp
priseborough: What if there is a consumer before the prediction operation n the next frame? I thought it safer to keep them consistent after a major reset and this is an infrequent operation.

EKF/EKFGSF_yaw.cpp
priseborough: The description is above the declaration

	// Efficient propagation of a delta angle in body frame applied to the body to earth frame rotation matrix

EKF/EKFGSF_yaw.h
```diff
@@ -15,10 +15,21 @@ using matrix::wrap_pi;
...
+#ifndef M_PI_2_F
+#define M_PI_2_F 1.57079632f
+#endif
+
+#ifndef M_TWOPI_INV
+#define M_TWOPI_INV 0.159154943f
+#endif
+
+#ifndef CONSTANTS_ONE_G
+#define CONSTANTS_ONE_G 9.80665f
```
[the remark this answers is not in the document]
priseborough: It is, but was not getting picked up on some builds.

EKF/mag_fusion.cpp
```diff
@@ -624,46 +624,20 @@ void Ekf::fuseHeading()
...
-			R_to_earth(0,2) = cy*sp+sy*cp*sr;
-			R_to_earth(1,0) = sy*cp+cy*sp*sr;
-			R_to_earth(1,1) = cy*cr;
-			R_to_earth(1,2) = sy*sp-cy*cp*sr;
-			R_to_earth(2,0) = -sp*cr;
-			R_to_earth(2,1) = sr;
-			R_to_earth(2,2) = cp*cr;
+			// with yaw angle set to to zero
 
+			Dcmf R_to_earth = taitBryan312ToRotMat(rotVec312);
```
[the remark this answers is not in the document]
priseborough: Done.

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+{
+	float P01 = 0.5f * (_ekf_gsf[model_index].P[0][1] + _ekf_gsf[model_index].P[1][0]);
+	float P02 = 0.5f * (_ekf_gsf[model_index].P[0][2] + _ekf_gsf[model_index].P[2][0]);
+	float P12 = 0.5f * (_ekf_gsf[model_index].P[1][2] + _ekf_gsf[model_index].P[2][1]);
+	_ekf_gsf[model_index].P[0][1] = _ekf_gsf[model_index].P[1][0] = P01;
+	_ekf_gsf[model_index].P[0][2] = _ekf_gsf[model_index].P[2][0] = P02;
+	_ekf_gsf[model_index].P[1][2] = _ekf_gsf[model_index].P[2][1] = P12;
+}
+
+Dcmf EKFGSF_yaw::taitBryan312ToRotMat(Vector3f &rot312)
```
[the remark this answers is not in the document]
priseborough: Done

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+{
+		// Calculate the frame2 to frame 1 rotation matrix from a 312 rotation sequence
+		const float c2 = cosf(rot312(2));
```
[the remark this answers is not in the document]
priseborough: Done

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1790,141 @@ void Ekf::stopFlowFusion()
...
 	memset(&_optflow_test_ratio,0.0f,sizeof(_optflow_test_ratio));
 }
+
+// Reset main nav filter yaw to value from EKF-GSF and reset velocity and position to GPS
+bool Ekf::resetYawToEKFGSF()
+{
+	// don't allow reet using the EKF-GSF estimate until the filter has started fusing velocity
+	// data and the yaw estimate has converged
+	float new_yaw, new_yaw_variance;
+	if (yawEstimator.getYawData(&new_yaw, &new_yaw_variance) && new_yaw_variance < sq(_params.EKFGSF_yaw_err_max)) {
```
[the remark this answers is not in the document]
priseborough: My understanding of the  C language was that it is guaranteed. @dagar do you know if this is the case?

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1790,141 @@ void Ekf::stopFlowFusion()
...
+		// save a copy of the quaternion state for later use in calculating the amount of reset change
+		Quatf quat_before_reset = _state.quat_nominal;
```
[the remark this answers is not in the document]
priseborough: Fixed. Also reworked code so that we calculate Euler angles from rotation matrix, not quaternion.

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1790,141 @@ void Ekf::stopFlowFusion()
...
+			rot312(0) = new_yaw;
+			rot312(1) = asinf(_R_to_earth(2, 1));
+			rot312(2) = atan2f(-_R_to_earth(2, 0), _R_to_earth(2, 2));
+
+			_R_to_earth = taitBryan312ToRotMat(rot312);
+
+		}
+
+		// calculate the amount that the quaternion has changed by
+		Quatf quat_after_reset = _R_to_earth;
```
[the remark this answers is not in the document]
priseborough: Done

EKF/EKFGSF_yaw.h
```diff
@@ -0,0 +1,155 @@
...
+	Vector3f _delta_vel;	// IMU delta velocity (m/s)
+	float _delta_ang_dt;	// _delta_ang integration time interval (sec)
+	float _delta_vel_dt;	// _delta_vel integration time interval (sec)
+	float _true_airspeed;	// true airspeed used for centripetal accel compensation (m/s)
+
+	struct _ahrs_ekf_gsf_struct{
+		Dcmf R;				// matrix that rotates a vector from body to earth frame
+		Vector3f gyro_bias;		// gyro bias learned and used by the quaternion calculation
+		bool aligned{false};		// true when AHRS has been aligned
+		float accel_FR[2] {};		// front-right acceleration vector in a horizontal plane (m/s/s)
```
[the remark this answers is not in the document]
priseborough: Not any more. It has been removed.

EKF/ekf_helper.cpp
```diff
@@ -571,13 +571,15 @@ bool Ekf::resetMagHeading(const Vector3f &mag_init, bool increase_yaw_var, bool
...
 			const Dcmf R_to_earth_ev(_ev_sample_delayed.quat);	// transformation matrix from body to world frame
+
 			// calculate the yaw angle for a 312 sequence
 			euler321(2) = atan2f(R_to_earth_ev(1, 0), R_to_earth_ev(0, 0));
 
 		} else if (_params.mag_fusion_type <= MAG_FUSE_TYPE_3D) {
-			const Dcmf R_to_earth(euler321);
 			// rotate the magnetometer measurements into earth frame using a zero yaw angle
-			const Vector3f mag_earth_pred = R_to_earth * mag_init;
+			_R_to_earth = euler321;
```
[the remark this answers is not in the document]
priseborough: Done

EKF/EKFGSF_yaw.cpp
priseborough: name changed to ahrsPredictRotMat

EKF/EKFGSF_yaw.cpp
priseborough: That function contains all the information required to use the filter output and provides a minimal code usage in the calling program with the current use case. We can easily extend the interface later if a different use case arises.

EKF/EKFGSF_yaw.cpp
priseborough: Done

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1790,141 @@ void Ekf::stopFlowFusion()
...
+		q_error.normalize();
+
+		// update quaternion states
+		_state.quat_nominal = quat_after_reset;
+		uncorrelateQuatFromOtherStates();
+
+		// record the state change
+		_state_reset_status.quat_change = q_error;
+
+		// update transformation matrix from body to world frame using the current estimate
```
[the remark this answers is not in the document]
priseborough: Fixed.

EKF/ekf_helper.cpp
```diff
@@ -648,18 +648,15 @@ bool Ekf::resetMagHeading(const Vector3f &mag_init, bool increase_yaw_var, bool
...
+		// update the rotation matrix
 		s0 = sinf(euler312(0));
 		c0 = cosf(euler312(0));
 		R_to_earth(0, 0) = c0 * c2 - s0 * s1 * s2;
-		R_to_earth(1, 1) = c0 * c1;
-		R_to_earth(2, 2) = c2 * c1;
 		R_to_earth(0, 1) = -c1 * s0;
 		R_to_earth(0, 2) = s2 * c0 + c2 * s1 * s0;
 		R_to_earth(1, 0) = c2 * s0 + s2 * s1 * c0;
+		R_to_earth(1, 1) = c0 * c1;
```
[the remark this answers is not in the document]
priseborough: We now use the taitBryan312ToRotMat function

EKF/EKFGSF_yaw.h
```diff
@@ -0,0 +1,155 @@
...
+	// align all AHRS yaw orientations to initial values
+	void ahrsAlignYaw();
+
+	// Efficient propagation of a delta angle in body frame applied to the body to earth frame rotation matrix
+	Matrix3f updateRotMat(Matrix3f &R, Vector3f &g);
+
+	// Declarations used by a bank of N_MODELS_EKFGSF EKFs
+
+	struct _ekf_gsf_struct{
+		float X[3]; // Vel North (m/s),  Vel East (m/s), yaw (rad)
```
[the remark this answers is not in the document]
priseborough: It contains disparate quantities, so is not a vector in the way that type gets used normally (eg velocity or position) and we only loop though the indexes in one place, so my first answer is no.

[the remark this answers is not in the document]
bresch: As you said, it contains disparate quantities so I think you shouldn't use Vector3f with its .xy()/.z() specific functions for that, but the state vector is a vector, so using `matrix::Vectorf<3>` seems justified to me and enables implicit matrix-vector multiplications.

EKF/EKFGSF_yaw.h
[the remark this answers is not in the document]
kamilritz: Let me see how we can build this in CI. CI should not be in the way of good programming practice. I will try in the next days.

EKF/EKFGSF_yaw.h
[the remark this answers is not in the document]
kamilritz: I would try to use `matrix::Vector` as soon as you can benefit from at least one offered matrix library operation. Therefore you don't need to implement this operation yourself, which leaves less room for new bugs or code duplication.

EKF/EKFGSF_yaw.cpp
priseborough: That makes sense. Done.

EKF/EKFGSF_yaw.cpp
priseborough: Doesnt work

EKFGSF_yaw.cpp.o -c ../../src/lib/ecl/EKF/EKFGSF_yaw.cpp
/Users/paul/src/Firmware/src/lib/ecl/EKF/EKFGSF_yaw.cpp:634:15: fatal error: no viable overloaded '*='
                        ret.row(r) *= rowLengthInv;
                        ~~~~~~~~~~ ^  ~~~~~~~~~~~~

kamilritz: Ohh, there is no "*=" operator for `matrix::Slice`. So it is indeed not possible. Sry

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	const float t5 = _ekf_gsf[model_index].S[0][1] * _ekf_gsf[model_index].S[1][0];
+	const float t3 = t2 - t5; // determinant
+	float t4; // determinant inverse
+	if (fabsf(t3) > 1e-6f) {
+		t4 = 1.0f/t3;
+	} else {
+		t4 = 1.0f/t2;
+	}
+
+	// inv(S)
```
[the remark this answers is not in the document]
priseborough: This function needs a customised protection against a badly conditioned innovation variance matrix.

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	float invMat[2][2];
+	invMat[0][0] =   t4 * _ekf_gsf[model_index].S[1][1];
+	invMat[1][1] =   t4 * _ekf_gsf[model_index].S[0][0];
+	invMat[0][1] = - t4 * _ekf_gsf[model_index].S[0][1];
+	invMat[1][0] = - t4 * _ekf_gsf[model_index].S[1][0];
+
+ 	// inv(S) * innovation
+	float tempVec[2];
```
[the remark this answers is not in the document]
priseborough: Done.

EKF/EKFGSF_yaw.h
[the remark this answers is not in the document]
priseborough: I've been through the other suggestions and made use of those types where appropriate.

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,661 @@
...
+	// force symmetry
+	_ekf_gsf[model_index].P.makeBlockSymmetric<3>(0);
+
+	float oldYaw = _ekf_gsf[model_index].X[2];
+	for (uint8_t obs_index = 0; obs_index < 2; obs_index++) {
+		// apply the state corrections including the compression scale factor
+		for (unsigned row = 0; row < 3; row++) {
+			_ekf_gsf[model_index].X[row] -= K[row][obs_index] * _ekf_gsf[model_index].innov(obs_index) * innov_comp_scale_factor;
+		}
+	}
```
kamilritz: If you make K an matrix::Matrix<float, 3, 2> or (2,3) you can replace this with a single matrix multiplication.

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,661 @@
...
+
+	// calculate inv(S)
+	matrix::SquareMatrix<float, 2> invMat;
+	invMat(0,0) =   t4 * _ekf_gsf[model_index].S[1][1];
+	invMat(1,1) =   t4 * _ekf_gsf[model_index].S[0][0];
+	invMat(0,1) = - t4 * _ekf_gsf[model_index].S[0][1];
+	invMat(1,0) = - t4 * _ekf_gsf[model_index].S[1][0];
+
+ 	// calculate inv(S) * innovation
+	matrix::Vector2f tempVec = invMat * _ekf_gsf[model_index].innov;
```
kamilritz:
```suggestion
	const matrix::Vector2f tempVec = invMat * _ekf_gsf[model_index].innov;
```

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,661 @@
...
+
+	// calculate transpose(innovation) * inv(S) * innovation
+	// * operator is overloaded to provide a dot product
+	float normDist = _ekf_gsf[model_index].innov * tempVec;
```
kamilritz:
```suggestion
	const float normDist = _ekf_gsf[model_index].innov * tempVec;
```

EKF/EKFGSF_yaw.h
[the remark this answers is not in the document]
kamilritz: After removing this gravity constant definition, the CI is still passing for me: https://github.com/kamilritz/ecl/pull/1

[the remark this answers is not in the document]
kamilritz: Please give it another try

priseborough:
@bresch I've split out the common element of the various yaw resets into a separate function. I now need to do a reversion test with a RW and FW SITL flight to ensure this refactor has not broken something.

EKF/EKFGSF_yaw.cpp
priseborough: Done

EKF/EKFGSF_yaw.cpp
priseborough: Done

EKF/EKFGSF_yaw.cpp
priseborough: Done

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+{
+	if(_ekf_gsf_vel_fuse_started) {
+		memcpy(yaw, &_gsf_yaw, sizeof(_gsf_yaw));
+		memcpy(yaw_variance, &_gsf_yaw_variance, sizeof(_gsf_yaw_variance));
+		return true;
+	}
+	return false;
+}
+
+void EKFGSF_yaw::pushVelData(Vector2f velocity, float accuracy)
```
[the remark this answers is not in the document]
priseborough: Was the other name not clear? I'll change it .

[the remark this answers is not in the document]
kamilritz: It was, but with "push" I associate more pushing data in a container/buffer. In this case it is more setting a member.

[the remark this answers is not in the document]
kamilritz: But it is also about passing the vector by const-reference.

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	const float P20 = _ekf_gsf[model_index].P[2][0];
+	const float P21 = _ekf_gsf[model_index].P[2][1];
+	const float P22 = _ekf_gsf[model_index].P[2][2];
+
+	// Use fixed values for delta velocity and delta angle process noise variances
+	float dvxVar = _accel_noise * _delta_vel_dt; // variance of forward delta velocity - (m/s)^2
+	dvxVar *= dvxVar;
+	float dvyVar = dvxVar; // variance of right delta velocity - (m/s)^2
+	float dazVar = _gyro_noise * _delta_ang_dt; // variance of yaw delta angle - rad^2
+	dazVar *= dazVar;
```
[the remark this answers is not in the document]
priseborough: Done

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	// force symmetry
+	makeCovSymEKF(model_index);
+}
+
+// Update EKF states and covariance for specified model index using velocity measurement
+void EKFGSF_yaw::updateEKF(const uint8_t model_index)
+{
+	// set observation variance from accuracy estimate supplied by GPS and apply a sanity check minimum
+	float velObsVar = fmaxf(_vel_accuracy, 0.5f);
+	velObsVar *= velObsVar;
```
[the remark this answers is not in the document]
priseborough: Done

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	// copy covariance matrix to temporary variables
+	const float P00 = _ekf_gsf[model_index].P[0][0];
+	const float P01 = _ekf_gsf[model_index].P[0][1];
+	const float P02 = _ekf_gsf[model_index].P[0][2];
+	const float P10 = _ekf_gsf[model_index].P[1][0];
+	const float P11 = _ekf_gsf[model_index].P[1][1];
+	const float P12 = _ekf_gsf[model_index].P[1][2];
+	const float P20 = _ekf_gsf[model_index].P[2][0];
+	const float P21 = _ekf_gsf[model_index].P[2][1];
+	const float P22 = _ekf_gsf[model_index].P[2][2];
```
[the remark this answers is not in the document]
priseborough: Done

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,654 @@
...
+
+	// sum delta velocities in earth frame:
+	_ekf_gsf[model_index].X(0) += del_vel_NED(0);
+	_ekf_gsf[model_index].X(1) += del_vel_NED(1);
+
+	// predict covariance - autocode from https://github.com/priseborough/3_state_filter/blob/flightLogReplay-wip/calcPupdate.txt
+
+	// Local short variable name copies required for readability
+	// Compiler might be smart enough to optimise these out
+	const float P00 = _ekf_gsf[model_index].P(0,0);
```
kamilritz: You can apply the same reference trick here

EKF/EKFGSF_yaw.cpp
```diff
@@ -0,0 +1,668 @@
...
+	float innov_comp_scale_factor = 1.0f;
+	if (fabsf(S_det_inv) > 1E-6f) {
+		// Calculate elements for innovation covariance inverse matrix assuming symmetry
+		S_det_inv = 1.0f / S_det_inv;
+		const float S_inv_NN = _ekf_gsf[model_index].S[1][1] * S_det_inv;
+		const float S_inv_EE = _ekf_gsf[model_index].S[0][0] * S_det_inv;
+		const float S_inv_NE = _ekf_gsf[model_index].S[0][1] * S_det_inv;
+
+		// The following expression was derived symbolically from test ratio = transpose(innovation) * inverse(innovation variance) * innovation = [1x2] * [2,2] * [2,1] = [1,1]
+		const float test_ratio = _ekf_gsf[model_index].innov[0]*(_ekf_gsf[model_index].innov[0]*S_inv_NN + _ekf_gsf[model_index].innov[1]*S_inv_NE) + _ekf_gsf[model_index].innov[1]*(_ekf_gsf[model_index].innov[0]*S_inv_NE + _ekf_gsf[model_index].innov[1]*S_inv_EE);
```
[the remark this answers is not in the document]
priseborough: I've added the determinant inverse and matrix inverse to the ekf_gsf struct so that we only calculate them once and have added a helper function to compute the inverse. It adds 5 floats to stack for each filter bank but is worth it in terms of eliminating the duplicated matrix inverse calculation and making the code cleaner.

EKF/EKFGSF_yaw.cpp
priseborough: Will do. Not sure how much it buys in practice given the complier optimisations.

EKF/EKFGSF_yaw.h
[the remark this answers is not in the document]
priseborough: Done

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1790,141 @@ void Ekf::stopFlowFusion()
...
+		_do_emergency_yaw_reset = true;
+	}
+}
+
+bool Ekf::getDataEKFGSF(float *yaw_composite, float *yaw_variance, float yaw[N_MODELS_EKFGSF], float innov_VN[N_MODELS_EKFGSF], float innov_VE[N_MODELS_EKFGSF], float weight[N_MODELS_EKFGSF])
+{
+	return yawEstimator.getLogData(yaw_composite,yaw_variance,yaw,innov_VN,innov_VE,weight);
+}
+
+Dcmf Ekf::taitBryan312ToRotMat(Vector3f &rot312)
```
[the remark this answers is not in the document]
priseborough: Better to add it to the math library. It should support 312 rotation sequences. I'l tackle that in a separate PR.

EKF/ekf_helper.cpp
```diff
@@ -1819,3 +1790,141 @@ void Ekf::stopFlowFusion()
...
+		ECL_INFO_TIMESTAMPED("Emergency yaw reset - magnetometer use stopped");
+
+		return true;
+	}
+
+	return false;
+
+}
+
+void Ekf::requestEmergencyNavReset(uint8_t counter)
```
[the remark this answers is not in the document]
priseborough: Original thinking was this reset command was coming across on another message that was received every frame, so a counter would enable the EKF to tell if it was a new request,

priseborough:
> @priseborough This is great. I also thought about that. May be we can even move the update of the rotation matrix `_R_to_earth` into the `resetQuatStateYaw` function. Probably `resetHeading` would then be a better name. With this we can make better sure that the quaternion and rotation_matrix are synced.

_R_to_earth is updated by this function.

kamilritz:
@priseborough it might make sense to reopen this as a new PR. With the existing feedback it is quite hard to keep the overview.

priseborough:
@kamilritz New PR opened here: https://github.com/PX4/ecl/pull/766. It has been rebased.

PR #765, opened 2020-03-02
delete attitude_fw

dagar:
Let's merge this back into PX4/Firmware. A large proportion of the code simply goes away if the work is handled in place.


kamilritz:
LGTM. @dagar if you name both PR branches (Firmware and ECL) exactly the same. The `Firmware Build Test` will checkout to the right Firmware branch and show that it compiles with the ECL PR version. All the tests are passing here: https://github.com/PX4/ecl/pull/779

PR #767, opened 2020-03-04
EKF: do not fuse multiple times the same height

bresch:
The _fuse_height flag was never set to zero, hence the fusion was called at each iteration, even if no new data is available.
The effects were: high CPU usage and virtually less measurement noise due to multiple fusion of the same sample: see https://github.com/PX4/Firmware/issues/14204

Before this PR:
> Feb 26, 2020
ekf2: update: 20247 events, 7280866us elapsed, **359.60us avg**, min 12us max 979us 401.864us rms

With this PR:
> ekf2: update: 4172 events, 777347us elapsed, **186.32us avg**, min 11us max 855us 289.353us rms

Note that I had to increase the margins in the unit test because we're not fusing height at full speed anymore.

CPU load on a Pixhawk4 (FMUv5, F7), disarmed:
Before this PR:
DeepinScreenshot_select-area_20200304095534

With this PR:
DeepinScreenshot_select-area_20200304095755


jkflying:
Is there some way we can add tests to make sure this kind of thing can't happen in the future? 
Eg. making sure the variances of all states increase if there are no new measurements

bresch:
@jkflying Yes, I was thinking about that as well. I'll add that

EKF/control.cpp
```diff
@@ -1331,7 +1328,7 @@ void Ekf::controlFakePosFusion()
 		_using_synthetic_position = true;
 
 		// Fuse synthetic position observations every 200msec
-		if (isTimedOut(_time_last_fake_pos, (uint64_t)2e5) || _fuse_height) {
+		if (isTimedOut(_time_last_fake_pos, (uint64_t)2e5)) {
```
kamilritz: My interpretation of the intent here was that we wanted to fuse the horizontal fake position the same amount as we fuse the baro data which we assume is available. Not sure if this is necessary.

bresch: That's what I also thought, but then I think we have to do it differently because with that, we're fusing every 200ms + every baro update. We can discuss about that in the next call but for now, it seems fine not to do it.

bresch: And the temptation to remove a global variable was too big....

kamilritz: Yes, I completely agree. If you feel comfortable to do so, please go ahead.

PR #769, opened 2020-03-05
CleanUp of the EKFGSF

kamilritz:
@priseborough Could you please test if this is still working on the replay logs? And I noticed that we never properly align the tilt of the [AHRS](https://github.com/PX4/ecl/blob/master/EKF/EKFGSF_yaw.cpp#L238)

@bresch The change indication file changed on my machine locally without any changes to the code.

### DIFF
--- a/EKF/EKFGSF_yaw.cpp
+++ b/EKF/EKFGSF_yaw.cpp
@@ -222,6 +222,9 @@ void EKFGSF_yaw::ahrsAlignTilt()
 	R.setRow(1, east_in_bf);
 	R.setRow(2, down_in_bf);
 
+	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index++) {
+		_ahrs_ekf_gsf[model_index].R = R;
+	}
 }
 
 void EKFGSF_yaw::ahrsAlignYaw()
--- a/EKF/EKFGSF_yaw.cpp
+++ b/EKF/EKFGSF_yaw.cpp
@@ -81,7 +81,7 @@ void EKFGSF_yaw::update(const Vector3f del_ang, // IMU delta angle rotation vect
 
 				// normalise the weighting function
 				if (_ekf_gsf_vel_fuse_started && total_w > 1e-15f && !bad_update) {
-					float total_w_inv = 1.0f / total_w;
+					const float total_w_inv = 1.0f / total_w;
 					for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
 						_ekf_gsf[model_index].W = newWeight[model_index] * total_w_inv;
 					}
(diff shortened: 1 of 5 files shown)


EKF/EKFGSF_yaw.cpp
```diff
@@ -73,18 +68,16 @@ void EKFGSF_yaw::update(const Vector3f del_ang, // IMU delta angle rotation vect
...
-				if (_ekf_gsf_vel_fuse_started && total_w > 1e-15f && !bad_update) {
-					float total_w_inv = 1.0f / total_w;
-					for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
-						_ekf_gsf[model_index].W = newWeight[model_index] * total_w_inv;
-					}
+				if (total_weight > 1e-15f) {
+					_model_weights /= total_weight;
 				}
 
 				// Enforce a minimum weighting value. This was added during initial development but has not been needed
```
kamilritz: @priseborough Is this minimum weight computation still necessary? Could we not just bound it [here](https://github.com/PX4/ecl/pull/769/files#diff-9a4cd2c2327ab18e053e90ec5eb0accaR74)

jkflying:
@kamilritz for the change indicator, make sure you are using the exact gcc version as on CI.

EKF/EKFGSF_yaw.cpp
priseborough: After normalisation the minimum value would be changed. After we get through a period if wider use/testing I'm hoping to remove the min weight parameter and that block of code.

priseborough:
Your removal of the #defines is causing a build error - they were put there for a reason. 

This error is generated after the following steps to make sure everything is being built from the correct modules:

 rm -rf *
git submodule update --init --recursive 
export replay_mode=ekf2
export replay=<path to file>
make px4_sitl none

/Users/paul/src/Firmware/src/lib/ecl/EKF/EKFGSF_yaw.cpp:479:22: fatal error: use of undeclared identifier 'M_PI_2_F'
        _gsf_yaw_variance = M_PI_2_F * M_PI_2_F;

Repeating with ecl master and it builds OK. My preference would be to use const float value for all of these and remove dependencies on those #defines. They have been an ongoing source of frustration. @dagar  FYI.

priseborough:
Replay results look OK.

kamilritz:
@priseborough I would also prefer to add them as constexpr float to common.h
+
!!! ISSUE !!!: I noticed that we never properly align the tilt of the [AHRS](https://github.com/PX4/ecl/blob/master/EKF/EKFGSF_yaw.cpp#L238). I dont think this is on purpose.

priseborough:
> @priseborough I would also prefer to add them as constexpr float to common.h
> +
> !!! ISSUE !!!: I noticed that we never properly align the tilt of the [AHRS](https://github.com/PX4/ecl/blob/master/EKF/EKFGSF_yaw.cpp#L238). I dont think this is on purpose.

Yes, please add:

	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index++) {
		_ahrs_ekf_gsf[model_index].R = R;
	}

I think it got lost when I restructured and there was enough time between boot ad flight for the AHRS to self align.

priseborough:
@mhkabir Not until another program uses them. The EKF is using the exisiting #defines. Once this goes in feel free to raise a PR changing that.

kamilritz:
@priseborough Great, I am happy with this. LGTM

PR #770, opened 2020-03-08
EKF: Enable GPS flight without magnetometer

priseborough:
Enables takeoff in a non-GPS flight mode with mag fusion type set to MAG_FUSE_TYPE_NONE. After sufficient horizontal movement the EKF will reset the yaw to the EKF-GSF estimate. After that occurs GPS fusion will commence and normal flight operations can commence.

This would enable operation where GPS quality is good, but magnetic field conditions do not allow use of a magnetometer. Examples include operation from ships or close to the earths magnetic poles.

Replay yaw data - the EKF yaw is 0 at initial alignment and is reset to the GSF estimate when the GSF yaw 1-sigma accuracy estimate drops below 15 degrees:

Screen Shot 2020-03-08 at 9 50 40 pm

GPS velocity innovations - these are low after the reset indicating that it was a good yaw alignment:

Screen Shot 2020-03-08 at 9 53 26 pm

If an optical flow sensor and range finder was fitted, the initial takeoff could be made in a position control mode and the maximum height would be limited by the position controller until the yaw has aligned and GPS fusion commenced.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -644,6 +644,22 @@ void Ekf::controlGpsFusion()
 			ECL_WARN_TIMESTAMPED("GPS data quality poor - stopping use");
 		}
 
+		// handle case where we are not currently using GPS, but need to align yaw angle using EKF-GSF before
+		// we can start using GPS
+		if (!_control_status.flags.gps && _do_ekfgsf_yaw_reset && isTimedOut(_ekfgsf_yaw_reset_time, 5000000)) {
+			if (resetYawToEKFGSF()) {
+				_ekfgsf_yaw_reset_time = _time_last_imu;
+				_do_ekfgsf_yaw_reset = false;
+
+				// Reset the timeout counters
+				_time_last_hor_pos_fuse = _time_last_imu;
+				_time_last_delpos_fuse = _time_last_imu;
+				_time_last_hor_vel_fuse = _time_last_imu;
+				_time_last_of_fuse = _time_last_imu;
+
+			}
+		}
+
 		// handle the case when we now have GPS, but have not been using it for an extended period
 		if (_control_status.flags.gps) {
 			// We are relying on aiding to constrain drift so after a specified time
@@ -667,14 +683,14 @@ void Ekf::controlGpsFusion()
 				_time_last_on_ground_us = _time_last_imu;
 			}
 			const bool recent_takeoff = _control_status.flags.in_air && !isTimedOut(_time_last_on_ground_us, 30000000);
-			const bool reset_yaw_to_EKFGSF = (do_reset || _do_emergency_yaw_reset || stopped_following_gps_velocity) &&
+			const bool reset_yaw_to_EKFGSF = (do_reset || _do_ekfgsf_yaw_reset || stopped_following_gps_velocity) &&
 								recent_takeoff &&
-								isTimedOut(_emergency_yaw_reset_time, 5000000);
+								isTimedOut(_ekfgsf_yaw_reset_time, 5000000);
 
 			if (reset_yaw_to_EKFGSF) {
 				if (resetYawToEKFGSF()) {
-					_emergency_yaw_reset_time = _time_last_imu;
-					_do_emergency_yaw_reset = false;
+					_ekfgsf_yaw_reset_time = _time_last_imu;
+					_do_ekfgsf_yaw_reset = false;
 
 					// Reset the timeout counters
 					_time_last_hor_pos_fuse = _time_last_imu;
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -664,13 +664,13 @@ void Ekf::controlGpsFusion()
 		if (_control_status.flags.gps) {
 			// We are relying on aiding to constrain drift so after a specified time
 			// with no aiding we need to do something
-			bool do_reset = isTimedOut(_time_last_hor_pos_fuse, _params.reset_timeout_max)
+			bool do_vel_pos_reset = isTimedOut(_time_last_hor_pos_fuse, _params.reset_timeout_max)
 					&& isTimedOut(_time_last_delpos_fuse, _params.reset_timeout_max)
 					&& isTimedOut(_time_last_hor_vel_fuse, _params.reset_timeout_max)
 					&& isTimedOut(_time_last_of_fuse, _params.reset_timeout_max);
 
 			// We haven't had an absolute position fix for a longer time so need to do something
-			do_reset = do_reset || isTimedOut(_time_last_hor_pos_fuse, 2 * _params.reset_timeout_max);
+			do_vel_pos_reset = do_vel_pos_reset || isTimedOut(_time_last_hor_pos_fuse, 2 * _params.reset_timeout_max);
 
 			// A reset to the EKF-GSF estimate can be performed after a recent takeoff which will enable
 			// recovery from a bad magnetometer or field estimate.
@@ -683,23 +683,21 @@ void Ekf::controlGpsFusion()
 				_time_last_on_ground_us = _time_last_imu;
 			}
 			const bool recent_takeoff = _control_status.flags.in_air && !isTimedOut(_time_last_on_ground_us, 30000000);
-			const bool reset_yaw_to_EKFGSF = (do_reset || _do_ekfgsf_yaw_reset || stopped_following_gps_velocity) &&
+			const bool do_yaw_vel_pos_reset = (do_vel_pos_reset || _do_ekfgsf_yaw_reset || stopped_following_gps_velocity) &&
 								recent_takeoff &&
 								isTimedOut(_ekfgsf_yaw_reset_time, 5000000);
 
-			if (reset_yaw_to_EKFGSF) {
-				if (resetYawToEKFGSF()) {
-					_ekfgsf_yaw_reset_time = _time_last_imu;
-					_do_ekfgsf_yaw_reset = false;
+			if (do_yaw_vel_pos_reset && resetYawToEKFGSF()) {
+				_ekfgsf_yaw_reset_time = _time_last_imu;
+				_do_ekfgsf_yaw_reset = false;
 
-					// Reset the timeout counters
-					_time_last_hor_pos_fuse = _time_last_imu;
-					_time_last_delpos_fuse = _time_last_imu;
-					_time_last_hor_vel_fuse = _time_last_imu;
-					_time_last_of_fuse = _time_last_imu;
+				// Reset the timeout counters
+				_time_last_hor_pos_fuse = _time_last_imu;
+				_time_last_delpos_fuse = _time_last_imu;
+				_time_last_hor_vel_fuse = _time_last_imu;
+				_time_last_of_fuse = _time_last_imu;
 
-				}
-			} else if (do_reset) {
+			} else if (do_vel_pos_reset) {
 				// use GPS velocity data to check and correct yaw angle if a FW vehicle
 				if (_control_status.flags.fixed_wing && _control_status.flags.in_air) {
 					// if flying a fixed wing aircraft, do a complete reset that includes yaw
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -646,17 +646,11 @@ void Ekf::controlGpsFusion()
 
 		// handle case where we are not currently using GPS, but need to align yaw angle using EKF-GSF before
 		// we can start using GPS
-		if (!_control_status.flags.gps && _do_ekfgsf_yaw_reset && isTimedOut(_ekfgsf_yaw_reset_time, 5000000)) {
+		bool align_yaw_using_gsf = !_control_status.flags.gps && _do_ekfgsf_yaw_reset && isTimedOut(_ekfgsf_yaw_reset_time, 5000000);
+		if (align_yaw_using_gsf) {
 			if (resetYawToEKFGSF()) {
 				_ekfgsf_yaw_reset_time = _time_last_imu;
 				_do_ekfgsf_yaw_reset = false;
-
-				// Reset the timeout counters
-				_time_last_hor_pos_fuse = _time_last_imu;
-				_time_last_delpos_fuse = _time_last_imu;
-				_time_last_hor_vel_fuse = _time_last_imu;
-				_time_last_of_fuse = _time_last_imu;
-
 			}
 		}
 
@@ -687,16 +681,18 @@ void Ekf::controlGpsFusion()
 								recent_takeoff &&
 								isTimedOut(_ekfgsf_yaw_reset_time, 5000000);
 
-			if (do_yaw_vel_pos_reset && resetYawToEKFGSF()) {
-				_ekfgsf_yaw_reset_time = _time_last_imu;
-				_do_ekfgsf_yaw_reset = false;
+			if (do_yaw_vel_pos_reset) {
+				if (resetYawToEKFGSF()) {
+					_ekfgsf_yaw_reset_time = _time_last_imu;
+					_do_ekfgsf_yaw_reset = false;
 
-				// Reset the timeout counters
-				_time_last_hor_pos_fuse = _time_last_imu;
-				_time_last_delpos_fuse = _time_last_imu;
-				_time_last_hor_vel_fuse = _time_last_imu;
-				_time_last_of_fuse = _time_last_imu;
+					// Reset the timeout counters
+					_time_last_hor_pos_fuse = _time_last_imu;
+					_time_last_delpos_fuse = _time_last_imu;
+					_time_last_hor_vel_fuse = _time_last_imu;
+					_time_last_of_fuse = _time_last_imu;
 
+				}
 			} else if (do_vel_pos_reset) {
 				// use GPS velocity data to check and correct yaw angle if a FW vehicle
 				if (_control_status.flags.fixed_wing && _control_status.flags.in_air) {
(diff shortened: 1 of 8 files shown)


priseborough:
The console messages still need to be tidied up. At the moment the Ekf::resetYawToEKFGSF() function also does a no-GPS reset of the velocity and position states, eg:

INFO  [ecl/EKF] 39647640: reset velocity to zero
INFO  [ecl/EKF] 39647640: reset position to last known position

This is followed 200msec later by a reset to the GPS velocity and position when GPS aiding commences, eg:

INFO  [ecl/EKF] 39849228: reset position to GPS
INFO  [ecl/EKF] 39849228: reset velocity to GPS

LorenzMeier:
Can we just disable them and turn them into PX4_DEBUG statements?

priseborough:
Testing has highlighted that unconstrained quaternion variance growth preflight is an issue that needs to be addressed, eg:

Screen Shot 2020-03-09 at 9 52 14 am

Otherwise the yaw reset worked well. I was able to takeoff in ALTCTRL mode and switch to POSITION mode a few seconds after takeoff. If the vehicle was left to loiter, yaw error would build up until a mild 'toilet bowl' developed. This would enable the EKF to correct the yaw and the motion would subside. This process repeated approximately 2 minutes later:

Screen Shot 2020-03-09 at 9 55 32 am

priseborough:
Replay indicates that the yaw drift and variance growth on ground issue has been fixed by https://github.com/PX4/ecl/pull/770/commits/4ff75e5e7d03f6f92540d1c13205a9033c267ef3

priseborough:
Quaternion variances from replay:

Screen Shot 2020-03-09 at 11 25 31 am

GPS velocity innovations from replay:
Screen Shot 2020-03-09 at 11 27 31 am

priseborough:
After fixing the bug causing takeoff with badly conditioned quaternion state covariances, the main EKF and GSF now agree closely in yaw throughout flight. 

Screen Shot 2020-03-09 at 11 30 50 am

This and the lower GPS velocity innovations indicate that the post takeoff 'toilet bowling' that resulted from the buildup of yaw error should be significantly reduced. A flight test will be performed to verify this.

Note: The build up of yaw error, the resulting 'toilet bowling' and its recovery is an interaction between the estimator, control loops and vehicle dynamics which means improvements demonstrated using ekf2 replay are a necessary, but not sufficient condition.

priseborough:
Flight test of https://github.com/PX4/ecl/pull/770/commits/4ff75e5e7d03f6f92540d1c13205a9033c267ef3 showed no evidence of yaw drift, toilet bowling or other anomalous behaviour during 4 minutes of hovering without magnetomer use.

2020-03-09_14-53-21_03_plot

2020-03-09_14-53-21_01_plot

2020-03-09_14-53-21_02_plot

priseborough:
> Can we just disable them and turn them into PX4_DEBUG statements?

I think we can.

priseborough:
@LorenzMeier I don't think the messages other than the initial alignment belong in the GCS console, but changing them to PX4_DEBUG messages has removed them from the flash log and log analysis will become more time consuming without them. Is there a way that they can be logged, but not sent to the GCS?

EKF/ekf_helper.cpp
```diff
@@ -1775,8 +1775,14 @@ bool Ekf::resetYawToEKFGSF()
 
 		// record a magnetic field alignment event to prevent possibility of the EKF trying to reset the yaw to the mag later in flight
 		_flt_mag_align_start_time = _imu_sample_delayed.time_us;
+		_control_status.flags.mag_aligned_in_flight = true;
```
bresch: This isn't true.
The definition of `mag_aligned_in_flight` is "true when the in-flight mag field alignment has been completed".
I understand that this variable needs to be set to true to have it working, but we then need to rename that variable and make sure that is doesn't have other side effects. For example [getMagDeclination](https://github.com/PX4/ecl/blob/d09473da89ca7f56fcbfe5d3eea4c3e1ca84fa25/EKF/ekf_helper.cpp#L601-L621) assumes that `mag_aligned_in_flight` means that the earth mag field states of the EKF can now be used to compute the declination.
Misusing that variable will create corner cases that will be extremely difficult to debug

EKF/gps_checks.cpp
```diff
@@ -85,7 +85,13 @@ bool Ekf::collect_gps(const gps_message &gps)
 		_mag_strength_gps = 0.01f * get_mag_strength(lat, lon);
 
 		// request a reset of the yaw using the new declination
-		_mag_yaw_reset_req = true;
+		if (_params.mag_fusion_type == MAG_FUSE_TYPE_NONE) {
```
bresch: I would really like to avoid having the yaw estimator reset silently hacked in the code through those `if (_params.mag_fusion_type == MAG_FUSE_TYPE_NONE) {` a bit everywhere.

In the context of this function, what's required is a yaw reset. Not specifically from a magnetometer or a yaw estimator implemented using an EKF GSF or a dual antenna GNSS receiver (btw, this is not handled yet), just a yaw reset. I think we could we simply set a generic variable `_yaw_reset_req = true` and let an external logic do the conversion to a specific reset depending on what's available and selected (we should also to a yaw reset using the solution of the yaw estimator if not `MAG_FUSE_TYPE_NONE` but that the mag data is not usable).

EKF/mag_control.cpp
```diff
@@ -148,7 +148,7 @@ bool Ekf::isYawResetAuthorized() const
 
 bool Ekf::canResetMagHeading() const
 {
-	return !isStrongMagneticDisturbance();
+	return !isStrongMagneticDisturbance() && (_params.mag_fusion_type != MAG_FUSE_TYPE_NONE);
```
bresch: Thanks for that!

hamishwillee:
Anything need to be added in https://docs.px4.io/master/en/advanced_config/tuning_the_ecl_ekf.html ?

@priseborough , perhaps an update at least to https://docs.px4.io/master/en/advanced_config/tuning_the_ecl_ekf.html#magnetometer ?

EKF/mag_control.cpp
```diff
@@ -271,15 +271,17 @@ bool Ekf::shouldInhibitMag() const
...
 	const bool user_selected = (_params.mag_fusion_type == MAG_FUSE_TYPE_INDOOR);
 
 	const bool heading_not_required_for_navigation = !_control_status.flags.gps
 							 && !_control_status.flags.ev_pos
 							 && !_control_status.flags.ev_vel;
 
 	return (user_selected && heading_not_required_for_navigation)
-	       || isStrongMagneticDisturbance();
+	       || isStrongMagneticDisturbance()
+	       || (_params.mag_fusion_type == MAG_FUSE_TYPE_NONE);
```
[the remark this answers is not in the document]
priseborough: Done.

priseborough:
@bresch I restructured the yaw fusion to make it more modular and the flight log replays, but I have noticed that the yaw estimate for the main EKF is no longer as stable, so will investigate further.

priseborough:
The change in behaviour is due to the removal of continual fusion of a zero innovation heading observation during up and away flight. This had been replaced by code that checked the magnitude of the quaternion variances and only fused in the zero innovation if the sum of variances was > 0.01.

Removal of this condition fixes the issue, however doing this will also slow down how quickly the EKF is able to correct yaw error buildup using horizontal velocity changes. Further replay testing should determine what the appropriate max variance threshold is.

EKF/mag_fusion.cpp
```diff
@@ -858,6 +766,155 @@ void Ekf::fuseHeading()
...
+				measured_hdg = _last_static_yaw;
+
+			}
+		} else {
+			_last_static_yaw = predicted_hdg;
+
+		}
+
+		fuseYaw312(measured_hdg, R_YAW, false);
+
```
bresch:
```suggestion

		// handle special case where yaw measurement is unavailable
		bool fuse_zero_innov = false;
		if (_yaw_use_inhibit) {
			// The yaw measurement cannot be trusted but we need to fuse something to prevent a badly
			// conditioned covariance matrix developing over time.
			if (!_control_status.flags.vehicle_at_rest) {
				// Vehicle is not at rest so fuse a zero innovation if necessary to prevent
				// unconstrained quaternion variance growth and record the predicted heading
				// to use as an observation when movement ceases.
				const float sumQuatVar = P(0,0) + P(1,1) + P(2,2) + P(3,3);
				if (sumQuatVar > 0.01f) {
					measured_hdg = 0.f
					fuse_zero_innov = true;
					R_YAW = 0.25f;
				}
				_last_static_yaw = predicted_hdg;

			} else {
				// Vehicle is at rest so use the last moving prediction as an observation
				// to prevent the heading from drifting and to enable yaw gyro bias learning
				// before takeoff.
				measured_hdg = _last_static_yaw;

			}
		} else {
			_last_static_yaw = predicted_hdg;

		}

		fuseYaw312(measured_hdg, R_YAW, fuse_zero_innov);

```

EKF/mag_fusion.cpp
```diff
@@ -435,304 +435,212 @@ void Ekf::fuseMag()
...
-	_mag_use_inhibit_prev = _mag_use_inhibit;
-
-	// wrap the innovation to the interval between +-pi
-	_heading_innov = wrap_pi(_heading_innov);
+	// Update the quaternion states and covariance matrix
+	updateQuaternion(_heading_innov, R_YAW, innov_gate, H_YAW);
+}
 
+// update quaternion states and covariances using the yaw innovation, yaw observation variance and yaw Jacobian
+void Ekf::updateQuaternion(const float innovation, const float variance, const float gate_sigma, const float *yaw_jacobian)
```
bresch:
```suggestion
void Ekf::updateQuaternion(const float innovation, const float variance, const float gate_sigma, const float (&yaw_jacobian)[4])
```
Passing by reference is safer.

EKF/control.cpp
```diff
@@ -644,6 +644,22 @@ void Ekf::controlGpsFusion()
...
+				_do_ekfgsf_yaw_reset = false;
+
+				// Reset the timeout counters
+				_time_last_hor_pos_fuse = _time_last_imu;
+				_time_last_delpos_fuse = _time_last_imu;
+				_time_last_hor_vel_fuse = _time_last_imu;
+				_time_last_of_fuse = _time_last_imu;
+
+			}
+		}
```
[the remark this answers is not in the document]
priseborough: The reset to GSF yaw and GPS within 30 seconds of takeoff is a special case. It should not be || with do_reset. Will fix.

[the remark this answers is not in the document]
priseborough: I don't understand. The resetYawToEKFGSF function does both yaw, vel and pos. I will update variable names to make the intended function clearer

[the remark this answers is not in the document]
priseborough: See https://github.com/PX4/ecl/pull/770/commits/99ba768aac1e1adff56b91ee9d3187313095d608

priseborough:
I tested this today with two scenarios:

The first was a takeoff in POSCTL with EKF2_MAG_TYPE = 0 (auto), EKF2_MAG_DECL = -90 and EKF2_DECL_TYPE = 0 to initiate flyaway. Yaw reset behaviour was as before.

https://logs.px4.io/plot_app?log=6ea2faf1-93d6-458b-826a-ea3727a24838

The second flight was a takeoff in ALTITUDE mode with EKF2_MAG_TYPE = 5 (no mag) with POSCTL selected after a few seconds of flight to give time for the yaw to align.

https://logs.px4.io/plot_app?log=e2c8ecf6-f1e1-4fc0-8b19-3be58a0ab197

priseborough:
Rebased for updated testing

EKF/control.cpp
```diff
@@ -667,23 +683,21 @@ void Ekf::controlGpsFusion()
...
+			const bool do_yaw_vel_pos_reset = (do_vel_pos_reset || _do_ekfgsf_yaw_reset || stopped_following_gps_velocity) &&
 								recent_takeoff &&
-								isTimedOut(_emergency_yaw_reset_time, 5000000);
+								isTimedOut(_ekfgsf_yaw_reset_time, 5000000);
 
-			if (reset_yaw_to_EKFGSF) {
-				if (resetYawToEKFGSF()) {
-					_emergency_yaw_reset_time = _time_last_imu;
-					_do_emergency_yaw_reset = false;
+			if (do_yaw_vel_pos_reset && resetYawToEKFGSF()) {
```
bresch: Wait, this is not correct, it means that we constantly reset to the yaw estimator when using GPS

EKF/control.cpp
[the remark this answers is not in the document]
bresch: What I meant is that we have twice the same lines of code in the same function so we could set a flag if we want to do a yaw reset to the yaw estimate and execute the logic at the end of the function only.
This also prevents doing double resets if there is a race condition.

We could also call `requestEmergencyNavReset()` and create a `processEmergencyNavReset()` function that looks for `_do_ekfgsf_yaw_reset` and in a separate function.
For example:
```c++
void processEmergencyNavReset()
{
   if (_do_ekfgsf_yaw_reset) {
       if (resetYawToEKFGSF()) { 
 		_emergency_yaw_reset_time = _time_last_imu; 
 		_do_emergency_yaw_reset = false; 
  
 		// Reset the timeout counters 
 		_time_last_hor_pos_fuse = _time_last_imu; 
 		_time_last_delpos_fuse = _time_last_imu; 
 		_time_last_hor_vel_fuse = _time_last_imu; 
 		_time_last_of_fuse = _time_last_imu; 
 	} 
   }
}

**Comment by priseborough** [comment]
Rebased to continue testing.

File: EKF/control.cpp (hunk 7 above)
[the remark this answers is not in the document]
Comment (priseborough) [reply EKF/control.cpp]: The difference is the first reset is done when we aren't using GPS and is needed to align yaw so GPs fusion can start.  The second is when we are already using GPS.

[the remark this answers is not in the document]
Comment (priseborough) [reply EKF/control.cpp]: The first reset actually doesn't require the timeout counters to be reset, so will remove those.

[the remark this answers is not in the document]
Comment (priseborough) [reply EKF/control.cpp]: See https://github.com/PX4/ecl/pull/770/commits/87531f4eb33c07475b6952833a4469e8260975fa

**Comment by priseborough** [comment]
Rebased and flight tested with:

1) A takeoff with EKF2_MAG_TYPE=6

https://logs.px4.io/plot_app?log=79dfcec5-41b4-4bd2-8bcc-9d60fd3c2ff7

2) A takeoff with the magnetic declination set to -90 deg via EKF2_MAG_DECL = -90 and EKF2_DECL_TYPE = 0

https://logs.px4.io/plot_app?log=7ab45a14-eb80-4e94-ab8f-610557734fbb

**Comment by priseborough** [comment]
> Anything need to be added in https://docs.px4.io/master/en/advanced_config/tuning_the_ecl_ekf.html ?
> 
> @priseborough , perhaps an update at least to https://docs.px4.io/master/en/advanced_config/tuning_the_ecl_ekf.html#magnetometer ?

I think the logical place for me to add some info is in https://docs.px4.io/master/en/advanced_config/tuning_the_ecl_ekf.html#magnetometer 

I'll give it a go.

**Comment by priseborough** [comment]
@hamishwillee see https://github.com/PX4/px4_user_guide/tree/pr-ecl-ekf-update

**Comment by hamishwillee** [comment]
Thanks @priseborough Best to create a PR so that you get associated comments. Note, that you updated a korean tree generated from Firmware source. So can you please check that this PR is correct: https://github.com/PX4/Firmware/pull/14434 (created fromt hat change)

File: EKF/control.cpp (hunk 8 above)
Comment (priseborough) [reply EKF/control.cpp]: I think this is OK in current code This reset can be triggered two ways, one is via timeouts and the other is via _do_emergency_yaw_reset. Both timeouts and _do_emergency_yaw_reset are cleared and also the next reset is blocked by the isTimedOut(_emergency_yaw_reset_time, 5000000) condition that must be met.

**Comment by priseborough** [comment]
@bresch Does the current code address the issues you raised?

File: EKF/ekf_helper.cpp [hunk 9]
```diff
@@ -1743,8 +1745,13 @@ bool Ekf::resetYawToEKFGSF()
```
bresch: We should only set this flag `_control_status.flags.mag_fault = true;` if the reason for the reset is due to an "Emergency yaw reset", otherwise we always get the `mag_fault` flag when flying without magnetometer (this produces the PX4 error `WARN  [commander] Stopping compass use! Check calibration on landing`).

A quick fix is to place `_control_status.flags.mag_fault = true;` a few lines below, with the `ECL_INFO_TIMESTAMPED("Emergency yaw reset - magnetometer use stopped");`. 
IMO, this function `resetYawToEKFGSF` does too many things (i.e.: reset yaw + reset velocity + reset position + set mag fault flag + gives an info message depending on the type of mag fusion) but we can still clean it later, I'm fine with the minimal fix.

EKF/ekf_helper.cpp
```diff
@@ -1755,7 +1762,7 @@ bool Ekf::resetYawToEKFGSF()
 
 void Ekf::requestEmergencyNavReset()
```
bresch: Is the purpose of this function to give the possibility to the client to trigger a yaw reset using the solution of he yaw estimator? It's never used at the moment.

bresch:
(Also rebase on master please)

EKF/ekf_helper.cpp
priseborough: Will fix

EKF/ekf_helper.cpp
priseborough: Yes, the original experimentation tried to get the commander failsafe to trigger it prior to entering a non-position mode, but I postponed that at the time as it was getting too complicated to try and get both changes in. I would still like to investigate the possibility of doing so. I don't mind removing that function if flash space is an issue.

PR #772, opened 2020-03-09
Small height reset cleanup

bresch:
Summary:
- replace all the `isRecent(gps_newest.time_us, 2 * GPS_MAX_INTERVAL)` by `!_gps_hgt_intermittent`, because it is already computed here: https://github.com/PX4/ecl/blob/93be6ea49a22b08a4bc42010a7c537640eae09fc/EKF/control.cpp#L91-L92
replace all the `isRecent(baro_init.time_us, 2 * BARO_MAX_INTERVAL)` by `!_baro_hgt_faulty`, because it is already computed here: https://github.com/PX4/ecl/blob/93be6ea49a22b08a4bc42010a7c537640eae09fc/EKF/control.cpp#L88-L89
- remove the useless setControl...Height() when the fusion mode is alredy active: 
e.g.:
```c++
if(_control_status.flags.baro_hgt) {
   ...
   setControlBaroHeight();
   ...
}
```
- extract the check of vertical velocity divergence due to accelerometer clipping to `checkVerticalAccelerationHealth()`

**Note** : it should be easy to review commit-by-commit

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -777,31 +777,7 @@ void Ekf::controlHeightSensorTimeouts()
 	 * source failed if we have switched.
 	*/
 
-	// Check for IMU accelerometer vibration induced clipping as evidenced by the vertical innovations being positive and not stale.
-	// Clipping causes the average accel reading to move towards zero which makes the INS think it is falling and produces positive vertical innovations
-	float var_product_lim = sq(_params.vert_innov_test_lim) * sq(_params.vert_innov_test_lim);
-	bool bad_vert_accel = (_control_status.flags.baro_hgt && // we can only run this check if vertical position and velocity observations are independent
-			(sq(_gps_pos_innov(2) * fmaxf(fabsf(_gps_vel_innov(2)),fabsf(_ev_vel_innov(2)))) > var_product_lim * (_gps_pos_innov_var(2) * fmaxf(fabsf(_gps_vel_innov_var(2)),fabsf(_ev_vel_innov_var(2))))) && // vertical position and velocity sensors are in agreement that we have a significant error
-			(_gps_vel_innov(2) > 0.0f || _ev_vel_innov(2) > 0.0f) && // positive innovation indicates that the inertial nav thinks it is falling
-			((_imu_sample_delayed.time_us - _baro_sample_delayed.time_us) < 2 * BARO_MAX_INTERVAL) && // vertical position data is fresh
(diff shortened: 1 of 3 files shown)


jkflying:
I guess this wasn't meant to trigger the change indicator?

bresch:
> I guess this wasn't meant to trigger the change indicator?

Nope, I would have to check if this is just a timing change

bresch:
@jkflying It's actually not my PR that makes it fail
DeepinScreenshot_select-area_20200309183119

EKF/control.cpp
```diff
@@ -789,21 +789,17 @@ void Ekf::controlHeightSensorTimeouts()
...
-			const bool baro_data_available = isRecent(baro_init.time_us, 2 * BARO_MAX_INTERVAL);
-
-			// check for inertial sensing errors in the last 10 seconds
+			// check for inertial sensing errors in the last BADACC_PROBATION seconds
 			const bool prev_bad_vert_accel = isRecent(_time_bad_vert_accel, BADACC_PROBATION);
 
 			// reset to GPS if adequate GPS data is available and the timeout cannot be blamed on IMU data
 			const bool reset_to_gps = !_gps_hgt_intermittent &&
-					    ((gps_hgt_accurate && !prev_bad_vert_accel) || !baro_data_available);
+					    ((gps_hgt_accurate && !prev_bad_vert_accel) || _baro_hgt_faulty);
```
kamilritz: `_baro_hgt_faulty` will not only be set by the isRecent check. Are you aware of that?

EKF/ekf_helper.cpp
```diff
@@ -227,20 +227,19 @@ void Ekf::resetHeight()
 
 	// reset the vertical position
 	if (_control_status.flags.rng_hgt) {
+		float new_pos_down = _hgt_sensor_offset - _range_sample_delayed.rng * _R_rng_to_earth_2_2;
```
kamilritz: const

EKF/control.cpp
bresch: Yes, the 2nd place where it is set is 3 lines below

bresch:
> Some of the isRecent checks were performed at the delayed time horizon, while the global one is performed at the newest time.

Yes, the one in `checkVerticalAccelerationHealth` was using the delayed-time check to know if there is "fresh data". Given that `2 * (BARO|GPS)_MAX_INTERVAL = 1s`, the quality of this "freshness" metric is to me, still the same.

PR #774, opened 2020-03-09
Check change indication on PR

kamilritz:
We missed to check the change indication on open PRs.

### DIFF
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -1,7 +1,13 @@
 
 name: Build Tests
 
-on: [pull_request]
+on:
+  push:
+    branches:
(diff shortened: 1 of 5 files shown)


mhkabir:
Shouldn't we do it on both push and pull request?

kamilritz:
We could. That would be helpful for running the checks before opening a PR. The downside is that each test is listed twice on the PR. Once for the one running on the PR and once for the one that was running on push.

mhkabir:
Yeah, but I think it's acceptable, because we want it to run on master on pushes/when PRs are merged.

kamilritz:
@mhkabir  This should be exactly what we want. See newest commit.

PR #781, opened 2020-03-18
YawEst: simplify AHRS accelerometer gain computation

bresch:
I just remembered that I found a nice generic equation to compute the accel gain for the AHRS (https://github.com/PX4/ecl/pull/751#discussion_r381228066)
https://www.desmos.com/calculator/dbqbxvnwfg

DeepinScreenshot_select-area_20200220135708

Replace the multiple if-else statements by a generic equation.
- For a multicopter, the attenuation factor is 2 and symmetrical
- For a fixedwing, the attenuation factor is 1 if the acceleration is positive and that centripetal correction is available and 2 otherwise.

Note that the function `sq` needs to be `const` in order to be used in a `const` function.

### DIFF
--- a/EKF/EKFGSF_yaw.cpp
+++ b/EKF/EKFGSF_yaw.cpp
@@ -46,8 +46,11 @@ void EKFGSF_yaw::update(const imuSample& imu_sample,
 		return;
 	}
 
+	// calculate common values used by the AHRS complementary filter models
+	_ahrs_accel_norm = _ahrs_accel.norm();
+
 	// AHRS prediction cycle for each model - this always runs
-	ahrsCalcAccelGain();
+	_ahrs_accel_fusion_gain = ahrsCalcAccelGain();
 	for (uint8_t model_index = 0; model_index < N_MODELS_EKFGSF; model_index ++) {
 		predictEKF(model_index);
 	}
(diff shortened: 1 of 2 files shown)


EKF/EKFGSF_yaw.cpp
```diff
@@ -567,29 +570,25 @@ Dcmf EKFGSF_yaw::taitBryan312ToRotMat(const Vector3f &rot312)
...
+	const bool centripetal_accel_compensation_enabled = (_true_airspeed > FLT_EPSILON);
+
+	if (centripetal_accel_compensation_enabled
+	    && _ahrs_accel_norm > CONSTANTS_ONE_G) {
+		attenuation = 1.f;
 	}
+
+	const float delta_accel_g = (_ahrs_accel_norm - CONSTANTS_ONE_G) / CONSTANTS_ONE_G;
+	const float accel_ratio_abs = math::min(fabsf(delta_accel_g), 1.f / attenuation);
+	return _tilt_gain * sq(1.f - attenuation * accel_ratio_abs);
```
kamilritz:
```suggestion
	const float accel_ratio_abs = math::min(attenuation * fabsf(delta_accel_g), 1.f);
	return _tilt_gain * sq(1.f - accel_ratio_abs);
```

kamilritz: This would save one division.

bresch: Even better, thanks

bresch:
Quick STIL test showing that everything still works properly:
2020-03-18_17-42-59_01_plot

PR #782, opened 2020-03-18
Range check cleanup

bresch:
### Summary of the changes:

- There is now only one clear path that can lead to the validity being true.
Furthermore, if the `_rng_hgt_valid` is true, we can trust it and we don't need for additional checks such as tilt.

- The case where we need to provide fake measurements because the drone is on the ground and the range finder data is bad is already handled in "controlHeightFusion" so there is no need to hack the range finder checks with that.
https://github.com/PX4/ecl/blob/f38388cdabbf5c12da0a26e91ed1a7b0b258e5ce/EKF/control.cpp#L1094-L1106

- `isRangeAidSuitable` can only return true if the terrain estimate is valid
so there is no need for an additional check

#### TODO:
- [x] Add unit tests for `_rng_hgt_valid`
- [x] SITL tests with range finder

### DIFF
--- a/EKF/Sensor.hpp
+++ b/EKF/Sensor.hpp
@@ -53,15 +53,23 @@ class Sensor
 public:
 	virtual ~Sensor() {};
 
+	/*
+	 * run sanity checks on the current data
+	 * this has to be called immediately after
+	 * setting new data
+	 */
 	virtual void runChecks(){};
-	virtual bool isHealthy() const { return _is_healthy; }
-	virtual bool isNewHealthyData() const = 0;
-	virtual bool isDelayedHealthyData() const = 0;
-	virtual bool canBeusedAsFailover() const = 0;
-	virtual bool canResetOnSensor() const = 0;
 
-protected:
-	bool _is_healthy{};
+	/*
+	 * return true if the sensor is healthy
+	 */
+	virtual bool isHealthy() const = 0;
+
+	/*
+	 * return true if the delayed sample is healthy
+	 * and can be fused in the estimator
+	 */
+	virtual bool isDelayedDataHealthy() const = 0;
 };
 
 } // namespace sensor
--- a/EKF/Sensor.hpp
+++ b/EKF/Sensor.hpp
@@ -43,6 +43,11 @@
 
 #include "common.h"
 
+namespace estimator
+{
+namespace sensor
+{
+
 class Sensor
 {
 public:
@@ -58,3 +63,6 @@ public:
 protected:
 	bool _is_healthy{};
 };
+
+} // namespace sensor
+} // namespace estimator
--- a/EKF/Sensor.hpp
+++ b/EKF/Sensor.hpp
@@ -1,0 +1,60 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2020 Estimation and Control Library (ECL). All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ * 3. Neither the name ECL nor the names of its contributors may be
+ *    used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ ****************************************************************************/
+
+/**
+ * @file Sensor.hpp
+ * Abstract class for sensors
+ *
+ * @author Mathieu Bresciani <brescianimathieu@gmail.com>
+ *
+ */
+
+#pragma once
+
+#include "common.h"
+
+class Sensor
+{
+public:
+	virtual ~Sensor() {};
+
+	virtual void runChecks(){};
+	virtual bool isHealthy() const { return _is_healthy; }
+	virtual bool isNewHealthyData() const = 0;
+	virtual bool isDelayedHealthyData() const = 0;
+	virtual bool canBeusedAsFailover() const = 0;
+	virtual bool canResetOnSensor() const = 0;
+
+protected:
+	bool _is_healthy{};
+};
--- a/EKF/Sensor.hpp
+++ b/EKF/Sensor.hpp
@@ -69,7 +69,7 @@ public:
 	 * return true if the delayed sample is healthy
 	 * and can be fused in the estimator
 	 */
-	virtual bool isDelayedDataHealthy() const = 0;
+	virtual bool isDataHealthy() const = 0;
 };
 
 } // namespace sensor
--- a/EKF/SensorRangeFinder.cpp
+++ b/EKF/SensorRangeFinder.cpp
@@ -139,15 +139,5 @@ void SensorRangeFinder::updateRangeDataStuck()
 	}
 }
 
-bool SensorRangeFinder::canBeusedAsFailover() const
-{
-	return false;
-}
-
-bool SensorRangeFinder::canResetOnSensor() const
-{
-	return false;
-}
-
 } // namespace sensor
 } // namespace estimator
--- a/EKF/SensorRangeFinder.cpp
+++ b/EKF/SensorRangeFinder.cpp
@@ -73,8 +73,8 @@ void SensorRangeFinder::updateRangeDataValidity(uint64_t time_delayed_us)
 			_time_bad_rng_signal_quality = time_delayed_us;
 
 		} else if (time_delayed_us - _time_bad_rng_signal_quality > _range_signal_hysteresis_ms) {
-			const bool is_in_range = ((_range_sample_delayed.rng > _rng_valid_min_val)
-						  && (_range_sample_delayed.rng < _rng_valid_max_val));
+			const bool is_in_range = ((_range_sample_delayed.rng >= _rng_valid_min_val)
+						  && (_range_sample_delayed.rng <= _rng_valid_max_val));
 
 			if (isTiltOk() || is_in_range) {
 				updateRangeDataStuck();
--- a/EKF/SensorRangeFinder.cpp
+++ b/EKF/SensorRangeFinder.cpp
@@ -1,6 +1,6 @@
 /****************************************************************************
  *
- *   Copyright (c) 2019 Estimation and Control Library (ECL). All rights reserved.
+ *   Copyright (c) 2020 Estimation and Control Library (ECL). All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -32,33 +32,29 @@
  ****************************************************************************/
 
 /**
- * @file range_finder_checks.cpp
- * Perform checks on range finder data in order to evaluate validity.
+ * @file SensorRangeFinder.cpp
  *
+ * @author Mathieu Bresciani <brescianimathieu@gmail.com>
  *
  */
 
-#include "ekf.h"
+#include "SensorRangeFinder.hpp"
 
-void Ekf::updateRangeDataContinuity()
+void SensorRangeFinder::runChecks(const uint64_t time_delayed_us, const Dcmf &R_to_earth)
 {
-	// update range data continuous flag (1Hz ie 2000 ms)
-	/* Timing in micro seconds */
+	// calculate 2,2 element of rotation matrix from sensor frame to earth frame
+	// this is required for use of range finder and flow data
+	_R_rng_to_earth_2_2 = R_to_earth(2, 0) * _sin_tilt_rng + R_to_earth(2, 2) * _cos_tilt_rng;
 
-	/* Apply a 2.0 sec low pass filter to the time delta from the last range finder updates */
-	float alpha = 0.5f * _dt_update;
-	_dt_last_range_update_filt_us = _dt_last_range_update_filt_us * (1.0f - alpha) + alpha *
-					(_imu_sample_delayed.time_us - _range_sample_delayed.time_us);
-
-	_dt_last_range_update_filt_us = fminf(_dt_last_range_update_filt_us, 4e6f);
+	updateRangeDataValidity(time_delayed_us);
 }
 
-void Ekf::updateRangeDataValidity()
+void SensorRangeFinder::updateRangeDataValidity(uint64_t time_delayed_us)
 {
-	updateRangeDataContinuity();
+	updateRangeDataContinuity(time_delayed_us);
 
 	// check if out of date
-	if (((_imu_sample_delayed.time_us - _range_sample_delayed.time_us) > 2 * RNG_MAX_INTERVAL)
+	if (((time_delayed_us - _range_sample_delayed.time_us) > 2 * RNG_MAX_INTERVAL)
 	    || !isRangeDataContinuous()) {
 		_rng_hgt_valid = false;
 		return;
@@ -69,16 +65,16 @@ void Ekf::updateRangeDataValidity()
 		_rng_hgt_valid = false;
 
 		if (_range_sample_delayed.quality == 0) {
-			_time_bad_rng_signal_quality = _imu_sample_delayed.time_us;
+			_time_bad_rng_signal_quality = time_delayed_us;
 
-		} else if (_imu_sample_delayed.time_us - _time_bad_rng_signal_quality > (unsigned)_params.range_signal_hysteresis_ms) {
+		} else if (time_delayed_us - _time_bad_rng_signal_quality > _range_signal_hysteresis_ms) {
 			const bool is_in_range = ((_range_sample_delayed.rng > _rng_valid_min_val)
 						  && (_range_sample_delayed.rng < _rng_valid_max_val));
 
-			if (_is_rng_tilt_ok || is_in_range) {
+			if (isTiltOk() || is_in_range) {
 				updateRangeDataStuck();
 
-				if (!_control_status.flags.rng_stuck) {
+				if (!_is_stuck) {
 					_rng_hgt_valid = true;
 				}
 			}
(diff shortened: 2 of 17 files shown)


EKF/terrain_estimator.cpp
```diff
@@ -132,53 +131,47 @@ void Ekf::runTerrainEstimator()
...
+	// calculate the innovation variance - limiting it to prevent a badly conditioned fusion
+	_hagl_innov_var = fmaxf(_terrain_var + obs_variance, obs_variance);
+
+	// perform an innovation consistency check and only fuse data if it passes
+	const float gate_size = fmaxf(_params.range_innov_gate, 1.0f);
+	_hagl_test_ratio = sq(_hagl_innov) / (sq(gate_size) * _hagl_innov_var);
+
+	if (_hagl_test_ratio <= 1.0f) {
+		// calculate the Kalman gain
+		float gain = _terrain_var / _hagl_innov_var;
```
kamilritz:
```suggestion
		const float gain = _terrain_var / _hagl_innov_var;
```

bresch:
@kamilritz Yes, I also thought about creating a class.... maybe I'll do that, it's true that writing the unit tests will be much simpler than it is now.

EKF/ekf_helper.cpp
```diff
@@ -67,7 +67,7 @@ bool Ekf::resetVelocity()
 		float heightAboveGndEst = fmaxf((_terrain_vpos - _state.pos(2)), _params.rng_gnd_clearance);
```
kamilritz:
```suggestion
		const float heightAboveGndEst = fmaxf((_terrain_vpos - _state.pos(2)), _params.rng_gnd_clearance);
```

EKF/ekf_helper.cpp
```diff
@@ -67,7 +67,7 @@ bool Ekf::resetVelocity()
...
 
 		// calculate absolute distance from focal point to centre of frame assuming a flat earth
-		float range = heightAboveGndEst / _R_rng_to_earth_2_2;
+		float range = heightAboveGndEst / _sensor_rng.getRToEarth();
```
kamilritz:
```suggestion
		const float range = heightAboveGndEst / _sensor_rng.getRToEarth();
```

EKF/ekf_helper.cpp
```diff
@@ -227,7 +227,7 @@ void Ekf::resetHeight()
 
 	// reset the vertical position
 	if (_control_status.flags.rng_hgt) {
-		float new_pos_down = _hgt_sensor_offset - _range_sample_delayed.rng * _R_rng_to_earth_2_2;
+		float new_pos_down = _hgt_sensor_offset - _sensor_rng.getDelayedRng() * _sensor_rng.getRToEarth();
```
kamilritz:
```suggestion
		const float new_pos_down = _hgt_sensor_offset - _sensor_rng.getDelayedRng() * _sensor_rng.getRToEarth();
```

EKF/estimator_interface.h
```diff
@@ -455,7 +455,7 @@ class EstimatorInterface
 	magSample _mag_sample_delayed{};
 	baroSample _baro_sample_delayed{};
 	gpsSample _gps_sample_delayed{};
-	rangeSample _range_sample_delayed{};
+	SensorRangeFinder _sensor_rng{};
```
kamilritz:
```suggestion
	SensorRangeFinder _range_finder (or _range_sensor){};
```

EKF/SensorRangeFinder.hpp
```diff
@@ -0,0 +1,131 @@
...
+	~SensorRangeFinder() override = default;
+
+	void runChecks(uint64_t time_delayed_us, const Dcmf &R_to_earth);
+	bool isHealthy() const override { return _rng_hgt_valid; }
+	bool isNewHealthyData() const override { return _range_data_ready && _rng_hgt_valid; }
+	bool isDelayedHealthyData() const override { return _range_data_ready && _rng_hgt_valid; }
+	bool canBeusedAsFailover() const override;
+	bool canResetOnSensor() const override;
+
+	void setNewestSample(rangeSample sample) { _newest_sample = sample; }
```
kamilritz: `setNewestSample() and `_newest_sample` is not used at the moment. I understand that the interface can then be extended to other sensors easily. But should we also allocate memory for `_newest_sample` and allow setting it, even if it is not used?

EKF/Sensor.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * Abstract class for sensors
+ *
+ * @author Mathieu Bresciani <brescianimathieu@gmail.com>
+ *
+ */
+
+#pragma once
+
+#include "common.h"
+
```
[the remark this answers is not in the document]
bresch: I'll put everything in `namespace estimator::sensor`

EKF/estimator_interface.h
```diff
@@ -54,6 +54,7 @@
 #include <mathlib/mathlib.h>
 
 using namespace estimator;
+using namespace estimator::sensor;
```
jkflying: This is already an issue here, but it's better not to do `using` in a header file, since then anybody who includes it also gets that, without any namespace.

Maybe put these both inside the EstimatorInterface class?

EKF/control.cpp
```diff
@@ -1093,14 +1083,13 @@ void Ekf::controlHeightFusion()
 	}
 
 	if (isTimedOut(_time_last_hgt_fuse, 2 * RNG_MAX_INTERVAL) && _control_status.flags.rng_hgt
-	    && (!_range_data_ready || !_rng_hgt_valid)) {
+	    && (!(_range_data_ready && _rng_hgt_valid))) {
```
bresch: Note that I made this change because I planned to change it to a single flag that contains the result of `_range_data_ready && _rng_hgt_valid` that we use everywhere

EKF/SensorRangeFinder.cpp
```diff
@@ -69,36 +65,48 @@ void Ekf::updateRangeDataValidity()
...
+
+	_dt_last_range_update_filt_us = fminf(_dt_last_range_update_filt_us, 4e6f);
+}
+
+void SensorRangeFinder::updateRangeDataStuck()
 {
 	// Check for "stuck" range finder measurements when range was not valid for certain period
 	// This handles a failure mode observed with some lidar sensors
-	if (((_range_sample_delayed.time_us - _time_last_rng_ready) > (uint64_t)10e6) &&
-	    _control_status.flags.in_air) {
```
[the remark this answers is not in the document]
bresch: Yes, because this part of the code is only executed if all the other checks (quality and range) are passing. I would also like to have the sensor not having to perform differently if on the ground or in air, we're not fusing it the same way from the outside anyway (e.g.: terrain estimator reset to ground clearance and we create fake data in the height fusion).

EKF/estimator_interface.h
bresch: I think the estimator interface should be in the estimator namespace. If not, I'll anyway remove the `using` to avoid namespace leaking

EKF/SensorRangeFinder.cpp
```diff
@@ -32,94 +32,86 @@
...
+				}
+			}
 		}
 	}
+}
 
-	updateRangeDataStuck();
+void SensorRangeFinder::updateRangeDataContinuity(uint64_t time_delayed_us)
+{
+	// update range data continuous flag (1Hz ie 2000 ms)
```
priseborough: These comments are out of date. I would replace them with something like:

// Calculate a first order IIR low-pass filtered time of arrival between samples using a 2 second time constant.
// Apply spike protection to the filter state.
// Timing is in micro seconds.

bresch:
I added a few simple unit tests, I'll add more to cover all the checks.

bresch:
Ready for review, the most important things to review are the unit tests, thanks :D

bresch:
Now the refactoring is done, I renamed a few variables; since the scope is limited, most of the variables can have a shorter name

test/test_SensorRangeFinder.cpp
```diff
@@ -0,0 +1,253 @@
...
+		new_sample.time_us += dt;
+	}
+
+	// THEN: the data is again declared healthy
+	EXPECT_TRUE(_range_finder.isDataHealthy());
+}
+
+TEST_F(SensorRangeFinderTest, continuity)
+{
+	Dcmf attitude{Eulerf(0.f, 0.f, 0.f)};
```
kamilritz:
```suggestion
	const Dcmf attitude{Eulerf(0.f, 0.f, 0.f)};
```

test/test_SensorRangeFinder.cpp
```diff
@@ -0,0 +1,253 @@
...
+	{
+	}
+
+protected:
+	SensorRangeFinder _range_finder{};
+	const rangeSample _good_sample{1.f, (uint64_t)2e6, 100}; // {range, time_us, quality}
+	const float _min_range{0.5f};
+	const float _max_range{10.f};
+
+	void runDurationRateSensorRate(uint64_t duration_us, uint64_t dt_update_us, uint64_t dt_sensor_us);
```
kamilritz:
```suggestion
	void updateSensorAtRate(uint64_t duration_us, uint64_t dt_update_us, uint64_t dt_sensor_us);
```

EKF/ekf.h
```diff
@@ -491,17 +488,12 @@ class Ekf : public EstimatorInterface
 	float _terrain_var{1e4f};		///< variance of terrain position estimate (m**2)
 	uint64_t _time_last_hagl_fuse{0};		///< last system time that the hagl measurement failed it's checks (uSec)
 	bool _terrain_initialised{false};	///< true when the terrain estimator has been initialized
-	float _sin_tilt_rng{0.0f};		///< sine of the range finder tilt rotation about the Y body axis
-	float _cos_tilt_rng{0.0f};		///< cosine of the range finder tilt rotation about the Y body axis
-	float _R_rng_to_earth_2_2{0.0f};	///< 2,2 element of the rotation matrix from sensor frame to earth frame
-	float _dt_last_range_update_filt_us{0.0f};	///< filtered value of the delta time elapsed since the last range measurement came into the filter (uSec)
+	bool _is_rng_tilt_ok{false};		///< true when the tilt condition for the range finder is passing
```
kamilritz: `_is_rng_tilt_ok`  says: "Please remove me, I feel so useless"

EKF/SensorRangeFinder.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "SensorRangeFinder.hpp"
+
+namespace estimator
+{
+namespace sensor
+{
+
+void SensorRangeFinder::runChecks(const uint64_t current_time_us, const Dcmf &R_to_earth)
+{
+	updateSensorToEarthRotation(R_to_earth);
```
kamilritz: As an option we could call this function directly before `isTiltOk()` and after the check `isDataInRange()` has passed. Then we would avoid some computations in case previous checks failed.

EKF/SensorRangeFinder.hpp
```diff
@@ -0,0 +1,150 @@
...
+{
+public:
+	SensorRangeFinder() = default;
+	~SensorRangeFinder() override = default;
+
+	void runChecks(uint64_t current_time_us, const matrix::Dcmf &R_to_earth);
+	bool isHealthy() const override { return _is_sample_valid; }
+	bool isDataHealthy() const override { return _is_sample_ready && _is_sample_valid; }
+
+	void setSample(rangeSample sample) {
```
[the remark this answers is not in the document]
bresch: IMO this class should contain the value and not be dependent on the scope of the referenced value. For example, let's imagine that we have a class member `rangeSample _rng` and that in some function we do:
```c++
void createNewSample() {
   rangeSample new_sample{10.f, 2e6, 100};
   _rng.setSample(new_sample);
}
```
Since the original variable is destroyed at the end of the function, the reference will be broken.

EKF/control.cpp
```diff
@@ -116,20 +116,24 @@ void Ekf::controlFusionModes()
...
 
-	// calculate 2,2 element of rotation matrix from sensor frame to earth frame
-	// this is required for use of range finder and flow data
-	_R_rng_to_earth_2_2 = _R_to_earth(2, 0) * _sin_tilt_rng + _R_to_earth(2, 2) * _cos_tilt_rng;
 
+
+	{
 	// Get range data from buffer and check validity
-	_range_data_ready = _range_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_range_sample_delayed);
+	const bool is_rng_data_ready = _range_buffer.pop_first_older_than(_imu_sample_delayed.time_us, _range_sensor.getSampleAddress());
```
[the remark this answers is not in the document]
bresch: In fact, my next step would be to place the ring buffer in the `SensorRangeFinder` class directly. I'll leave it like that for now.

EKF/terrain_estimator.cpp
```diff
@@ -57,11 +57,11 @@ bool Ekf::initHagl()
...
 	} else if ((_params.terrain_fusion_mode & TerrainFusionMask::TerrainFuseRangeFinder)
-		   && _rng_hgt_valid
-		   && isRecent(latest_measurement.time_us, (uint64_t)2e5)
-		   && _R_rng_to_earth_2_2 > _params.range_cos_max_tilt) {
+		   && _range_sensor.isHealthy()
+		   && isRecent(latest_measurement.time_us, (uint64_t)2e5)) {
 		// if we have a fresh measurement, use it to initialise the terrain estimator
-		_terrain_vpos = _state.pos(2) + latest_measurement.rng * _R_rng_to_earth_2_2;
+		// TODO: the latest measurment did not go through the checks and could be invalid!
+		_terrain_vpos = _state.pos(2) + latest_measurement.rng * _range_sensor.getCosTilt();
```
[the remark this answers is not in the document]
bresch: I think we should, yes. I don't understand why we're using the latest sample now; the EKF lives in delayed time so it should only use that.
@priseborough  What's your opinion?

EKF/SensorRangeFinder.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+	void setSensorTilt(float new_tilt)
+	{
+		if (fabsf(_tilt - new_tilt) > FLT_EPSILON) {
+			_sin_tilt_rng = sinf(new_tilt);
+			_cos_tilt_rng = cosf(new_tilt);
+		}
+	}
+
+	void setCosMaxTilt(float cos_max_tilt) { _range_cos_max_tilt = cos_max_tilt; }
```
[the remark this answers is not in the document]
bresch: well, the value is really `cos(tilt)`, why should we call it just tilt?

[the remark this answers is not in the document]
bresch: And we already have a function `setTilt` that sets `_tilt` and computes `_sin_tilt_rng` and `_cos_tilt_rng`

EKF/SensorRangeFinder.hpp
```diff
@@ -0,0 +1,150 @@
...
+	static constexpr float _dt_update{0.01f}; 	///< delta time since last ekf update TODO: this should be a parameter
+	float _dt_data_lpf{};	///< filtered value of the delta time elapsed since the last range measurement came into the filter (uSec)
+
+	/*
+	 * Tilt check
+	 */
+	float _cos_tilt_rng_to_earth{};		///< 2,2 element of the rotation matrix from sensor frame to earth frame
+	float _range_cos_max_tilt{0.7071f};	///< cosine of the maximum tilt angle from the vertical that permits use of range finder and flow data
+	float _tilt{3.14f}; 			///< range finder tilt rotation about the Y body axis
+	float _sin_tilt_rng{0.0f}; 		///< sine of the range finder tilt rotation about the Y body axis
```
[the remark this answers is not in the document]
bresch: I'm ok with "tilt offset", yes

test/test_SensorRangeFinder.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+void SensorRangeFinderTest::runDurationRateSensorRate(uint64_t duration_us, uint64_t dt_update_us, uint64_t dt_sensor_us)
+{
+	const Dcmf attitude{Eulerf(0.f, 0.f, 0.f)};
+
+	rangeSample new_sample = _good_sample;
+	uint64_t t_now_us = _good_sample.time_us;
+	for (int i = 0; i < int(duration_us / dt_update_us); i++) {
+		t_now_us += dt_update_us;
+		if ((i % int(dt_sensor_us / dt_update_us)) == 0) {
```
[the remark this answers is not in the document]
bresch: Not sure what you mean by that. The cast to int makes sure that the modulo operator works properly

test/test_SensorRangeFinder.cpp
```diff
@@ -0,0 +1,253 @@
...
+TEST_F(SensorRangeFinderTest, tiltExceeded)
+{
+	// WHEN: the drone is excessively tilted
+	Dcmf attitude{Eulerf(0.f, 1.f, 0.f)};
+
+	_range_finder.setSample(_good_sample);
+	_range_finder.runChecks(_good_sample.time_us, attitude);
+
+	// THEN: the data should be marked as unhealthy
+	EXPECT_FALSE(_range_finder.isDataHealthy());
```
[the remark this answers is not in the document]
bresch: Correct

EKF/Sensor.hpp
```diff
@@ -0,0 +1,76 @@
...
+	/*
+	 * return true if the sensor is healthy
+	 */
+	virtual bool isHealthy() const = 0;
+
+	/*
+	 * return true if the delayed sample is healthy
+	 * and can be fused in the estimator
+	 */
+	virtual bool isDataHealthy() const = 0;
```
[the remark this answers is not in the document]
bresch: I'm open for suggestion. I thought about "ready" but that's commonly used to say the a new sample has fallen behind the delayed horizon. Maybe `isDataPassingChecks` ? `isDataReadyToBeFused` ?

EKF/SensorRangeFinder.cpp
bresch: No, because the cos(tilt) is required by several other computations outside of SenorRangeFinder

EKF/SensorRangeFinder.hpp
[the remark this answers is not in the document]
kamilritz: Yeah, but that is why you have am private member `_range_sample` in your `SensorRangeFinder`class. Inside setSample(), you make a copy assignment of the sample to the private member. Since the sample does not contain any pointers, after the assignment the private member `_range_sample` is completely independent of the outside sample.
But passing the argument to the function by const-ref avoids the construction of an intermediate sample object. This intermediate sample would only get assigned to the private member variable before getting destroyed again. Instead we could assign the outside sample that got passed by reference.

EKF/control.cpp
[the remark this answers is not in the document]
kamilritz: OK

EKF/SensorRangeFinder.hpp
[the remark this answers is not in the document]
kamilritz: Ohh, it is the cosinus of the tilt angle. I perceived the cosinus of the tilt angle as "tilt". Sry.

EKF/SensorRangeFinder.cpp
kamilritz: You're right. In most places it is only used after passing one of the healthy checks. Except for this case here: https://github.com/PX4/ecl/pull/782/files#diff-6ff089a846fa4bd452bf076e28a0bef7R70  . What a pitty! ;)

EKF/SensorRangeFinder.hpp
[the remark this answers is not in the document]
bresch: Of course, yes, I think I was tired xD

bresch:
@kamilritz I started to address your comments, but I'm not done yet

EKF/terrain_estimator.cpp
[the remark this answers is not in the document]
bresch: After discussion with @priseborough it makes sense to change that but not directly in this PR. I'll do an other one with that change following this PR

test/test_SensorRangeFinder.cpp
```diff
@@ -0,0 +1,253 @@
...
+	new_sample.quality = 0;
+	for (int i = 0; i < int(stuck_timeout / dt); i++) {
+		_range_finder.setSample(new_sample);
+		_range_finder.runChecks(new_sample.time_us, attitude);
+		new_sample.time_us += dt;
+	}
+	EXPECT_FALSE(_range_finder.isDataHealthy());
+
+	new_sample.quality = 100;
+	// we need a few sample to pass the hysteresis check
```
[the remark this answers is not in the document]
bresch: Actually it also took me a bit of time to understand what the check actually did and it doesn't check for stucked values all the time, but only if the samples were not valid for at least 10 seconds: https://github.com/PX4/ecl/blob/544385fd8c34957a0c65160c3141b0bb255af3b8/EKF/range_finder_checks.cpp#L112-L113
I also expected it to check all the time for "frozen measurements" but it's apparently not the intent of the original implementation.
So, to make it fail, we need to:
1. give bad samples for at least 10s
2. give good samples but with a constant range (it first need to pass the hysteresis and then is blocked by the "stuck" detector)

bresch:
@priseborough  I added more tilt tests, see https://github.com/PX4/ecl/pull/782/commits/8a5c82edc8180803b4b21f196a09135b8e360e2e

EKF/Sensor.hpp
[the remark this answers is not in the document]
kamilritz: `isDataReadyAndValid` sounds good.

EKF/sensor_range_finder.hpp
```diff
@@ -67,11 +67,11 @@ class SensorRangeFinder : public Sensor
 	// TODO: move the ring buffer here
 	rangeSample* getSampleAddress() { return &_sample; }
 
-	void setSensorTilt(float new_tilt)
+	void setTiltOffset(float new_tilt_offset)
 	{
-		if (fabsf(_tilt - new_tilt) > FLT_EPSILON) {
-			_sin_tilt_rng = sinf(new_tilt);
-			_cos_tilt_rng = cosf(new_tilt);
+		if (fabsf(_tilt_offset_rad - new_tilt_offset) > FLT_EPSILON) {
```
kamilritz, line 72: Can you explain me what the `_tilt_offset_rad` variable is used for? IMO the variable name does not match the current functionality.

test/test_SensorRangeFinder.cpp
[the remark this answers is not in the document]
kamilritz: Ok. Actually this seems more like a test that should happens on a driver level. But that is a story for another time.

EKF/sensor_range_finder.cpp
```diff
@@ -0,0 +1,139 @@
...
+{
+	updateSensorToEarthRotation(R_to_earth);
+	updateValidity(current_time_us);
+}
+
+void SensorRangeFinder::updateSensorToEarthRotation(const Dcmf &R_to_earth)
+{
+	// calculate 2,2 element of rotation matrix from sensor frame to earth frame
+	// this is required for use of range finder and flow data
+	_cos_tilt_rng_to_earth = R_to_earth(2, 0) * _sin_tilt_rng + R_to_earth(2, 2) * _cos_tilt_rng;
```
kamilritz, line 58: This tilt is in the pitch direction, right? We could maybe store the offset as a `sensor_pitch_offset`. And directly do `cos(sensor_pitch_offset)` and `sin(sensor_pitch_offset)` here, as the offset is only used here. And change the setter function to `setPitchOffset()`

EKF/sensor_range_finder.hpp
bresch: Good catch, the purpose is to avoid re-computing the cos and sin function every time, but since I never store the value, it's obviously not working

EKF/sensor_range_finder.cpp
bresch: Yes, but we only want to recompute cos and sin of tilt is the parameter changed, this is why we store them

EKF/sensor_range_finder.hpp
bresch: updated, you can re-approve

PR #783, opened 2020-03-19
Terrain estimator: Add sensor aiding bitmask

bresch:
This adds the ability to select range finder and/or optical flow fusion for terrain estimate.

Reason: I need to disable optical flow fusion for terrain estimate on a multirotor.

Note that I wanted to use `enum class` but then I have to define all the operators used for bitmasks and it adds to much code boilerplate for little benefit.

Firmware PR: https://github.com/PX4/Firmware/pull/14427

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -179,6 +179,11 @@ struct auxVelSample {
 #define MASK_USE_GPSYAW  (1<<7)		///< set to true to use GPS yaw data if available
 #define MASK_USE_EVVEL  (1<<8)		///< sset to true to use external vision velocity data
 
+enum TerrainFusionMask : int32_t {
+	TerrainFuseRangeFinder = (1 << 0),
+	TerrainFuseOpticalFlow = (1 << 1)
+};
+
 // Integer definitions for mag_fusion_type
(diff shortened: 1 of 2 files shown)


RomanBapst:
@bresch I assume you are going to do a firmware PR?

EKF/common.h
```diff
@@ -204,6 +209,7 @@ struct parameters {
 	// measurement source control
 	int32_t fusion_mode{MASK_USE_GPS};		///< bitmasked integer that selects which aiding sources will be used
 	int32_t vdist_sensor_type{VDIST_SENSOR_BARO};	///< selects the primary source for height data
+	int32_t terrain_fusion_mode{3};			///< aiding source(s) selection bitmask for the terrain estimator
```
dagar: It doesn't really matter, but you could use the actual bits here.

```suggestion
	int32_t terrain_fusion_mode{TerrainFuseOpticalFlow | TerrainFuseRangeFinder};			///< aiding source(s) selection bitmask for the terrain estimator
```

bresch: Yes, thanks

PR #788, opened 2020-03-29
Refactor velocity reset

kamilritz:
Todo:
- [x] Expand unit tests for velocity resets.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -81,8 +81,8 @@ struct outputSample {
 };
 
 struct outputVert {
-	float	    vel_d;		///< D velocity calculated using alternative algorithm (m/sec)
-	float	    vel_d_integ;	///< Integral of vel_d (m)
+	float	    pos_deriv;		///< position derivative calculated using alternative algorithm (m/sec)
+	float	    pos_deriv_integ;	///< Integral of position derivative (m)
 	float	    dt;			///< delta time (sec)
 	uint64_t    time_us;		///< timestamp of the measurement (uSec)
 };
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -81,8 +81,8 @@ struct outputSample {
 };
 
 struct outputVert {
-	float	    pos_deriv;		///< position derivative calculated using alternative algorithm (m/sec)
-	float	    pos_deriv_integ;	///< Integral of position derivative (m)
+	float	    vert_vel;		///< Vertical velocity calculated using alternative algorithm (m/sec)
+	float	    vert_vel_integ;	///< Integral of vertical velocity (m)
 	float	    dt;			///< delta time (sec)
 	uint64_t    time_us;		///< timestamp of the measurement (uSec)
 };
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -132,7 +132,7 @@ void Ekf::resetVerticalVelocityTo(float new_vert_vel) {
 	}
 	_output_new.vel(2) += delta_vert_vel;
 	_output_vert_delayed.pos_deriv = new_vert_vel;
-	_output_vert_new.pos_deriv = new_vert_vel;
+	_output_vert_new.pos_deriv += delta_vert_vel;
 
 	_state_reset_status.velD_change = delta_vert_vel;
 	_state_reset_status.velD_counter++;
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -128,11 +128,11 @@ void Ekf::resetVerticalVelocityTo(float new_vert_vel) {
 
 	for (uint8_t index = 0; index < _output_buffer.get_length(); index++) {
 		_output_buffer[index].vel(2) += delta_vert_vel;
-		_output_vert_buffer[index].pos_deriv += delta_vert_vel;
+		_output_vert_buffer[index].vert_vel += delta_vert_vel;
 	}
 	_output_new.vel(2) += delta_vert_vel;
-	_output_vert_delayed.pos_deriv = new_vert_vel;
-	_output_vert_new.pos_deriv += delta_vert_vel;
+	_output_vert_delayed.vert_vel = new_vert_vel;
+	_output_vert_new.vert_vel += delta_vert_vel;
 
 	_state_reset_status.velD_change = delta_vert_vel;
 	_state_reset_status.velD_counter++;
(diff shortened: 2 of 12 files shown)


EKF/ekf_helper.cpp
```diff
@@ -99,35 +90,52 @@ bool Ekf::resetVelocity()
...
+	return true;
+}
+
+void Ekf::resetVelocityTo(const Vector3f &new_vel) {
+	resetHorizontalVelocityTo(Vector2f(new_vel.xy()));
+	resetVerticalVelocityTo(new_vel(2));
+}
+
+void Ekf::resetHorizontalVelocityTo(const Vector2f &new_horz_vel) {
+	const Vector2f delta_horz_vel = new_horz_vel - Vector2f(_state.vel.xy());
```
bresch: There is a constructor from Vector3f to Vector2f, you can then just use `Vector2f(_state.vel)`

EKF/ekf_helper.cpp
```diff
@@ -99,35 +90,52 @@ bool Ekf::resetVelocity()
...
+	_state.vel(0) = new_horz_vel(0);
```
bresch: Can't we use `_state.vel.xy() = new_horz_vel` here?

EKF/ekf_helper.cpp
```diff
@@ -297,67 +303,45 @@ void Ekf::resetHeight()
...
 	// add the reset amount to the output observer buffered data
 	for (uint8_t i = 0; i < _output_buffer.get_length(); i++) {
 		if (vert_pos_reset) {
 			_output_buffer[i].pos(2) += _state_reset_status.posD_change;
 			_output_vert_buffer[i].vel_d_integ += _state_reset_status.posD_change;
 		}
-
-		if (vert_vel_reset) {
-			_output_buffer[i].vel(2) += _state_reset_status.velD_change;
-			_output_vert_buffer[i].vel_d += _state_reset_status.velD_change;
```
bresch: Is that not required anymore?

EKF/ekf_helper.cpp
```diff
@@ -297,67 +303,45 @@ void Ekf::resetHeight()
...
 
 	// add the reset amount to the output observer vertical position state
 	if (vert_pos_reset) {
 		_output_vert_delayed.vel_d_integ = _state.pos(2);
 		_output_vert_new.vel_d_integ = _state.pos(2);
 	}
-
-	if (vert_vel_reset) {
-		_output_vert_delayed.vel_d = _state.vel(2);
-		_output_vert_new.vel_d = _state.vel(2);
```
bresch: And that?

EKF/ekf_helper.cpp
kamilritz: Cool

EKF/ekf_helper.cpp
kamilritz: I am not sure. I will try after adding the tests.

EKF/ekf_helper.cpp
kamilritz: Thanks. Missed that.

EKF/ekf_helper.cpp
kamilritz: As well

EKF/ekf_helper.cpp
kamilritz: @bresch @priseborough  I guess this would also be needed if we reset the vertical velocity in the `resetVelocity()` function. Or is there a reason to apply this only inside `resetHeight()`?

And is there a reason why we add here a delta to the `vel_d` member and below for the delayed and new sample we set it to the new value instead of adding a delta as well?

kamilritz:
@bresch I added unit tests for the resets. By doing so I stumbled upon a bug. 
If we do a reset to optical flow velocity, we will always reset to zero in the beginning. Even if we are in air and get reasonable flow data. The issue is that the `_flow_compensated_XY_rad` which is used inside `resetVelocity()` will be updated after the call to `resetVelocity()` inside `controlOpticalFlowFusion()` . Also if we would do another reset to the flow sensor, we would reset to the last stored value in _flow_compensated_XY_rad, which can be already outdated.

EKF/ekf_helper.cpp
priseborough: As a general rule the output observer should always be reset following a state reset.

EKF/ekf_helper.cpp
priseborough: Ditto.

priseborough:
Lets log an issue for the optical flow reset change and handle it in a separate PR as a functional change.

EKF/ekf_helper.cpp
kamilritz: This is working. Will change it

kamilritz:
@bresch @priseborough I updated the PR as discussed. Therefore it is not completely non-functional, since we update the vertical output buffer more often. From my side this is ready. The only odd thing is that [here](https://github.com/PX4/ecl/pull/788/files#diff-6ff089a846fa4bd452bf076e28a0bef7R131-R136),  the `output_*` quantities are updated each time with `=+` operator, while this is not the case for the `output_vert_*` quantities.

priseborough:
> @bresch @priseborough I updated the PR as discussed. Therefore it is not completely non-functional, since we update the vertical output buffer more often. From my side this is ready. The only odd thing is that [here](https://github.com/PX4/ecl/pull/788/files#diff-6ff089a846fa4bd452bf076e28a0bef7R131-R136), the `output_*` quantities are updated each time with `=+` operator, while this is not the case for the `output_vert_*` quantities.

As a general rule if updating output observer at current (most recent) time, then it is better to += a delta. If updating at delayed or fusion time then it is best to use an = current state from EKF.

kamilritz:
@priseborough Can you clarify if output_vert.vel_d  is a velocity or a delta_velocity/acceleration? If it is a delta_velocity, the assignment at most recent time could still make sense.

priseborough:
> @priseborough Can you clarify if output_vert.vel_d is a velocity or a delta_velocity/acceleration? If it is a delta_velocity, the assignment at most recent time could still make sense.

It's the vertical position time derivative (velocity) state and is at the current time horizon. It is used to populate the vehicle_local_position.z_deriv message.

kamilritz:
@priseborough But if we reset the velocity, how should we reset the velocity derivative?
https://github.com/PX4/ecl/blob/master/EKF/ekf.cpp#L371-L372

kamilritz:
@priseborough I was confused by the variable name `vel_d`, which for me stands for a derivative of a velocity. I updated now the naming such that `vel_d` -> `pos_deriv` and `vel_d_integ` -> `pos_deriv_integ`.

kamilritz:
> As a general rule if updating output observer at current (most recent) time, then it is better to += a delta. If updating at delayed or fusion time then it is best to use an = current state from EKF.

I updated the update of `output_vert_new` according to this guideline.

priseborough:
> @priseborough I was confused by the variable name `vel_d`, which for me stands for a derivative of a velocity. I updated now the naming such that `vel_d` -> `pos_deriv` and `vel_d_integ` -> `pos_deriv_integ`.

The _d subscript was to indicate it was for the down axis, so don't lose that distinction

kamilritz:
> The _d subscript was to indicate it was for the down axis, so don't lose that distinction

@priseborough You are completely right, how did I overlook that. I gave it another try.

kamilritz:
@bresch LGTM

EKF/common.h
```diff
@@ -81,8 +81,8 @@ struct outputSample {
 };
 
 struct outputVert {
-	float	    vel_d;		///< D velocity calculated using alternative algorithm (m/sec)
-	float	    vel_d_integ;	///< Integral of vel_d (m)
+	float	    vert_vel;		///< Vertical velocity calculated using alternative algorithm (m/sec)
```
bresch, line 84: it seems a bit redundant to me to specify that the velocity is the vertical one in a struct dedicated to the vertical axis

kamilritz: Agree, had the same impression.

PR #792, opened 2020-04-03
TerrainEstimator initialization fix

bresch:
Following #782, we identified an inconsistency where the newest sample was used to initialize the terrain estimator instead of the delayed -and verified- measurement (fixed in https://github.com/PX4/ecl/commit/45de501e19a2bf866bfbc60b48d96683b24f43b2 ).

Additionally, in https://github.com/PX4/ecl/commit/b4f35d47bf95c3832446ce5ac085b09ab6ad154a I added the `distBottom` function to the `SensorRangeFinder` class to directly get the vertical measured distance below the drone, corrected for tilt.

**Please answer** https://github.com/PX4/ecl/pull/792/files#r402776911

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1001,7 +1001,7 @@ void Ekf::controlHeightFusion()
 					_hgt_sensor_offset = _terrain_vpos;
 
 				} else if (_control_status.flags.in_air) {
-					_hgt_sensor_offset = _range_sensor.getCosTilt() * _range_sensor.getRange() + _state.pos(2);
+					_hgt_sensor_offset = _range_sensor.getDistBottom() + _state.pos(2);
 
 				} else {
 					_hgt_sensor_offset = _params.rng_gnd_clearance;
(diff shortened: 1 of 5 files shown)


EKF/control.cpp
```diff
@@ -1161,10 +1161,11 @@ void Ekf::controlHeightFusion()
 			Vector2f rng_hgt_innov_gate;
 			Vector3f rng_hgt_obs_var;
 			// use range finder with tilt correction
-			_rng_hgt_innov(2) = _state.pos(2) - (-math::max(_range_sensor.getRange() * _range_sensor.getCosTilt(),
+			_rng_hgt_innov(2) = _state.pos(2) - (-math::max(_range_sensor.getDistBottom(),
 							 _params.rng_gnd_clearance)) - _hgt_sensor_offset;
 			// observation variance - user parameter defined
-			rng_hgt_obs_var(2) = fmaxf((sq(_params.range_noise) + sq(_params.range_noise_scaler * _range_sensor.getRange())) * sq(_range_sensor.getCosTilt()), 0.01f);
+			rng_hgt_obs_var(2) = fmaxf(sq(_params.range_noise)
+						   + sq(_params.range_noise_scaler * _range_sensor.getDistBottom()), 0.01f);
```
bresch, line 1168: Do we really want to scale the range noise with the distance below the drone and not the measured distance?

kamilritz: I am not sure what is better.

jkflying: I guess there's 2 things adding noise:

1) Sensor characteristics
2) Terrain unevenness

1) scales with measured distance, 2) scales with distance to ground and vehicle orientation change rate (looking at different points on the ground).

Not sure what the right solution is here, or which component dominates.

PR #794, opened 2020-04-08
Alpha filter merge with PX4 version

MaEtUgR:
For the battery filter time abstraction in PX4 https://github.com/PX4/Firmware/pull/14551/ I implemented the same filter there and am now merging the two versions here.

It's the best of both sides I think except I couldn't do the NAN checks because `matrix::Vectors` do not support `isfinite` checks. So I have to take care with the PX4 callers that they never run NAN through the filter.

Sidenote: While working with the unit tests I missed the gtest colors so I added them with `GTEST_COLOR=1` like in PX4.

### DIFF
--- a/EKF/AlphaFilter.hpp
+++ b/EKF/AlphaFilter.hpp
@@ -1,6 +1,6 @@
 /****************************************************************************
  *
- *   Copyright (c) 2019 PX4 Development Team. All rights reserved.
+ *   Copyright (c) 2019-2020 PX4 Development Team. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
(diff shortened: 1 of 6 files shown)


MaEtUgR:
It was quite a hassle to get `make format` use `clang-format version 9.0.1`. It would helpful if the CI output would directly show what's broken in the style or it's easier to adopt the style check locally.

MaEtUgR:
Why exactly is that diff CI check failing?

jkflying:
> Why exactly is that diff CI check failing?

This should fail if there is behavioral changes in the code. You need to re-run the test locally and commit the changes to the file it makes.

MaEtUgR:
Yeah but I want to find out why/what changed in it. Do I just revert things and see where it changes?

jkflying:
Basically yes, it is a final safety check for PRs that should be non-functional changes.

MaEtUgR:
My solution to not spend the day fiddling with docker because my compiler uses slightly different printf formating is to revert back to the internal definition of `alpha_mathieu = 1 - alpha_matthias` which results in no numerical differences. I used the other definition because almost all places where I could find the filter definition it was this way but in practise it doesn't matter if you know which way it is. And the time abstraction formula is the just `dt / (dt + tau)` instead of `tau / (dt + tau)` which leaves the setParameters inteface the same.

Any opinions? Anyone up for a review?

test/test_AlphaFilter.cpp
```diff
@@ -39,37 +45,36 @@
 
 using matrix::Vector3f;
 
-class AlphaFilterTest : public ::testing::Test {
```
jkflying, line 42: What was the reason for getting rid of the test fixture? Now it needs to be instantiated in each test instead of once at the top.

EKF/AlphaFilter.hpp
```diff
@@ -32,32 +32,60 @@
...
+	 * @param sample new initial value
+	 */
+	void reset(const T &sample) { _filter_state = sample; }
+
+	/**
+	 * Add a new raw value to the filter
+	 *
+	 * @return retrieve the filtered result
+	 */
+	virtual void update(const T &sample) { _filter_state = updateCalculation(sample); }
```
jkflying, line 82: Why is it virtual? Does it get extended?

dagar:
If you're planning to use this everywhere it probably makes sense to move it out of EKF to the top level of ecl.

test/test_AlphaFilter.cpp
MaEtUgR: Because I confused the two instances once and screwed up by setting the parameters on the wrong one and it silently passed. So I thought why not make it extra clear since they are completely exclusive to single tests. Having two fixtures to have two template instances seemed overkill.

MaEtUgR:
> If you're planning to use this everywhere it probably makes sense to move it out of EKF to the top level of ecl.

That makes sense. I wasn't sure about the structure of ecl. I think it's actually my first contribution.

MaEtUgR:
> maybe adding isfinite checks in Matrix would make things easier?

@jkflying That would allow having that nan checks in the templated filter. But I was thinking about the real case where you get nans and suddenly the filter would silently not update anymore making you think your data is still available and filtered working with a stale output. When I followed the history behind that nan check it seems it was just precautiously added for one case and then copy-pasted to all battery filters in a refactor.

EKF/AlphaFilter.hpp
bresch: @MaEtUgR Is the plan to implement the same filter but with a saturation at the input maybe?

MaEtUgR: No, it doesn't get extended yet. I can remove this 👍

bresch, line 82:
```suggestion
	void update(const T &sample) { _filter_state = updateCalculation(sample); }
```

MaEtUgR:
Thanks @bresch !! 👍

Issue #771, opened 2020-03-09
Vehicle crash due do estimator GPS failure


ndepal:
We just had another vehicle crash. EKF2 reported a GPS glitch and a delta velocity bias estimate issue.

The firmware is based on PX4 v1.10.0.
We are using the Ublox F9P GPS module in the vehicle and on the ground.

The vehicle took off in mission mode and briefly after started drifting when it crashed into a tree.
At the point of takeoff, the vehicle had been turned on for ~6 hours.

The symptoms are very similar to the crash mentioned in https://github.com/PX4/Firmware/pull/14312#issuecomment-595762171, but this time we do have a log, fortunately. This crash did not happen with the same vehicle.

Looking at the log, the `vehicle_gps_position` looks sane, whereas the `vehicle_global_position` and the local position does not. Therefore, I feel that the issue is not with the GPS measurements (i.e. they should not have been rejected) but rather with the estimator that falsely rejected the measurements.

As this is not the first incident of this kind, I would really appreciate some assistance in looking into this and preventing it from happening in the future.

The flight log can be found here: https://logs.px4.io/plot_app?log=c231b7b3-ab35-4f48-be8c-0cb0a836cae4

CarlOlsson:
Without more context and the preflight log it is hard to say for sure. But you definitely have magnetic interference on the ground disturbing the estimated heading.
![image](https://user-images.githubusercontent.com/7795133/76204340-833fc680-61f8-11ea-864f-0e02c226fc03.png)

I would first try to remove this. Second thing you can try is to update ECL to master since there is now a yaw recovery estimator running in the background, specifically designed to handle failure cases like this
https://github.com/PX4/ecl/pull/766

LorenzMeier:
@ndepal Given your mag interference issues I will close this issue since this is not a PX4 performance issue at the core - it is a specific use case (lots of metal around the vehicle) that requires a new feature (as below):

https://github.com/PX4/ecl/pull/766

ndepal:
Thanks @CarlOlsson for having a look and pointing out the magnetic interference. That is potentially an issue.

However, EKF2 already has a feature that resets the heading to the current mag measurement once the vehicle reaches a certain altitude (1.5m above ground) https://github.com/PX4/ecl/blob/dec1ed0aeb395a885b60b1d61cfb57bd932c7d34/EKF/control.cpp#L1451
(this code looks quite different in master, the above links to what PX4 v1.10.0 uses)
[...]

PR #795, opened 2020-04-10
EKF: Improve covariance prediction stability

priseborough:
Eliminates collapse of vertical velocity state variance due to rounding errors that can occur under some operating conditions.

See issue  https://github.com/PX4/ecl/issues/771 for discussion and testing against stable release version.

TODO: testing of this version


EKF/covariance.cpp
```diff
@@ -734,12 +735,12 @@ void Ekf::fixCovarianceErrors(bool force_symmetry)
 
 	for (int i = 4; i <= 6; i++) {
 		// NED velocity states
-		P(i,i) = math::constrain(P(i,i), 0.0f, P_lim[1]);
+		P(i,i) = math::constrain(P(i,i), 1E-6f P_lim[1]);
```
jkflying, line 738: Missing a comma here
```suggestion
		P(i,i) = math::constrain(P(i,i), 1E-6f, P_lim[1]);
```

dusan19:
Would it be benefitial if the filter update period was increased to 12ms to keep it as a multiple of the incoming  imu samples?

priseborough:
> Would it be benefitial if the filter update period was increased to 12ms to keep it as a multiple of the incoming imu samples?

We can have significant dt jitter on IMU sampling depending on the board, so there is no benefit  in selecting an integer multiple given the way the imu_down_sampler.cpp is adjusting the time cutoff to maintain the average rate at the target.

priseborough:
We could move to a fixed down sampling ratio, eg 3:1, but it would rely on the published IMU data dt having low jitter.

priseborough:
For example the BMI088 sensor is sampled at 800Hz which does not downsample to 250Hz with even time steps.

priseborough:
If we went to 12, then maybe we could use the instantaneous IMU dt instead of the average dt inside the covariance prediction as the likelihood of a 1 frame IMU sample being consumed then becomes small, but I haven't tested that option. Using 12 instead of 10 will also increase timing jitter for other sensors wrt the fusion time horizon.

priseborough:
I'm working through the CI test failures one by one adjusting the unit tests to get them to pass. Can someone else have a look at the  imuSamplingAtMultipleRates test. What it is doing doesn't make sense to me, but I may be misunderstanding it.

priseborough:
The only tests failing now are:

EkfExternalVisionTest.visionVelocityReset
EkfExternalVisionTest.visionVelocityResetWithAlignment
 
It's not obvious how this change is impacting these cases, so further investigation is required.

jkflying:
@kamilritz do you have time to take a quick look at why those tests are failing?

priseborough:
I've got them all passing now.

kamilritz:
Cool, Another option is to increase the time before the reset and capture the state of after the reset a bit earlier, like this: https://github.com/kamilritz/ecl/commit/eea01467b75d5f4f822cdc950e8ec58398729dc0
I would prefer this option since with this check we want to test if the reset happens to the correct value and not if we converge towards the right value.
@priseborough Are you able to update the change indication locally? Otherwise you could copy-replace it from here: https://github.com/kamilritz/ecl/commit/528a235afb190a20c499056a445294b49a8f9d1c

dagar:
> We can have significant dt jitter on IMU sampling depending on the board, so there is no benefit in selecting an integer multiple given the way the imu_down_sampler.cpp is adjusting the time cutoff to maintain the average rate at the target.

It's significantly better with the new Invensense drivers in master (and the v1.11 release). Currently that's the primary IMU in the vast majority of usage. Note these still have the hard coded 4000 us integration period, but it's handled as an integer number of samples. We could easily change this to 2500 us.

> We could move to a fixed down sampling ratio, eg 3:1, but it would rely on the published IMU data dt having low jitter.

I would like to consider that after this release. I don't see why we can't optimize the system holistically, including accommodating these fundamental differences between sensors.

dagar:
For testing I've created https://github.com/PX4/Firmware/pull/14676 that pulls in this ECL branch and also changes the default integration period.

priseborough:
> @priseborough Are you able to update the change indication locally? Otherwise you could copy-replace it from here: [kamilritz@528a235](https://github.com/kamilritz/ecl/commit/528a235afb190a20c499056a445294b49a8f9d1c)

Not with the current level of documentation for the change indication. I know you showed me once before, but I thought you were going to write a readme that specifies the steps required to generate a new file

priseborough:
@kamilritz I pulled in your suggested test changes, but had to resolve a conflict and there are still there are issues with unit  tests failing after the rebase.

kamilritz:
@priseborough Great, that you could get those tests to pass. These tests seem to be really time sensitive. Maybe it is due to the fact that we compare the output velocities and not the velocities of the delayed state. I will look in to this.

dusan19:
I tested this changes on a log that was failing due to covariance prediction instability that I recently recorded (sorry i didnt do it sooner before this PR has been merged). The results are posted here: https://github.com/PX4/ecl/issues/771#issuecomment-618596373
As the plots show, this changes don't solve the problem but might make it slightly better, although the drone would still perform the fallback landing due to invalid position estimate (and crash)

PR #796, opened 2020-04-11
Increase precision of azimuthal equidistant re/projection

kamilritz:
Previously, when applying a reprojection and projection in sequence, a local input coordinate of size 0.5 got mapped by the composition back to the value 0.75.

Some of the operation in `geo.cpp`, where performed with floating point precision, while the rest is done in double precision. Changing this to double precision, resolved the above issue.

Not sure if this should also change for other functions in `geo.cpp`


geo/geo.cpp
```diff
@@ -135,15 +135,16 @@ int map_projection_project(const struct map_projection_reference_s *ref, double
 		return -1;
 	}
 
-	const double lat_rad = math::radians(lat);
-	const double lon_rad = math::radians(lon);
+	const double lat_rad = (lat / 180.0) * M_PI;
```
dagar, line 138: Why?

dagar: https://github.com/PX4/Firmware/blob/751b3497a049925a1230786706b2ab0a4b6cda3f/src/lib/mathlib/math/Limits.hpp#L84-L88

kamilritz:
That is a nice idea. This would mean that we would also perform the other radians<->degrees conversions in `geo.cpp` with double point precision. I guess with all the other double point operations this one does not carry much weight.

dagar:
Actually, I'm now wondering if other people are aware of how this is currently built. The mathlib within PX4/ecl is only used in the standalone build. When built within PX4/Firmware (most usage) the templated mathlib is already used.

Yes this is convoluted and should probably be fixed...

kamilritz:
Yeah, this is probably not optimal, but out of the scope of this PR. LGTM

PR #799, opened 2020-04-16
Fix baro fail-over in GPS hgt mode

bresch:
When the primary height source is GPS but the GPS data times out, EKF2 switches to baro and makes a reset. However, the `_hgt_sensor_offset` was not set to 0 and the height reset creates an incorrect jump in the state.

partially fixing https://github.com/PX4/Firmware/issues/14677

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -848,7 +848,7 @@ void Ekf::controlHeightSensorTimeouts()
 						     _gps_hgt_intermittent);
 
 			if (reset_to_baro) {
-				setControlBaroHeight();
+				startBaroHgtFusion();
 
 				request_height_reset = true;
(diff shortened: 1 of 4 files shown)


bresch:
Reproduced in SITL:
2020-04-16_18-06-23_01_plot
Shows a jump of 7m

Replayed with this fix:
2020-04-17_08-54-26_01_plot
Jump of 3m only. We still have a small jump because it's a reset and the baro offset doesn't exactly compensate (it's a filtered value). But this should be all right, the reset counter increments, the delta_z is published and should be handled by the setpoint generator.
However, I found that we have a race condition in the handling logic and I'm trying to fix it.

bresch:
I even got a better one here after going in and out of range aid a few times:
2020-04-17_14-06-43_01_plot

Replayed with this PR:
2020-04-17_14-07-17_01_plot

PR #802, opened 2020-04-24
Update RingBuffer.h

worlddeworld:
*RingBuffer: Do not reset the timestamp.
When calling 'pop_first_older_than' function, it will reset the older data timestamp to 0. If the function is called faster than the new data push in, all the timestamp of the data in the buffer will be 0. In this case, calling 'get_newest' function will return data with 0 timestamp, this will case problem for which needs the right timestamp.

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -141,7 +141,7 @@ public:
 					_tail = (index + 1) % _size;
 				}
 
-				_buffer[index].time_us = 0;
+				//_buffer[index].time_us = 0;
 
 				return true;
 			}


worlddeworld:
> @worlddeworld This is a good observation. Thank you very much for raising this.
> There is indeed an issue that one can pop from an empty buffer. There is no explicit check if the buffer is empty. And if you try to pop from an empty buffer, it can happen that it returns a sample.
> Previously this is only happening for the first 0.1 seconds, as there is a check that the data is not older than 0.1 seconds, assuming the timestamp of the empty buffer entries is initialized to 0.
> 
> @bresch @priseborough @RomanBapst I am not sure if the check for an empty buffer was avoided to increase performance. Or should this be added?
> 
> The proposed change is actually causing the "pop from empty buffer"-bug to happen also after 0.1 seconds. Imagine if the buffer is empty due to low sensor update rate and small delay. Therefore head and tail will point to the last sample, that was popped from the buffer. If the sample is still new (not older than 0.1) it will be returned again, since it is still passing the timing checks. Setting the timestamp of a popped sample to 0 is inhibiting the same sample to be popped multiple times.

>@kamilritz  I will analyze a real problem what happened in my test. 
*Problem: When I use RTK in flight, two warning occurred alternately, which are "baro hgt timeout - reset to GPS" and "gps hgt timeout - reset to baro".

*My analysis: First, in function "controlFusionModes" of EKF in control.cpp,  we check "_gps_hgt_intermittent"  whith 
               const gpsSample &gps_init = _gps_buffer.get_newest();
	      _gps_hgt_intermittent = !isRecent(gps_init.time_us, 2 * GPS_MAX_INTERVAL);
Where '_gps_hgt_intermittent' will be true if calling 'pop_first_older_than' function faster than the new data push in, because all the timestamp had been reset to 0 and gps_init.time_us=0.

Then, in function "controlHeightSensorTimeouts", when '_control_status.flags.gps_hgt' is true which means gps height used, we will check gps height healthy with  '_gps_hgt_intermittent'. As gps updating slowly, we calling '_gps_buffer.pop_first_older_than' function faster than the new gps data push in, '_gps_hgt_intermittent' will be true between times, which will put the warning "gps hgt timeout - reset to baro" and switch to use baro height. When gps data updated, gps height will be ok, so we put the warning "baro hgt timeout - reset to gps" and switch to use gps height. And the two warning will occur alternately.

But, I checked the logic again, not reseting the tiemstamp also cause another problem that "pop_first_older_than" will get the duplicated data in 0.1s if no new data push in. I try to give another solution, that is, define a new variable to record the newest  data, and "get_newest" just return this value.  Like this,
-----------------------------------------------
         data_type _newest;

        void push(const data_type &sample) {

		uint8_t head_new = _head;

		if (!_first_write) {
			head_new = (_head + 1) % _size;
		}

		_buffer[head_new] = sample;
		_head = head_new;

		// move tail if we overwrite it
		if (_head == _tail && !_first_write) {
			_tail = (_tail + 1) % _size;

		} else {
			_first_write = false;
		}

                //record the newest data
               _newest = sample;
	}

      const data_type &get_newest() { return _newest; }  //just return  _newest

priseborough:
Please resubmit when the underlying problem with this approach has been addressed.

PR #803, opened 2020-04-26
Avoid subtraction of two uint inside RingBuffer

kamilritz:
Subtraction of two unsigned integers can be dangerous. I think in this case it does not really matter, but it is better to be safe.

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -127,7 +127,7 @@ public:
 			int index = (_head - i);
 			index = index < 0 ? _size + index : index;
 
-			if (timestamp >= _buffer[index].time_us && timestamp - _buffer[index].time_us < (uint64_t)1e5) {
+			if (timestamp >= _buffer[index].time_us && timestamp < _buffer[index].time_us + (uint64_t)1e5) {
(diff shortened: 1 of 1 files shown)


EKF/RingBuffer.h
```diff
@@ -127,7 +127,7 @@ class RingBuffer {
 			int index = (_head - i);
 			index = index < 0 ? _size + index : index;
 
-			if (timestamp >= _buffer[index].time_us && timestamp - _buffer[index].time_us < (uint64_t)1e5) {
+			if (timestamp >= _buffer[index].time_us && timestamp < _buffer[index].time_us + (uint64_t)1e5) {
```
dagar, line 130: Why half a second?

kamilritz: It is 0.1 seconds.
But I don't know why.

dagar: Oops, confused with 5e5 that we use in a few other places.

dagar: It would be nice to document if we get an answer, unrelated to the potential change/fix though.

PR #807, opened 2020-05-06
EKF: mag_control: Reset yaw to mag when restarting fusion

mhkabir:
When mag fusion was stopped (e.g when EV yaw data fusion was started) and restarted, the yaw angle was not reset to that provided by the mag, leading to filter divergence.

The fix is to make sure we reset the yaw when re-starting mag fusion.

**Before this PR:**
image

**After this PR:**
image

**EV Data is stopped at the 79s mark, and restarted at the 95s mark**

Also see https://github.com/PX4/Firmware/issues/14840 where this discovered.

Flight log which illustrates old issue: https://logs.px4.io/plot_app?log=c0a43438-fb28-45fd-9a5b-92329748fb93. 

The fix was tested by running replay on the above log.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -331,7 +331,7 @@ void Ekf::controlExternalVisionFusion()
 			fuseHeading();
 		}
 
-	} else if ((_control_status.flags.ev_pos || _control_status.flags.ev_vel ||  _control_status.flags.ev_yaw)
+	} else if ((_control_status.flags.ev_pos || _control_status.flags.ev_vel || _control_status.flags.ev_yaw)
 		   && isTimedOut(_time_last_ext_vision, (uint64_t)_params.reset_timeout_max)) {
 
 		// Turn off EV fusion mode if no data has been received
--- a/EKF/mag_control.cpp
+++ b/EKF/mag_control.cpp
@@ -75,6 +75,9 @@ void Ekf::controlMagFusion()
 	_mag_yaw_reset_req |= otherHeadingSourcesHaveStopped();
 
 	if (noOtherYawAidingThanMag() && _mag_data_ready) {
+		// Restarting mag aiding after a period of no aiding
+		_mag_yaw_reset_req |= !_control_status.flags.mag_hdg;
+
 		if (_control_status.flags.in_air) {
 			checkHaglYawResetReq();
 			runInAirYawReset();
--- a/EKF/mag_control.cpp
+++ b/EKF/mag_control.cpp
@@ -76,7 +76,7 @@ void Ekf::controlMagFusion()
 
 	if (noOtherYawAidingThanMag() && _mag_data_ready) {
 		// Restarting mag aiding after a period of no aiding
-		_mag_yaw_reset_req |= !_control_status.flags.mag_hdg;
+		_mag_yaw_reset_req |= !_control_status.flags.mag_hdg && !_control_status.flags.mag_3D;
 
 		if (_control_status.flags.in_air) {
 			checkHaglYawResetReq();


EKF/control.cpp
```diff
@@ -391,7 +391,7 @@ void Ekf::controlExternalVisionFusion()
 			fuseHeading();
 		}
 
-	} else if ((_control_status.flags.ev_pos || _control_status.flags.ev_vel)
+	} else if ((_control_status.flags.ev_pos || _control_status.flags.ev_vel || _control_status.flags.ev_yaw)
 		   && isTimedOut(_time_last_ext_vision, (uint64_t)_params.reset_timeout_max)) {
```
kamilritz, line 396:
```suggestion
	} else if ((_control_status.flags.ev_pos || _control_status.flags.ev_vel || _control_status.flags.ev_yaw)
		   && isTimedOut(_time_last_ext_vision, (uint64_t)_params.reset_timeout_max)) {
if(_control_status.flags.ev_yaw) {_mag_yaw_reset_req=true};
```

kamilritz (changes requested):
Thanks for solving this. 
At the moment we have a variable `_mag_yaw_reset_req` that is used to request mag resets. I suggest that we use this variable for solving this. 
I understand that it is not nice that each heading sensor needs to activate the default heading sensor, but I prefer to have not different solutions for this.

mhkabir:
Can you explain your rationale?

kamilritz:
Sure. Apparently there are certain situations where a heading reset is needed. At the moment for this to happen the part of the software that realizes that a reset is necessary, requests a reset through the `_mag_yaw_reset_req` variable. 

You are adding a second option for the reset to happen by `restarting_fusion ||`. This is changing the functionality of the `run*YawReset` function a lot as it is now also responsible for finding out if it should do a reset. Previously it was only told from other parts of the software. The function will do now both which I find not that clean.

I would prefer if the nature of the function stays the same and you command it to do a reset from else where by setting the `_mag_yaw_reset_req` variable. As a benefit you don't have to add the line `const bool restarting_fusion = !_control_status.flags.mag_hdg && canRunMagFusion();` twice.

If you don't want to trigger the reset from the `controlExternalVision()` call, you could also do this from the controlMagFusion function with `_mag_yaw_reset_req = !_control_status.flags.mag_hdg;` right after [this line](https://github.com/PX4/ecl/blob/master/EKF/mag_control.cpp#L73).

If you want to do me a favor you could also rename the `canRunMagFusion()` to `shouldRunMagFusion()`, which would make it more clear.

mhkabir:
@kamilritz your changes are applied, and I rebased. Will give this state a test now.

kamilritz:
@mhkabir Did you have a chance to test the new version?

mhkabir:
Ah yes, it worked fine. I forgot to pull logs, but I flew it several times and things seemed to be fine.

mhkabir:
I have updated the change_indicator reference outptuts @kamilritz @priseborough. This should be good to merge, and is a crucial bugfix for PX4 v1.11.

EKF/mag_control.cpp
```diff
@@ -70,7 +70,8 @@ void Ekf::controlMagFusion()
 		return;
 	}
 
-	if (canRunMagFusion()) {
+	if (shouldRunMagFusion()) {
+		_mag_yaw_reset_req |= !_control_status.flags.mag_hdg;
```
kamilritz, line 74: I guess this is changing the change indicator output. I am a bit surprised that this change is also affecting the output for a model that is not using any vision measurement.

kamilritz (dismissed):
It is a bit a surprise that the change indicator need to be updated. But the difference is minor.

mhkabir:
@kamilritz @bresch can we get this in

kamilritz:
@mhkabir This should be ready to get in. I do not have merge rights, so there is nothing I can do about it.

EKF/mag_control.cpp
bresch: This will continuously request resets in `mag_3D` mode.

bresch (changes requested):

priseborough (dismissed):
OK with this going in if the change indicator file is updated

dagar:
What happened here? Should we rebase this?

priseborough:
Yes it needs to be rebased and the change indicator file updated which was what was holding up the approval and merge last time.

dagar:
Rebased.

PR #808, opened 2020-05-08
Random cleanup

bresch:
I ran a static analysis using `cppcheck --enable=style` and found a few things to fix.

### DIFF
--- a/EKF/EKFGSF_yaw.h
+++ b/EKF/EKFGSF_yaw.h
@@ -60,25 +60,25 @@ private:
 
 	// Declarations used by the bank of N_MODELS_EKFGSF AHRS complementary filters
 
-	Vector3f _delta_ang;	// IMU delta angle (rad)
-	Vector3f _delta_vel;	// IMU delta velocity (m/s)
-	float _delta_ang_dt;	// _delta_ang integration time interval (sec)
-	float _delta_vel_dt;	// _delta_vel integration time interval (sec)
-	float _true_airspeed;	// true airspeed used for centripetal accel compensation (m/s)
+	Vector3f _delta_ang{};	// IMU delta angle (rad)
+	Vector3f _delta_vel{};	// IMU delta velocity (m/s)
+	float _delta_ang_dt{};	// _delta_ang integration time interval (sec)
+	float _delta_vel_dt{};	// _delta_vel integration time interval (sec)
+	float _true_airspeed{};	// true airspeed used for centripetal accel compensation (m/s)
 
 	struct _ahrs_ekf_gsf_struct{
-		Dcmf R;				// matrix that rotates a vector from body to earth frame
-		Vector3f gyro_bias;		// gyro bias learned and used by the quaternion calculation
-		bool aligned{false};		// true when AHRS has been aligned
-		float vel_NE[2] {};		// NE velocity vector from last GPS measurement (m/s)
-		bool fuse_gps = false;		// true when GPS should be fused on that frame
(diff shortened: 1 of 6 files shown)


EKF/EKFGSF_yaw.h
```diff
@@ -60,25 +60,25 @@ class EKFGSF_yaw
...
-		float accel_dt = 0;		// time step used when generating _simple_accel_FR data (sec)
-	};
-	_ahrs_ekf_gsf_struct _ahrs_ekf_gsf[N_MODELS_EKFGSF];
-	bool _ahrs_ekf_gsf_tilt_aligned = false;// true the initial tilt alignment has been calculated
-	float _ahrs_accel_fusion_gain;		// gain from accel vector tilt error to rate gyro correction used by AHRS calculation
-	Vector3f _ahrs_accel;			// low pass filtered body frame specific force vector used by AHRS calculation (m/s/s)
-	float _ahrs_accel_norm;			// length of _ahrs_accel specific force vector used by AHRS calculation (m/s/s)
+		Dcmf R;			// matrix that rotates a vector from body to earth frame
+		Vector3f gyro_bias;	// gyro bias learned and used by the quaternion calculation
+		bool aligned;		// true when AHRS has been aligned
```
kamilritz, line 72: To what will these struct members be initialized?

bresch: To zero (see https://github.com/PX4/ecl/pull/808/files#diff-42b69a6a10d9759dc44d3de61bb830b2R76)

kamilritz: So if we would have it manually initialized to true, cppcheck would have left the line?

bresch: cppcheck just complained that some variables were uninitialized. I could have added the default values in the struct declaration, but I believe it isn't a good practice. You should only specify the init values when you get an instance of it e.g.: 
```c++
_ahrs_ekf_gsf_struct _ahrs_ekf_gsf[N_MODELS_EKFGSF] = {{}, {}, true, {}.....};
```
Here, since it is the only instance, I directly declared `_ahrs_ekf_gsf[N_MODELS_EKFGSF]{}` at the end on the struct definition

kamilritz: Thanks

bresch:
Yes, it's really cool!

EKF/EKFGSF_yaw.h
```diff
@@ -98,18 +98,18 @@ class EKFGSF_yaw
...
-		float S_det_inverse; // inverse of the innovation covariance matrix determinant
-		matrix::Vector2f innov; // Velocity N,E innovation (m/s)
-	};
-	_ekf_gsf_struct _ekf_gsf[N_MODELS_EKFGSF];
-	bool _vel_data_updated;  // true when velocity data has been updated
-	bool _run_ekf_gsf;       // true when operating condition is suitable for to run the GSF and EKF models and fuse velocity data
-	Vector2f _vel_NE;        // NE velocity observations (m/s)
-	float _vel_accuracy;     // 1-sigma accuracy of velocity observations (m/s)
-	bool _ekf_gsf_vel_fuse_started;	// true when the EKF's have started fusing velocity data and the prediction and update processing is active
+		matrix::Vector3f X; 				// Vel North (m/s),  Vel East (m/s), yaw (rad)s
```
dagar, line 101: FYI it should probably be tabs for indentation spaces for alignment.

bresch: Thanks, I'll keep that in mind for next time

PR #809, opened 2020-05-09
Fix ev_pos_obs_var(1) entry

kamilritz:
What would we do without the lower bound protection?


kamilritz:
@bresch Lets merge this!

PR #810, opened 2020-05-09
Remove duplicate code during vision yaw reset

kamilritz:
The `resetQuatStateYaw` function is already providing the functionality for resetting the yaw to a certain angle.

Would be cool, if this is merged together with the [vision velocity in body frame](https://github.com/PX4/ecl/pull/708), since it will profit from this flash size reduction.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -216,41 +216,14 @@ void Ekf::controlExternalVisionFusion()
 
 		// external vision yaw aiding selection logic
 		if (!_control_status.flags.gps && (_params.fusion_mode & MASK_USE_EVYAW) && !_control_status.flags.ev_yaw && _control_status.flags.tilt_align) {
-			// don't start using EV data unless daa is arriving frequently
+			// don't start using EV data unless data is arriving frequently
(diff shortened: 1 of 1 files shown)


EKF/control.cpp
```diff
@@ -216,41 +216,14 @@ void Ekf::controlExternalVisionFusion()
...
 		if (!_control_status.flags.gps && (_params.fusion_mode & MASK_USE_EVYAW) && !_control_status.flags.ev_yaw && _control_status.flags.tilt_align) {
-			// don't start using EV data unless daa is arriving frequently
+			// don't start using EV data unless data is arriving frequently
 			if (isRecent(_time_last_ext_vision, 2 * EV_MAX_INTERVAL)) {
-				// reset the yaw angle to the value from the observation quaternion
-				// get the roll, pitch, yaw estimates from the quaternion states
-				Eulerf euler_init(_state.quat_nominal);
+				// reset the yaw angle to the value from the vision quaternion
+				const Eulerf euler_obs(_ev_sample_delayed.quat);
+				const float& yaw = euler_obs(2);
```
bresch, line 223:
```suggestion
				const float yaw = euler_obs(2);
```

Issue #786, opened 2020-03-24
EKF: GPS altitude not fused if GPS checks fail and GPS is primary height source


CarlOlsson:
Two replayed logfiles.
At t=600 baro data is stopped and at t=710 the ekf internal GPS checks fail.

Top is `EKF2_HGT_MODE` 1 (GPS as primary height source) and bottom is 0 (baro as primary height source). As can be seen for the `EKF2_HGT_MODE` 0 case the EKF switches to using GPS as altitude source and keeps fusing the GPS data even after t=710. This is not true for the `EKF2_HGT_MODE` 1 case
![image](https://user-images.githubusercontent.com/7795133/77447625-2a734f00-6df0-11ea-878a-ca55d233763e.png)

The bug was introduced [here](https://github.com/PX4/ecl/pull/614) and discovered [here](https://github.com/PX4/Firmware/issues/14459)

PR #813, opened 2020-05-11
Fix GPS altitude not fused if GPS checks fail and GPS is primary height source

priseborough:
See https://github.com/PX4/ecl/issues/786 for description of problem

This fixes the failure case where if GPS quality resulted in GPS pre-flight checks failing, no height sensor was used. The intended behaviour is now:

1) Once GPS height use commences, it will continue to be use provided 3D fix data is available. 

2) If range finder is being used as part of the user selectable 'range aid' functionality (uses range finder when low and slow), then switch back to using GPS height does require quality checks to pass. This is because GPS glitches combined with use of GPS height at low altitude is undesirable. If GPS quality checks do not pass then baro will be used as a backup. Using baro as a backup may be something that is undesirable for some vehicles.

3) If using baro and GPS checks pass, then switch back to GPs height will occur.

SITL testing with EKF2_HGT_MODE = 1 was performed by reducing EKF check thresholds to force the check to fail. No loss of height fusion was encountered. Further testing of the switching between GPSand  range finder and fallback to baro including the 'range aid' functionality is pending.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1003,16 +1003,16 @@ void Ekf::controlHeightFusion()
 
 	case VDIST_SENSOR_GPS:
 
-		// Do switching between GPS and rangefinder if using range finder as
-		// a height source when close to ground and moving slowly
+		// NOTE: emergency fallback due to extended loss of currently selected sensor data or failure
+		// to pass innovation cinsistency checks is handled elsewhere in Ekf::controlHeightSensorTimeouts.
+		// Do switching between GPS and rangefinder if using range finder as a height source when close
+		// to ground and moving slowly. Also handle switch back from emergency Baro sensor when GPS recovers.
 		if (!_control_status_prev.flags.rng_hgt && do_range_aid && _range_sensor.isDataHealthy()) {
 			setControlRangeHeight();
 
 			// we have just switched to using range finder, calculate height sensor offset such that current
 			// measurement matches our current height estimate
-			if (_control_status_prev.flags.rng_hgt != _control_status.flags.rng_hgt) {
-				_hgt_sensor_offset = _terrain_vpos;
-			}
+			_hgt_sensor_offset = _terrain_vpos;
 
 		} else if (_control_status_prev.flags.rng_hgt && !do_range_aid) {
 			// must stop using range finder so find another sensor now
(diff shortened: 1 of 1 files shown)


EKF/control.cpp
```diff
@@ -1003,26 +1003,37 @@ void Ekf::controlHeightFusion()
...
-		if (do_range_aid && _range_sensor.isDataHealthy()) {
+		// Do switching between GPS and rangefinder if using range finder as
+		// a height source when close to ground and moving slowly
+		if (!_control_status_prev.flags.rng_hgt && do_range_aid && _range_sensor.isDataHealthy()) {
 			setControlRangeHeight();
-			fuse_height = true;
 
 			// we have just switched to using range finder, calculate height sensor offset such that current
 			// measurement matches our current height estimate
 			if (_control_status_prev.flags.rng_hgt != _control_status.flags.rng_hgt) {
```
bresch, line 1013: This check can be removed because we only enter that part of the code if `!_control_status_prev.flags.rng_hgt` and we set `_control_status.flags.rng_hgt` just above (in `setControlRangeHeight();`)

EKF/control.cpp
```diff
@@ -1003,26 +1003,37 @@ void Ekf::controlHeightFusion()
...
 				_hgt_sensor_offset = _terrain_vpos;
 			}
 
-		} else if (!do_range_aid && _gps_data_ready && !_gps_hgt_intermittent && _gps_checks_passed) {
-			fuse_height = true;
+		} else if (_control_status_prev.flags.rng_hgt && !do_range_aid) {
```
bresch, line 1017: If we're not using RNG aid, we'll never enter that if and can't fallback to baro. How does that solve the issue?

EKF/control.cpp
priseborough: Will fix

EKF/control.cpp
priseborough: This function is there to select a height sensor and handle switching associated with normal operation.  Fallbacks due to a sensor fusion or data timeout are handled elsewhere in Ekf::controlHeightSensorTimeouts. I will add comment to make this clear.

priseborough:
The logic is harder to follow than it should be due to being in two places - see https://github.com/PX4/ecl/issues/814

PR #815, opened 2020-05-12
Reduce stored strings, to save flash space

kamilritz:
This is reducing the flash size by roughly 470 bytes.
The draw back is that some of those messages are not printed with a timestamp to the "SITL" terminal.

Firmware build tests are failing since ECL master is not compatible with Firmware master at the moment

### DIFF
--- a/EKF/airspeed_fusion.cpp
+++ b/EKF/airspeed_fusion.cpp
@@ -94,16 +94,18 @@ void Ekf::fuseAirspeed()
 			_fault_status.flags.bad_airspeed = true;
 
 			// if we are getting aiding from other sources, warn and reset the wind states and covariances only
+			const char* action_string = nullptr;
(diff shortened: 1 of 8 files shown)


dagar:
We probably shouldn't have any of these strings to be honest. In addition to them wasting a quite a lot of code space, printing status messages in relatively high rate important loops can be pretty terrible for real time performance.

For any of these things that matter they should be easily decipherable from logger status fields/flags.

As a short term crutch we could also have debug messages (enabled in SITL and replay), but really we should get in the habit of debugging/reviewing with the same mechanisms available with real flight data.

kamilritz:
I agree, a lot of the strings were probably never displayed. I just need some fast flash reduction for fitting the vision velocity in body frame PR on fmu-v2. 
@bresch If these changes are okay, lets get this in.

PR #816, opened 2020-05-14
ekf: disable xy accel bias learning before takeoff

bresch:
As those biases are usually poorly observable before takeoff because
they are almost perpendicular to the gravity vector, learning is often
driven by noise and numerical issues. This results in incorrect bias
learning before takeoff when the drone is static on ground for a long
period of time.

Testing on Pixhawk4 by changing the "in_air" check by "at_rest". XY bias learning is correctly inhibited and the covariances are forced to 0
2020-05-13_16-44-08_01_plot

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -163,8 +163,9 @@ void Ekf::predictCovariance()
 	for (unsigned stateIndex = 13; stateIndex <= 15; stateIndex++) {
 		const int index = stateIndex - 13;
 
-		// When on ground, only consider an accel bias observable if aligned with the gravity vector
-		const bool is_bias_observable = (fabsf(_R_to_earth(2, index)) > 0.8f) || _control_status.flags.in_air;
+		// TODO: When on ground, only consider an accel bias observable if aligned with the gravity vector
+		// const bool is_bias_observable = (fabsf(_R_to_earth(2, index)) > 0.8f) || _control_status.flags.in_air;
+		const bool is_bias_observable = true;
 		const bool do_inhibit_axis = do_inhibit_all_axes || !is_bias_observable;
 
 		if (do_inhibit_axis) {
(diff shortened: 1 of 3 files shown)


CarlOlsson:
This implementation seems to assume that body XY is orthogonal to the gravity vector. What will happen if this is not the case, e.g. the user leaves the vehicle powered on at 90 degrees pitch before taking off?

bresch:
>This implementation seems to assume that body XY is orthogonal to the gravity vector. What will happen if this is not the case, e.g. the user leaves the vehicle powered on at 90 degrees pitch before taking off?

That's correct. I think I have to split the bias learning even further and toggle the inhibition based on some observability criteria (e.g.: angle between gravity vector and each axis)

CarlOlsson:
yeah that is probably the way to go. Otherwise we risk having everything working fine until suddenly it doesn't

bresch:
@CarlOlsson Ok, it should be easy using the DCM as a criteria to enable each component independently.

dagar:
Note that ecl/EKF in air status is set from PX4/Firmware's land detector landed flag which is forced to true when disarmed.

bresch:
@CarlOlsson Now it would work for all orientations: I did set an arbitrary value of 0.8 against the cos of the axis to the gravity vector as an observability criteria.

Interestingly enough, the unit test now fails for the initializeWithTilt case, because all accel biases are inhibited (note that I have the same failure if I enable `MASK_INHIBIT_ACC_BIAS` in master, so it's not a bug introduced by this PR)

bresch:
>Note that ecl/EKF in air status is set from PX4/Firmware's land detector landed flag which is forced to true when disarmed.

@dagar That's all good. I just want to have this feature working before takeoff

bresch:
I'll bring the non-functional change that splits the accel bias learning first. This will make the merge process easier

bresch:
Non-functional change in https://github.com/PX4/ecl/pull/817

PR #817, opened 2020-05-14
ekf: split accel bias learning in independant xyz components

bresch:
This is a non-functional change required to select accel bias estimation per axis selection. The intent is then to disable the learning before takeoff of the components that are poorly observable.

### DIFF
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -153,15 +153,15 @@ void Ekf::predictCovariance()
 	_accel_magnitude_filt = fmaxf(dt_inv * _imu_sample_delayed.delta_vel.norm(), beta * _accel_magnitude_filt);
 	_accel_vec_filt = alpha * dt_inv * _imu_sample_delayed.delta_vel + beta * _accel_vec_filt;
 
-	const bool is_maneuvre_level_high = _ang_rate_magnitude_filt > _params.acc_bias_learn_gyr_lim
+	const bool is_manoeuvre_level_high = _ang_rate_magnitude_filt > _params.acc_bias_learn_gyr_lim
 					    || _accel_magnitude_filt > _params.acc_bias_learn_acc_lim;
 
 	const bool do_inhibit_all_axes = (_params.fusion_mode & MASK_INHIBIT_ACC_BIAS)
-				   	 || is_maneuvre_level_high
+					 || is_manoeuvre_level_high
 					 || _bad_vert_accel_detected;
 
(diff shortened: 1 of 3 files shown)


EKF/covariance.cpp
```diff
@@ -146,37 +146,42 @@ void Ekf::predictCovariance()
...
+		const int index = stateIndex - 13;
 
-	} else {
-		if (_accel_bias_inhibit) {
-			// reinstate the bias state variances
-			P(13,13) = _prev_dvel_bias_var(0);
-			P(14,14) = _prev_dvel_bias_var(1);
-			P(15,15) = _prev_dvel_bias_var(2);
+		// TODO: When on ground, only consider an accel bias observable if aligned with the gravity vector
+		// const bool is_bias_observable = (fabsf(_R_to_earth(2, index)) > 0.8f) || _control_status.flags.in_air;
```
bresch, line 167: This will be the functional change that will disable accel bias learning of the horizontal axes on ground.

EKF/covariance.cpp
```diff
@@ -146,37 +146,42 @@ void Ekf::predictCovariance()
...
+		const bool is_bias_observable = true;
+		const bool do_inhibit_axis = do_inhibit_all_axes || !is_bias_observable;
 
-		} else {
+		if (do_inhibit_axis) {
 			// store the bias state variances to be reinstated later
-			_prev_dvel_bias_var = P.slice<3,3>(13,13).diag();
```
kamilritz, line 175: This case is missing now and I can't figure out if it matters.

RomanBapst: @kamilritz I think it's actually nicer the way it's done now, because you only save the previous variances once you start inhibiting. Previously from what I understand it was updates each cycle when biases were not inhibited.

bresch: Yes, that's why I removed it during the refactoring. It's only used after passing through the other path that stores it as well.

EKF/covariance.cpp
RomanBapst: @bresch Why did you comment this out?

EKF/covariance.cpp
kamilritz: Agree

EKF/covariance.cpp
bresch: @RomanBapst To introduce the non-functional change before. It's easier to merge the whole refactoring if all the unit tests and the change indicator are passing. I'll do another PR to add this functional change

PR #818, opened 2020-05-15
ekf: disable xy accel bias learning before takeoff

bresch:
replaces https://github.com/PX4/ecl/pull/816
follows https://github.com/PX4/ecl/pull/817
In the log below, I oriented sequentially each axis with the gravity vector
DeepinScreenshot_select-area_20200515095438
Note that the accel bias learning is quite bad because the only sources of aiding are baro and fake hpos.


kamilritz:
CI tests are checking if the variances are big enough after some initialisation time. These values are set based on previous results. They can be changed without hesitation.

kamilritz:
Can you do a test flight on uneven ground? Then we could replay before and after and see the difference.

priseborough:
Should be sufficient tolerance given acos of 0.8 is ~36 deg.

kamilritz:
@bresch It would be beneficial to also add unit tests for this.

bresch:
@kamilritz Add the checks that the correct axes are not learned in the roll90 and pitch90 initialization? Would that be enough or do you have another check in mind?

kamilritz:
Yes, for all main axis pointing down, a check that only the right bias is learned

PR #820, opened 2020-05-16
EKF: fix initialization of local position validity 2

kamilritz:
Alternative solution to https://github.com/PX4/ecl/pull/819.

> The _deadreckon_time_exceeded flag is used in local_position_is_valid(). This means that
_params.valid_timeout_max after startup, in my observed case 5 seconds, the local position switches from valid to invalid and then after a while back to valid again. @julianoes 

Since the local position flag is also responsible for entering into position control mode, the initial value should be false in my opinion. Otherwise take off in position control could be possible even tough there is no observation constraining the state. 
Only after a first aiding event we should allow it to get valid until the aiding gets timedout eventually.

SITL log:

Before
Screenshot from 2020-05-16 09-42-01

After
Screenshot from 2020-05-16 09-42-27

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -1212,8 +1212,8 @@ void Ekf::update_deadreckoning_status()
 		_time_last_aiding = _time_last_imu - _params.no_aid_timeout_max;
 	}
 
-	// report if we have been deadreckoning for too long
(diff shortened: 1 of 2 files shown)


jkflying:
This will still work for position mode with only optical flow, right?

kamilritz:
> This will still work for position mode with only optical flow, right?

Yes

PR #821, opened 2020-05-16
Use resetQuatStateYaw during gps yaw reset

kamilritz:
This is probably not completely non functional, since there was a check whether the reset should happen  based on the quaternion error.
Now this check is based on the absolute yaw delta. The value of 15 degrees is still the same, not sure if this should be changed.

### DIFF
--- a/test/test_EKF_gps_yaw.cpp
+++ b/test/test_EKF_gps_yaw.cpp
@@ -1,0 +1,113 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2020 ECL Development Team. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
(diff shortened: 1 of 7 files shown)


priseborough:
To be safe we can do a SITL test with a simulated yaw change from the GPS using a yaw offset applied in ekf2_main.cpp where a negative delta of 10 deg is applied below 5m and +10 deg is applied above. When the vehicle climbs above 5m, the yaw would reset.

Not an ideal way to test, but would be sufficient.

kamilritz:
I will add an unit test for this.

kamilritz:
Unit tests are added.

test/test_EKF_gps_yaw.cpp
```diff
@@ -0,0 +1,113 @@
...
+	const int initial_quat_reset_counter = _ekf_wrapper.getQuaternionResetCounter();
+	_sensor_simulator.runSeconds(0.2);
+
+	// THEN: GPS heading fusion should have started;
+	EXPECT_TRUE(_ekf_wrapper.isIntendingGpsHeadingFusion());
+
+	// AND: no reset to GPS heading should be performed
+	EXPECT_EQ(_ekf_wrapper.getQuaternionResetCounter(), initial_quat_reset_counter);
+
+	// TODO: investigate why heading is not converging exactly to GPS heading
```
bresch, line 112: @kamilritz We should then open an issue for this

PR #822, opened 2020-05-16
Refactor position resets

kamilritz:


EKF/ekf.h
```diff
@@ -588,8 +588,17 @@ class Ekf : public EstimatorInterface
 
 	inline void resetVerticalVelocityTo(float new_vert_vel);
 
+	// reset position states of the ekf (only horizontal position)
```
bresch, line 591: I would even rename it `resetHorizontalPosition` and get rid of the comment :D

kamilritz:
I renamed `resetPosition` to `resetHorizontalPosition`

PR #831, opened 2020-05-28
EKF: Reduce EKF-GSF vulnerability to large yaw gyro bias errors

priseborough:
The EKF-GSF yaw estimator uses a bank of AHRS filters that are unable to estimate rate gyro bias about the gravity vector. This can result in poor yaw estimates when large yaw gyro bias errors are present before takeoff. This is the effect of a 0.05 rad/sec yaw rate bias applied in SITL:

Screen Shot 2020-05-28 at 3 22 34 pm

After application of the patch, yaw drift during hover is corrected:

Screen Shot 2020-05-28 at 4 14 27 pm

### DIFF
--- a/EKF/EKFGSF_yaw.cpp
+++ b/EKF/EKFGSF_yaw.cpp
@@ -14,8 +14,9 @@ EKFGSF_yaw::EKFGSF_yaw()
 }
 
 void EKFGSF_yaw::update(const imuSample& imu_sample,
-                	bool run_EKF,           // set to true when flying or movement is suitable for yaw estimation
-                	float airspeed)   	// true airspeed used for centripetal accel compensation - set to 0 when not required.
+			bool run_EKF,			// set to true when flying or movement is suitable for yaw estimation
+			float airspeed,			// true airspeed used for centripetal accel compensation - set to 0 when not required.
+			const Vector3f &imu_gyro_bias)  // estimated rate gyro bias (rad/sec)
 {
 	// copy to class variables
 	_delta_ang = imu_sample.delta_ang;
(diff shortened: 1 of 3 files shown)


EKF/ekf_helper.cpp
```diff
@@ -1858,7 +1858,8 @@ void Ekf::runYawEKFGSF()
 		TAS = _airspeed_sample_delayed.true_airspeed;
 	}
 
-	yawEstimator.update(_imu_sample_delayed, _control_status.flags.in_air, TAS);
+	Vector3f imu_gyro_bias = getGyroBias();
```
bresch, line 1861:
```suggestion
	const Vector3f imu_gyro_bias = getGyroBias();
```

EKF/EKFGSF_yaw.cpp
```diff
@@ -60,6 +61,11 @@ void EKFGSF_yaw::update(const imuSample& imu_sample,
 		if (!_ekf_gsf_vel_fuse_started) {
 			initialiseEKFGSF();
 			ahrsAlignYaw();
+			// Initialise to gyro bias estimate from main filter because there could be a large
+			// uncorrected rate gyro bias error about the grabity vector
```
RomanBapst, line 65:
```suggestion
			// uncorrected rate gyro bias error about the gravity vector
```

RomanBapst:
@priseborough Changes look good. Is there any chance that during initialization the bias is not learned sufficiently well? Would it be initialized a second time after that?

priseborough:
The reset event is on takeoff and after that it relies on the movement of the vehicle and gravity vector to continue updating the bias. Using the bias too long after takeoff runs the risk of a bad yaw and loss of nav causing the EKF to have a bad gyro bias estimate.

Issue #835, opened 2020-06-02
EKF vibration induced loss of height protection when using GPS instead of Baro height


priseborough:
The detection of loss of height accuracy due to bad vibration used to more quickly reset the states to the primary height and vertical velocity sensor currently only works when Baro is used for height and GPS is used for vertical velocity. With the availability of better clipping data this should be extended to work for a bigger range of height sensors, particularly the use of GPS for height which is becoming more common.

francelico:
Hi @priseborough , do you think this could have been the cause of the behavior observed in #834 ? I was running the PX4 v1.10.1 ecl version at the time (https://github.com/PX4/ecl/tree/dec1ed0aeb395a885b60b1d61cfb57bd932c7d34). 

I was not using GPS for vertical velocity. Baro, Vision pose and Range finders were fused in the AID_MASK and EKF2_HGT_MODE was set to Baro

priseborough:
The use of clipping counter to improve robustness will have improved this behaviour, but the use of the same sensor for height and vertical velocity introduces a common mode failure that we cannot ignore.

PR #836, opened 2020-06-02
EKF: Use IMU clipping to adjudicate bad accel data check

priseborough:
Improvement in response to Issue #835 

Extends early reset to recover from bad vibration induced loss of height estimation to includes sensor combinations other than Baro for height and GPS for vertical velocity.

When height and vertical velocity sensing iis from the same sensor (GPS or EV), IMU clipping is used to adjudicate.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -898,8 +898,8 @@ void Ekf::checkVerticalAccelerationHealth()
 	// Don't use stale innovation data.
 	bool is_inertial_nav_falling = false;
 	bool are_vertical_pos_and_vel_independant = false;
-	if (_time_last_imu < _vert_pos_fuse_time_us + 1000000) {
-		if (_time_last_imu < _vert_vel_fuse_time_us + 1000000) {
+	if (isRecent(_vert_pos_fuse_time_us, 1000000)) {
+		if (isRecent(_vert_vel_fuse_time_us, 1000000)) {
 			// If vertical position and velocity come from independent sensors then we can
 			// trust them more if they disagree with the IMU, but need to check that they agree
 			const bool using_gps_for_both = _control_status.flags.gps_hgt && _control_status.flags.gps;
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -895,22 +895,45 @@ void Ekf::checkVerticalAccelerationHealth()
 	// innovations being positive and not stale.
 	// Clipping causes the average accel reading to move towards zero which makes the INS
 	// think it is falling and produces positive vertical innovations
+	// Don't use stale innovation data.
+	bool is_inertial_nav_falling = false;
+	bool are_vertical_pos_and_vel_independant = false;
+	if (_time_last_imu < _vert_pos_fuse_time_us + 1000000) {
+		if (_time_last_imu < _vert_vel_fuse_time_us + 1000000) {
+			// If vertical position and velocity come from independent sensors then we can
+			// trust them more if they disagree with the IMU, but need to check that they agree
+			const bool using_gps_for_both = _control_status.flags.gps_hgt && _control_status.flags.gps;
+			const bool using_ev_for_both = _control_status.flags.ev_hgt && _control_status.flags.ev_vel;
+			are_vertical_pos_and_vel_independant = !(using_gps_for_both || using_ev_for_both);
+			if (are_vertical_pos_and_vel_independant) {
+				if (_vert_pos_innov_ratio > 0.0f && _vert_vel_innov_ratio > 0.0f) {
+					is_inertial_nav_falling = _vert_pos_innov_ratio * _vert_vel_innov_ratio > 0.5f * sq(_params.vert_innov_test_lim);
+				}
+			} else {
+				is_inertial_nav_falling = _vert_vel_innov_ratio > _params.vert_innov_test_lim || _vert_pos_innov_ratio > _params.vert_innov_test_lim;
+			}
+		} else {
+			// only height sensing available
+			is_inertial_nav_falling = _vert_pos_innov_ratio > _params.vert_innov_test_lim;
+		}
+	}
 
-	const float var_product_lim = sq(_params.vert_innov_test_lim) * sq(_params.vert_innov_test_lim);
-	const bool is_fusing_gps_vel = !_gps_hgt_intermittent;
-	const bool is_fusing_baro_alt = _control_status.flags.baro_hgt && !_baro_hgt_faulty;
-	const bool are_vertical_pos_and_vel_independant = is_fusing_gps_vel && is_fusing_baro_alt; // TODO: should we add range hgt here?
-	const float pos_vel_innov_product = _gps_pos_innov(2) * fmaxf(fabsf(_gps_vel_innov(2)),fabsf(_ev_vel_innov(2)));
-	const float pos_vel_innov_var_product = _gps_pos_innov_var(2) * fmaxf(fabsf(_gps_vel_innov_var(2)),fabsf(_ev_vel_innov_var(2)));
-	const bool are_pos_vel_sensor_in_agreement = sq(pos_vel_innov_product) > var_product_lim * (pos_vel_innov_var_product);
+	// Check for > 50% clipping affected IMU samples within the past 1 second
+	const uint16_t clip_count_limit = 1000 / FILTER_UPDATE_PERIOD_MS;
+	if (_imu_sample_delayed.delta_vel_clipping[0] ||
+	    _imu_sample_delayed.delta_vel_clipping[1] ||
+	    _imu_sample_delayed.delta_vel_clipping[2]) {
+		if (_clip_counter < clip_count_limit) {
+			_clip_counter++;
+		}
+	} else if (_clip_counter > 0) {
+		_clip_counter--;
+	}
 
-	// A positive innovation indicates that the inertial nav thinks it is falling
-	// This is caused by average of the Z accelerometer being 0, due to clipping
-	const bool is_inertial_nav_falling = _gps_vel_innov(2) > 0.0f || _ev_vel_innov(2) > 0.0f;
-
-	const bool bad_vert_accel = are_vertical_pos_and_vel_independant &&
-			      are_pos_vel_sensor_in_agreement &&
-			      is_inertial_nav_falling;
+	// if vertical velocity and position are independent and agree, then do not require evidence of clipping if
+	// innovations are large
+	const bool bad_vert_accel = (are_vertical_pos_and_vel_independant || _clip_counter > clip_count_limit / 2) &&
+				is_inertial_nav_falling;
 
 	if (bad_vert_accel) {
 		_time_bad_vert_accel =  _time_last_imu;
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -893,8 +893,8 @@ void Ekf::checkVerticalAccelerationHealth()
 {
 	// Check for IMU accelerometer vibration induced clipping as evidenced by the vertical
 	// innovations being positive and not stale.
-	// Clipping causes the average accel reading to move towards zero which makes the INS
-	// think it is falling and produces positive vertical innovations
+	// Clipping usually causes the average accel reading to move towards zero which makes the INS
+	// think it is falling and produces positive vertical innovations.
 	// Don't use stale innovation data.
 	bool is_inertial_nav_falling = false;
 	bool are_vertical_pos_and_vel_independant = false;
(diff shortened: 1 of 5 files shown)


priseborough:
Flown twice (not  high vibe setup):

https://logs.px4.io/plot_app?log=13362bfa-eac2-4fa9-a066-ab0adbad5959
https://logs.px4.io/plot_app?log=e4b3b1f8-e1a8-400e-a90a-b36b48f0c59a

SITL testing of clipping will require addition of high frequency vibe to truth data , eg 200Hz

EKF/control.cpp
```diff
@@ -895,21 +895,34 @@ void Ekf::checkVerticalAccelerationHealth()
...
 	// This is caused by average of the Z accelerometer being 0, due to clipping
 	const bool is_inertial_nav_falling = _gps_vel_innov(2) > 0.0f || _ev_vel_innov(2) > 0.0f;
 
-	const bool bad_vert_accel = are_vertical_pos_and_vel_independant &&
-			      are_pos_vel_sensor_in_agreement &&
-			      is_inertial_nav_falling;
+	// If vertical position and velocity come from independent sensors then we can
+	// trust them more if they disagree with the IMU
+	const bool are_vertical_pos_and_vel_independant = !(_control_status.flags.gps_hgt && _control_status.flags.gps) &&
+							  !(_control_status.flags.ev_hgt && _control_status.flags.ev_vel);
```
bresch, line 910: Unfortunately, if there is only a baro or if GPS aiding is active but we detected bad data, this condition will be true. This is why we had the tests `is_fusing_baro_alt` and `is_fusing_gps_vel`

priseborough: It is getting messy and hard to follow since adding support for different combinations of velocity and height, so it needs to be abstracted a different way. Will rework.

EKF/ekf.h
```diff
@@ -385,6 +385,10 @@ class Ekf : public EstimatorInterface
 	Vector3f _delta_vel_bias_var_accum;		///< kahan summation algorithm accumulator for delta velocity bias variance
 	Vector3f _delta_angle_bias_var_accum;	///< kahan summation algorithm accumulator for delta angle bias variance
 
+	float _vert_pos_innov_ratio;		///< standard deviation of vertical position innovation
+	uint64_t _vert_pos_fuse_time_us;	///< last system time in usec vertical position measurement fuson was attempted
```
kamilritz, line 389: Does it have to be the time of the last fusion **attempt**? If so, it would be great if the variable name would include that. If not, there is already `_time_last_hgt_fuse`.

priseborough:
SITL testing showed the output observer vertical position derivative as it is currently calculated was also biased, but the good news was that the effect of the vertical velocity resets on the height controller are not as bad as first thought. I now need to do SITL testing without use of GPS vertical velocity.

EKF/control.cpp
```diff
@@ -884,24 +884,42 @@ void Ekf::checkVerticalAccelerationHealth()
...
+	const uint16_t clip_count_limit = 1000 / FILTER_UPDATE_PERIOD_MS;
+	if (_imu_sample_delayed.delta_vel_clipping[0] ||
+	    _imu_sample_delayed.delta_vel_clipping[1] ||
+	    _imu_sample_delayed.delta_vel_clipping[2]) {
+		if (_clip_counter < clip_count_limit) {
+			_clip_counter++;
+		}
+	} else if (_clip_counter > 0) {
+		_clip_counter--;
+	}
```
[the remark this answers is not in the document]
bresch: the 2nd if can be an `else if` since you'll never be able to enter both if at the same time

priseborough:
Replay of a recent log with bad clipping revealed an additional vulnerability which has been fixed via  https://github.com/PX4/ecl/pull/836/commits/4d999c2c57ac0b3083d7b40dbb04d1e12a99628e

Replay before changes:
replay-before-fix

Replay after changes:
replay-after-fix

@kamilritz I'll go through your review comments after Tuesdays estimation call.

EKF/control.cpp
```diff
@@ -884,24 +884,42 @@ void Ekf::checkVerticalAccelerationHealth()
...
-	const float pos_vel_innov_product = _gps_pos_innov(2) * fmaxf(fabsf(_gps_vel_innov(2)),fabsf(_ev_vel_innov(2)));
-	const float pos_vel_innov_var_product = _gps_pos_innov_var(2) * fmaxf(fabsf(_gps_vel_innov_var(2)),fabsf(_ev_vel_innov_var(2)));
-	const bool are_pos_vel_sensor_in_agreement = sq(pos_vel_innov_product) > var_product_lim * (pos_vel_innov_var_product);
+	// Clipping usually causes the average accel reading to move towards zero which makes the INS
+	// think it is falling and produces positive vertical innovations.
+	// Don't use stale innovation data.
+	bool is_inertial_nav_falling = false;
+	bool are_vertical_pos_and_vel_independant = false;
+	if (_time_last_imu < _vert_pos_fuse_time_us + 1000000) {
+		if (_time_last_imu < _vert_vel_fuse_time_us + 1000000) {
```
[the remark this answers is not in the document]
priseborough: Done

EKF/ekf.h
priseborough: Unfortunately _time_last_hgt_fuse will only be set when there has an actual fusion event, whereas this use case requires all fusion attempts, regardless of whether the fusion was successful.

priseborough:
Running the same replay with GPS fusion disabled shows that the early detection and reset did not activate. Will investigate.

Screen Shot 2020-07-07 at 6 32 34 pm

priseborough:
The behaviour with no indepenedant vertical source is not degraded compared to before. I will tackle it in a separate PR.

EKF/control.cpp
```diff
@@ -891,24 +891,42 @@ void Ekf::checkVerticalAccelerationHealth()
...
-	// A positive innovation indicates that the inertial nav thinks it is falling
-	// This is caused by average of the Z accelerometer being 0, due to clipping
-	const bool is_inertial_nav_falling = _gps_vel_innov(2) > 0.0f || _ev_vel_innov(2) > 0.0f;
+	// Check for more than 50% clipping affected IMU samples within the past 1 second
+	const uint16_t clip_count_limit = 1000 / FILTER_UPDATE_PERIOD_MS;
+	const bool is_clipping = _imu_sample_delayed.delta_vel_clipping[0] ||
+				 _imu_sample_delayed.delta_vel_clipping[1] ||
+				 _imu_sample_delayed.delta_vel_clipping[2];
+	if (is_clipping &&_clip_counter < clip_count_limit) {
+			_clip_counter++;
```
bresch, line 920:
```suggestion
		_clip_counter++;
```

EKF/ekf.h
kamilritz: What about `_vert_pos_fuse_attempt_time_us `?

EKF/vel_pos_fusion.cpp
```diff
@@ -132,14 +144,25 @@ bool Ekf::fuseVerticalPosition(const Vector3f &innov, const Vector2f &innov_gate
...
+	bool innov_check_pass = test_ratio(1) <= 1.0f;
+
+	// if there is bad vertical acceleration data, then don't reject measurement,
+	// but limit innovation to prevent spikes that could destabilise the filter
+	float innovation = innov(2);
+	if (_bad_vert_accel_detected && !innov_check_pass) {
+		const float innov_limit = innov_gate(1) * sqrtf(innov_var(2));
+		innovation = math::constrain(innov(2), -innov_limit, innov_limit);
+		innov_check_pass = true;
+	} else {
```
kamilritz, line 158: The else case can be left out

EKF/vel_pos_fusion.cpp
```diff
@@ -132,14 +144,25 @@ bool Ekf::fuseVerticalPosition(const Vector3f &innov, const Vector2f &innov_gate
...
+	_vert_pos_innov_ratio = innov(2) / sqrtf(innov_var(2));
+	_vert_pos_fuse_time_us = _time_last_imu;
+	bool innov_check_pass = test_ratio(1) <= 1.0f;
+
+	// if there is bad vertical acceleration data, then don't reject measurement,
+	// but limit innovation to prevent spikes that could destabilise the filter
+	float innovation = innov(2);
+	if (_bad_vert_accel_detected && !innov_check_pass) {
+		const float innov_limit = innov_gate(1) * sqrtf(innov_var(2));
+		innovation = math::constrain(innov(2), -innov_limit, innov_limit);
```
kamilritz, line 156: Do we log the detection of bad acceleration? Just wondering if we should log the constrained or unconstrained innovation.

EKF/ekf.h
```diff
@@ -387,6 +387,10 @@ class Ekf : public EstimatorInterface
 	Vector3f _delta_vel_bias_var_accum;		///< kahan summation algorithm accumulator for delta velocity bias variance
 	Vector3f _delta_angle_bias_var_accum;	///< kahan summation algorithm accumulator for delta angle bias variance
 
+	float _vert_pos_innov_ratio;		///< standard deviation of vertical position innovation
```
kamilritz, line 390:
```suggestion
	float _vert_pos_innov_ratio;		///<vertical position: innovation divided by standard deviation of  innovation
```

EKF/control.cpp
priseborough: See https://github.com/PX4/ecl/pull/836/commits/29f21aa036eb79ba21446099978f91b76021093d

EKF/vel_pos_fusion.cpp
priseborough: See https://github.com/PX4/ecl/pull/836/commits/29f21aa036eb79ba21446099978f91b76021093d

EKF/vel_pos_fusion.cpp
priseborough: See soln_status.flags.accel_error

priseborough:
rebased to check for compatibility with CI tests

priseborough:
@bresch - are you OK with this now?

dagar:
Other than a rebase what's needed to get this in?

EKF/vel_pos_fusion.cpp
```diff
@@ -77,14 +77,26 @@ bool Ekf::fuseVerticalVelocity(const Vector3f &innov, const Vector2f &innov_gate
...
+		innov_check_pass = true;
+	} else {
+		innovation = innov(2);
```
bresch, line 92:
```suggestion
```
`innovation` is already initialized to `innov(2)` above

EKF/vel_pos_fusion.cpp
```diff
@@ -132,14 +144,25 @@ bool Ekf::fuseVerticalPosition(const Vector3f &innov, const Vector2f &innov_gate
...
 	test_ratio(1) = sq(innov(2)) / (sq(innov_gate(1)) * innov_var(2));
-
-	const bool innov_check_pass = test_ratio(1) <= 1.0f;
+	_vert_pos_innov_ratio = innov(2) / sqrtf(innov_var(2));
+	_vert_pos_fuse_attempt_time_us = _time_last_imu;
+	bool innov_check_pass = test_ratio(1) <= 1.0f;
+
+	// if there is bad vertical acceleration data, then don't reject measurement,
+	// but limit innovation to prevent spikes that could destabilise the filter
+	float innovation;
```
bresch, line 153:
```suggestion
	float innovation = innov(2);
```

EKF/vel_pos_fusion.cpp
```diff
@@ -132,14 +144,25 @@ bool Ekf::fuseVerticalPosition(const Vector3f &innov, const Vector2f &innov_gate
...
+	if (_bad_vert_accel_detected && !innov_check_pass) {
+		const float innov_limit = innov_gate(1) * sqrtf(innov_var(2));
+		innovation = math::constrain(innov(2), -innov_limit, innov_limit);
+		innov_check_pass = true;
+	} else {
+		innovation = innov(2);
```
bresch, line 159:
```suggestion
```

bresch (changes requested):
Just a minor non-functional change, then rebase and it's all good for me

ludost:
In the incident flight reported at: https://github.com/PX4/PX4-Autopilot/issues/16490 I noticed that this "bad accel data check" didn't trigger, even though it was specifically meant to help in that situation. I traced it to the usage of GPS_innov(2) which was always 0 in my case. These changes seem to fix that issue, so I guess that that flight can be used as an extra testcase for these changes.

EKF/vel_pos_fusion.cpp
priseborough: Fixed

EKF/vel_pos_fusion.cpp
priseborough: Fixed - I removed the initialisation to innov(2)

EKF/vel_pos_fusion.cpp
priseborough: Fixed

priseborough:
Rebased to resolve conficts

kamilritz (approved):

bresch (approved):

PR #837, opened 2020-06-03
Terrain fusion aid selection fix

bresch:
I tried to split the fixes properly in different commits.

Summary of the fixes:
- possibility to choose the aiding sources of the terrain estimator
- ~get feedback in the log from the aiding sources currently used by the terrain estimator (will require Firmware update to actually log the data)~ edit: moved to #839
- ~fix sensor simulator that was updating the ekf faster than IMU data (this created instability of the filter in dynamic unit tests) FYI @kamilritz~ edit: moved to #838
- add unit tests for flow and range fusions in the terrain estimator

~Due to the `sensor_simulator` update rate fix, a lot of unit tests are now failing (because EKF prediction step runs slower) and I'm currently fixing them.~ edit: fixed in #838

### DIFF
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -179,7 +179,7 @@ void Ekf::fuseFlowForTerrain()
 	// calculate optical LOS rates using optical flow rates that have had the body angular rate contribution removed
 	// correct for gyro bias errors in the data used to do the motion compensation
 	// Note the sign convention used: A positive LOS rate is a RH rotation of the scene about that axis.
-	const Vector2f opt_flow_rate = Vector2f{_flow_compensated_XY_rad} / _flow_sample_delayed.dt + Vector2f{_flow_gyro_bias};
+	const Vector2f opt_flow_rate = Vector2f(_flow_compensated_XY_rad) / _flow_sample_delayed.dt + Vector2f(_flow_gyro_bias);
 
 	// get latest estimated orientation
 	const float q0 = _state.quat_nominal(0);
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -55,7 +55,7 @@ bool Ekf::initHagl()
 		_time_last_fake_hagl_fuse = _time_last_imu;
 		initialized = true;
 
-	} else if ((_params.terrain_fusion_mode & TerrainFusionMask::TerrainFuseRangeFinder)
+	} else if (shouldUseRangeFinderForHagl()
 		   && _range_sensor.isDataHealthy()) {
 		// if we have a fresh measurement, use it to initialise the terrain estimator
 		_terrain_vpos = _state.pos(2) + _range_sensor.getDistBottom();
@@ -64,7 +64,7 @@ bool Ekf::initHagl()
 		// success
 		initialized = true;
 
-	} else if ((_params.terrain_fusion_mode & TerrainFusionMask::TerrainFuseOpticalFlow)
+	} else if (shouldUseOpticalFlowForHagl()
 		   && _flow_for_terrain_data_ready) {
 		// initialise terrain vertical position to origin as this is the best guess we have
 		_terrain_vpos = fmaxf(0.0f,  _state.pos(2));
@@ -81,6 +81,16 @@ bool Ekf::initHagl()
 	}
 
 	return initialized;
+}
+
+bool Ekf::shouldUseRangeFinderForHagl() const
+{
+	return (_params.terrain_fusion_mode & TerrainFusionMask::TerrainFuseRangeFinder);
+}
+
+bool Ekf::shouldUseOpticalFlowForHagl() const
+{
+	return (_params.terrain_fusion_mode & TerrainFusionMask::TerrainFuseOpticalFlow);
 }
 
 void Ekf::runTerrainEstimator()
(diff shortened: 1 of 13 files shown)


bresch:
I fixed all the unit tests and placed the modifications in the commit that breaks them: 99f2e33

bresch:
This PR started to have too many changes so I pulled the sensor_simulator bugfix in https://github.com/PX4/ecl/pull/838

bresch:
Extracted the aid bitmask feedback to https://github.com/PX4/ecl/pull/839

EKF/terrain_estimator.cpp
```diff
@@ -281,6 +283,8 @@ bool Ekf::isTerrainEstimateValid() const
 
 void Ekf::updateTerrainValidity()
 {
+	const bool rng_selected = _params.terrain_fusion_mode & TerrainFusionMask::TerrainFuseRangeFinder;
```
kamilritz, line 286: Can we put `params.terrain_fusion_mode & TerrainFusionMask::TerrainFuseRangeFinder` in a function and use it in all location. Name suggestion `shouldUseRangeFinderForHagl` and `shouldUseOpticalFlowForHagl`

test/sensor_simulator/flow.cpp
```diff
@@ -15,7 +15,7 @@ Flow::~Flow()
 
 void Flow::send(uint64_t time)
 {
-	_flow_data.dt = time - _time_last_data_sent;
+	_flow_data.dt = 1e-6 * (time - _time_last_data_sent); // in seconds
```
kamilritz, line 18:
```suggestion
	_flow_data.dt = 1e-6f * (time - _time_last_data_sent); // in seconds
```

test/sensor_simulator/gps.cpp
```diff
@@ -15,7 +15,17 @@ Gps::~Gps()
 
 void Gps::send(uint64_t time)
 {
+	const float dt = float((time - _gps_data.time_usec) * 1e-6);
```
kamilritz, line 18:
```suggestion
	const float dt = float((time - _gps_data.time_usec) * 1e-6f);
```

kamilritz, line 18:
```suggestion
	const float dt = static_cast<float>((time - _gps_data.time_usec) * 1e-6);
```

kamilritz: What is your view on starting to use static_cast<>?

test/sensor_simulator/gps.h
```diff
@@ -53,6 +53,7 @@ class Gps: public Sensor
 	void setData(const gps_message& gps);
 	void stepHeightByMeters(float hgt_change);
 	void stepHorizontalPositionByMeters(Vector2f hpos_change);
+	void setPositionRate(Vector3f rate) { _gps_pos_rate = rate; }
```
kamilritz, line 56:
```suggestion
	void setPositionRate(const Vector3f& rate) { _gps_pos_rate = rate; }
```

test/sensor_simulator/gps.cpp
```diff
@@ -15,7 +15,17 @@ Gps::~Gps()
...
+
 	_gps_data.time_usec = time;
+
+	if (fabsf(_gps_pos_rate(0)) > FLT_EPSILON || fabsf(_gps_pos_rate(1)) > FLT_EPSILON) {
+		stepHorizontalPositionByMeters(Vector2f(_gps_pos_rate) * dt);
+	}
+	if (fabsf(_gps_pos_rate(2)) > FLT_EPSILON) {
+		stepHeightByMeters(-_gps_pos_rate(2) * dt);
```
kamilritz, line 26: Is the minus really needed? Can you specify in which frame the pos rate is!

test/test_EKF_flow.cpp
```diff
@@ -89,8 +89,8 @@ TEST_F(EkfFlowTest, resetToFlowVelocityInAir)
 	const Vector2f simulated_horz_velocity(0.5f, -0.2f);
 	flowSample flow_sample = _sensor_simulator._flow.dataAtRest();
 	flow_sample.flow_xy_rad =
-		Vector2f(- simulated_horz_velocity(1) * flow_sample.dt / estimated_distance_to_ground,
-			   simulated_horz_velocity(0) * flow_sample.dt / estimated_distance_to_ground);
+		Vector2f( simulated_horz_velocity(1) * flow_sample.dt / estimated_distance_to_ground,
```
kamilritz, line 92: Do you know why this was passing the test before even tough I got the signs wrong.

test/test_EKF_terrain_estimator.cpp
```diff
@@ -0,0 +1,176 @@
...
+	const float flow_height = 5.f;
+	runFlowAndRngScenario(rng_height, flow_height);
+
+	// THEN: the estimator should use flow for terrain and the estimated terrain height
+	// should converge to the simulated height
+	EXPECT_FALSE(_ekf_wrapper.isIntendingTerrainRngFusion());
+	EXPECT_TRUE(_ekf_wrapper.isIntendingTerrainFlowFusion());
+
+	const float estimated_distance_to_ground = _ekf->getTerrainVertPos();
+	EXPECT_NEAR(estimated_distance_to_ground, flow_height, 0.5f);
```
kamilritz, line 155: Do you have an idea why the error is here higher?

test/test_EKF_flow.cpp
bresch: Yes, because the test was just for the reset counter. There is still a bug in the flow reset that always initializes with zero. The value isn't important.

test/test_EKF_terrain_estimator.cpp
bresch: I think this is because the range finder fusion is stronger than the optical flow one. If we let it run more, it will converge to the true value.

bresch:
@kamilritz Thanks for the review, I'll make the required modifications

test/sensor_simulator/gps.cpp
bresch: @kamilritz I'm for it because it's safer.

test/sensor_simulator/gps.cpp
bresch: In NED, I added that to the function name

bresch:
@kamilritz Updated, please could you review again?

PR #838, opened 2020-06-03
sensor_simulator: update EKF at IMU rate

bresch:
This is how it is also done in ekf2_main. Otherwise, this leads to multiple integration of the same IMU data due to asynchronous sensor updates triggering a prediction step between IMU updates.

- Fix unit tests that broke because of this fix
- Set IMU rate to 200Hz (and update the csv for change indication)

I added a print in the code to find the exact tilt alignment time and is: 6.1s. I then added a bit of margin in the unit test and used a value of 7.
I tested on a Pixhawk4 and measured approximately 8s with a stopwatch.


test/sensor_simulator/sensor_simulator.cpp
```diff
@@ -146,9 +146,13 @@ void SensorSimulator::runMicroseconds(uint32_t duration)
...
 	for(;_time < start_time + (uint64_t)duration; _time+=1000)
 	{
+		bool update_imu = _imu.should_send(_time);
 		updateSensors();
 
-		_ekf->update();
+		if (update_imu) {
+			// Update at IMU rate
+			_ekf->update();
+		}
```
bresch, line 155: This is the fix

test/test_EKF_basics.cpp
```diff
@@ -164,27 +164,32 @@ TEST_F(EkfBasicsTest, gpsFusion)
...
+	_sensor_simulator.setImuBias(accel_bias_sim, Vector3f(0.0f,0.0f,0.0f));
+	_ekf->set_min_required_gps_health_time(1e6);
+	_sensor_simulator.runSeconds(30);
 
 	const Vector3f pos = _ekf->getPosition();
 	const Vector3f vel = _ekf->getVelocity();
 	const Vector3f accel_bias = _ekf->getAccelBias();
 	const Vector3f gyro_bias = _ekf->getGyroBias();
-	const Vector3f zero{0.0f, 0.0f, 0.0f};
+	const Vector3f zero = {0.0f, 0.0f, 0.0f};
```
kamilritz, line 182:
```suggestion
	const Vector3f zero(0.0f, 0.0f, 0.0f);
```

PR #839, opened 2020-06-03
Terrain estimate: Added height above ground source bitmask

bresch:
A PR in Firmware will follow this to actually log the bitmask

### DIFF
--- a/EKF/terrain_estimator.cpp
+++ b/EKF/terrain_estimator.cpp
@@ -52,6 +52,7 @@ bool Ekf::initHagl()
 		_terrain_vpos = _state.pos(2) + _params.rng_gnd_clearance;
 		// use the ground clearance value as our uncertainty
 		_terrain_var = sq(_params.rng_gnd_clearance);
+		_time_last_fake_hagl_fuse = _time_last_imu;
 		initialized = true;
 
 	} else if ((_params.terrain_fusion_mode & TerrainFusionMask::TerrainFuseRangeFinder)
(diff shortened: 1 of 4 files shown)


EKF/terrain_estimator.cpp
```diff
@@ -289,6 +290,9 @@ void Ekf::updateTerrainValidity()
 						    && !_control_status.flags.opt_flow;
 
 	_hagl_valid = (_terrain_initialised && (recent_range_fusion || recent_flow_for_terrain_fusion));
+
+	_hagl_sensor_status.flags.range_finder = recent_range_fusion && (_time_last_fake_hagl_fuse != _time_last_hagl_fuse);
```
kamilritz, line 294: Why is the second condition necessary?

RomanBapst: @kamilritz The idea of reporting this information is that the system can determine whether an actual range finder is in use or not. If the system is on the ground then the terrain estimate is valid but it does not tell us yet if a range finder is available. Distinguishing between fake range data (which really does not stem from a sensor but just from the assumption being on the ground) allows the system to understand when a range finder is in use.

RomanBapst: @kamilritz I would be ok to actually add a "fake_range_data" or "on_ground_assumption" flag if you think it's useful.

bresch:
I'll merge it as is and we can still patch it if needed.

EKF/terrain_estimator.cpp
kamilritz: @RomanBapst That would be nice. The "on_ground_assumption" sounds very clean

PR #840, opened 2020-06-04
Fix GPS reference altitude in case gps checks pass quickly

RomanBapst:
In SITL the gps reference altitude was always set to zero as the gps checks passed before the main filtered initialized.
This PR also sets the gps velocity and position derivative filter states to zero if they are not being updated (vehicle is in air or there is movement.)

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -893,4 +893,5 @@ private:
 	// Returns true if the reset was successful
 	bool resetYawToEKFGSF();
 
+	void resetGpsDriftCheckFilters();
 };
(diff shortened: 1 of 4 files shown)


EKF/ekf.h
```diff
@@ -890,4 +890,5 @@ class Ekf : public EstimatorInterface
 	// Returns true if the reset was successful
 	bool resetYawToEKFGSF();
 
+	void resetGPSDriftCheckFilters();
```
kamilritz, line 893:
```suggestion
	void resetGpsDriftCheckFilters();
```

kamilritz: Should this also be called in the main reset function?

RomanBapst:
@kamilritz Thanks for the review, I updated the PR.

PR #842, opened 2020-06-09
AlphaFilter: prevent setParameters division by zero

MaEtUgR:
fixes https://github.com/PX4/Firmware/issues/15065

The denominator of `setParameters()` can get zero if `time_constant` and `sample_interval` both are exactly zero. In this case updating the filter coefficient needs to be prevented.

### DIFF
--- a/EKF/AlphaFilter.hpp
+++ b/EKF/AlphaFilter.hpp
@@ -42,6 +42,8 @@
 
 #pragma once
 
+#include "../ecl.h"
+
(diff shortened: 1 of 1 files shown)


julianoes:
@MaEtUgR shouldn't the actual division be prevented in the first place?

MaEtUgR:
JFYI this was the follow-up: https://github.com/PX4/ecl/pull/843

PR #843, opened 2020-06-11
Move AlphaFilter, replace denominator check

MaEtUgR:
As discussed in the PX4 dev call of yesterday (10.06.2020) I:
- explicitly check for a division by zero in the `setParameters` calculation. Note that I decided that it doesn't make sense to check again for finiteness in the setter of the alpha variable. Before I thought I would kill two birds with one stone but explicitly checking for the error case makes more sense. Thanks for following up @julianoes . Note 2: The parameters of `setParameters` are both timespans and hence positive.
- moved the AlphaFilter out of EKF as suggested by @dagar


EKF/estimator_interface.h
```diff
@@ -44,7 +44,7 @@
 #include <ecl.h>
 #include "common.h"
 #include "RingBuffer.h"
-#include "AlphaFilter.hpp"
+#include "../AlphaFilter/AlphaFilter.hpp"
```
dagar, line 47:
```suggestion
#include <AlphaFilter/AlphaFilter.hpp>
```

MaEtUgR: Applied and rebase on master 👍

PR #845, opened 2020-06-16
geo_lookup: create simple python script and update tables to WMM-2020

dagar:
This also includes minor changes updating to WMM-2020.


dagar:
I expanded the table slightly because there's a little bit of data beyond the blackout zone we're missing. Probably not that helpful, but the cost is minimal.

PR #848, opened 2020-06-22
GPS yaw improvements

bresch:
Summary of the changes:
- always reset yaw estimate when starting GPS yaw fusion
   - This is done because otherwise, if close to the limit (15 degrees), a large gyro bias can be learned and it takes a loooong time for the estimate to converge to the yaw GPS value
- disable GPS yaw fusion if there is no yaw data in the GPS message, if the fusion is rejected or if disabled by the user. The estimator will automatically fall back to another source of yaw aiding if possible (e.g.: mag fusion if enabled).
- add unit tests

non-functional change:
- do not check if the GPS yaw data is nan in the fusion function but only in `control.cpp`

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -526,6 +526,7 @@ void Ekf::controlGpsFusion()
 
 		} else {
 			if (ISFINITE(_gps_sample_delayed.yaw)) {
+
 				if (!_control_status.flags.gps_yaw
 				    && _control_status.flags.tilt_align
 				    && !_gps_hgt_intermittent) {
@@ -548,7 +549,8 @@ void Ekf::controlGpsFusion()
 			// Check if the data is constantly fused by the estimator,
 			// if not, declare the sensor faulty and stop the fusion
 			// By doing this, another source of yaw aiding is allowed to start
-			if (isTimedOut(_time_last_gps_yaw_fuse, (uint64_t)5e6)) {
+			if (isTimedOut(_time_last_gps_yaw_fuse, (uint64_t)5e6)
+			    && _control_status.flags.gps_yaw) {
 				_is_gps_yaw_faulty = true;
 				stopGpsYawFusion();
 			}
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -522,20 +522,14 @@ void Ekf::controlGpsFusion()
 		if ((_params.fusion_mode & MASK_USE_GPSYAW)
 				&& ISFINITE(_gps_sample_delayed.yaw)
 				&& _control_status.flags.tilt_align
-				&& (!_control_status.flags.gps_yaw || !_control_status.flags.yaw_align)
+				&& !_control_status.flags.gps_yaw
 				&& !_gps_hgt_intermittent) {
 
 			if (resetGpsAntYaw()) {
 				// flag the yaw as aligned
 				_control_status.flags.yaw_align = true;
 
-				// turn on fusion of external vision yaw measurements and disable all other yaw fusion
-				_control_status.flags.gps_yaw = true;
-				_control_status.flags.ev_yaw = false;
-				_control_status.flags.mag_dec = false;
-
-				stopMagHdgFusion();
-				stopMag3DFusion();
+				startGpsYawFusion();
 
 				ECL_INFO_TIMESTAMPED("starting GPS yaw fusion");
 			}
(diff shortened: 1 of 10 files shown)


EKF/gps_yaw_fusion.cpp
```diff
@@ -58,82 +58,75 @@ void Ekf::fuseGpsAntYaw()
 	float H_YAW[4];
 	float measured_hdg;
 
-	// check if data has been set to NAN indicating no measurement
-	if (ISFINITE(_gps_sample_delayed.yaw)) {
```
bresch, line 62: This is the modification, the rest is just indentation

EKF/gps_yaw_fusion.cpp
```diff
@@ -291,35 +284,20 @@ void Ekf::fuseGpsAntYaw()
 
 bool Ekf::resetGpsAntYaw()
 {
-	// check if data has been set to NAN indicating no measurement
-	if (ISFINITE(_gps_sample_delayed.yaw)) {
-
```
bresch, line 296: This check is removed as we we don't call this function if NAN

EKF/control.cpp
```diff
@@ -518,32 +518,42 @@ void Ekf::controlGpsFusion()
...
+				    && !_gps_hgt_intermittent) {
 
-				ECL_INFO_TIMESTAMPED("starting GPS yaw fusion");
+					// Activate GPS yaw fusion
+					if (resetGpsAntYaw()) {
+						_control_status.flags.yaw_align = true;
+
+						startGpsYawFusion();
+
+						ECL_INFO_TIMESTAMPED("starting GPS yaw fusion");
```
kamilritz, line 540: You could move this print statement into the `startGpsYawFusion` function

dagar: Annoying reminder to also consider if it's actually needed or could be reflected in the status flags. Printing to STDOUT can potentially impact real time performance.

bresch: I was thinking of getting rid of it: there is no message to tell when the mag fusion starts, I don't see why we should have one for GPS yaw. The status is indeed captured in the status flags so it's a bit superfluous.

test/test_EKF_gps_yaw.cpp
```diff
@@ -88,26 +90,61 @@ TEST_F(EkfGpsHeadingTest, fusionStartWithReset)
 	_sensor_simulator.runSeconds(11);
 
 	// THEN: after a while the fusion should be stopped
-	// TODO: THIS IS NOT HAPPENING
 	EXPECT_FALSE(_ekf_wrapper.isIntendingGpsHeadingFusion());
 }
 
-TEST_F(EkfGpsHeadingTest, fusionStartWithoutReset)
+TEST_F(EkfGpsHeadingTest, yawConvergence)
```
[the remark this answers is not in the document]
bresch: Not totally, the `fusionStartWithReset` function tests if the reset counter increases properly and that the fusion stops when GPS data stops for 10s. We could merge them together, but it would maybe test too many different things.

test/test_EKF_gps_yaw.cpp
```diff
@@ -88,26 +90,61 @@ TEST_F(EkfGpsHeadingTest, fusionStartWithReset)
...
+	EXPECT_NEAR(yaw_est, gps_heading, math::radians(0.5f))
+		<< "yaw est: " << math::degrees(yaw_est) << "gps yaw: " << math::degrees(gps_heading);
+}
+
+TEST_F(EkfGpsHeadingTest, fallBackToMag)
+{
+	// GIVEN: an initial GPS yaw, not aligned with the current one
+	// GPS yaw is expected to arrive a bit later, first feed some NANs
+	// to the filter
+	_sensor_simulator.runSeconds(6);
```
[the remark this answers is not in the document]
bresch: Yes, it's just to make sure that the timeout doesn't trigger if the yaw data takes more time than the vel/pos.

EKF/gps_yaw_fusion.cpp
```diff
@@ -273,6 +259,7 @@ void Ekf::fuseGpsAntYaw()
 
 	// only apply covariance and state corrections if healthy
 	if (healthy) {
+		_time_last_gps_yaw_fuse = _time_last_imu;
```
bresch, line 262: @kamilritz @priseborough Note that we usually have the `_time_last_xxx_fuse = _time_last_imu` line just after the innovation gate. This means that if not `healthy`, the fusion isn't performed but the timeout will never trigger. This is quite unlikely, but we might want to change that.

bresch:
> If a GPS is declared as faulty, should we try to use it again at some point or stop using it for the rest of the flight?

We might want to try again after some time, yes. I'll make another PR for this.

bresch:
@kamilritz I addressed your comments and rebased on master, could you please check again?

PR #850, opened 2020-06-23
Provide alternative to `git` introspection for version information

sevenbill:
This is part of what is required for using `bloom` to generate packages on the ROS buildfarm

Signed-off-by: Bill Morris <bill@sevendof.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,16 +48,12 @@ message(STATUS "build type is ${CMAKE_BUILD_TYPE}")
 set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug;Release;RelWithDebInfo;MinSizeRel;Coverage")
 
 if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
-        execute_process(
-                COMMAND git describe --always --tags
-                OUTPUT_VARIABLE git_tag
-                OUTPUT_STRIP_TRAILING_WHITESPACE
-                WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-                )
-elseif($ENV{ECL_LIB_GIT_TAG_STR})
-        set(git_tag $ENV{ECL_LIB_GIT_TAG_STR})
-else()
-        set(git_tag "")
+	execute_process(
+		COMMAND git describe --always --tags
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -47,12 +47,18 @@ message(STATUS "build type is ${CMAKE_BUILD_TYPE}")
...
 set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug;Release;RelWithDebInfo;MinSizeRel;Coverage")
 
-execute_process(
-	COMMAND git describe --always --tags
-	OUTPUT_VARIABLE git_tag
-	OUTPUT_STRIP_TRAILING_WHITESPACE
-	WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-	)
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
+        execute_process(
```
dagar, line 51: This is super picky, but would you mind converting the spaces to tabs to be consistent with the rest of the file?

CMakeLists.txt
```diff
@@ -47,12 +47,18 @@ message(STATUS "build type is ${CMAKE_BUILD_TYPE}")
...
+                COMMAND git describe --always --tags
+                OUTPUT_VARIABLE git_tag
+                OUTPUT_STRIP_TRAILING_WHITESPACE
+                WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                )
+elseif($ENV{ECL_LIB_GIT_TAG_STR})
+        set(git_tag $ENV{ECL_LIB_GIT_TAG_STR})
+else()
+        set(git_tag "")
```
dagar, line 60: Any use for keeping this case or should it be a FATAL_ERROR?

CMakeLists.txt
sevenbill: Fixed. Also https://github.com/PX4/ecl/pull/852

CMakeLists.txt
sevenbill: I simplified these cases out, as it looks like ${git_tag} is only used for a [STATUS message](https://github.com/PX4/ecl/pull/850/files#diff-af3b638bc2a3e6c650974192a53c7291L57)

PR #851, opened 2020-06-24
geo_lookup: upgrade table size (2 bytes per element) and add testing

dagar:
The single byte storage was insufficient in a few areas of the declination table. I've updated it to 2 bytes and scaled the inclination/declination/strength tables to use most of the range without being too awkward. The tables have been extended to include the full latitude range.

To minimize confusion (hopefully) I've expanded the API slightly to offer declination/inclination in both degrees/radians and the magnetic strength in Gauss and Tesla.

I've also generated some simple testing that verifies the interpolating between points.


dagar:
Screenshot from 2020-06-23 20-32-18

geo_lookup/geo_mag_declination.cpp
```diff
@@ -101,17 +101,28 @@ static constexpr float get_table_data(float lat, float lon, const int8_t table[L
 	return lat_scale * (data_max - data_min) + data_min;
 }
 
-float get_mag_declination(float lat, float lon)
-{
-	return get_table_data(lat, lon, declination_table);
+float get_mag_declination_radians(float lat, float lon) {
```
kamilritz, line 104:
```suggestion
float getMagDeclinationRadians(float lat, float lon) {
```

kamilritz: When we already update the interface, why not also enforce camelCase

dagar: > why not also enforce camelCase

Lack of any style guide or enforcement tool. I don't want to go crazy adopting an enormous style guide, but clearly we need something minimal in place to get the naming consistent.

dagar: > Lack of any style guide or enforcement tool. 

I'll look at getting something simple together.

dagar:
I propose we merge this as is and when we have some time later break this out as a new standalone library that can be shared by ECL, Firmware, sitl_gazebo, flightgear_bridge, etc. In the new library we can worry about getting the naming right, proper namespace, etc.

PR #852, opened 2020-06-25
Enforce tabs via editorcofig

sevenbill:


dagar:
Thanks, I wasn't aware of this.

PR #856, opened 2020-06-27
Random cleanup

kamilritz:
Collection of cleanups. Can be easily reviewed commit by commit.

One commit is replacing division by 2 or 4 with the respective multiplication of the inverse floating point number inside `covariance.cpp`. I am not sure if this was done on purpose. If not this was speeding up the function by quite some bit. Callgrind showed a decrease of the self cost of `predictCovariance` from 2.48 to 1.96. Not sure how representative these numbers are.


EKF/ekf.cpp
```diff
@@ -454,7 +454,7 @@ void Ekf::calculateOutputStates()
 
 			// calculate a velocity correction that will be applied to the output state history
 			// using a PD feedback tuned to a 5% overshoot
-			const float vert_vel_correction = vert_vel_integ_err * pos_gain + vert_vel_err * pos_gain * 1.1f;
+			const float vert_vel_correction = vert_vel_integ_err * pos_gain + vert_vel_err * vel_gain * 1.1f;
```
kamilritz, line 449: Current value for both gains are the same.

EKF/vel_pos_fusion.cpp
```diff
@@ -172,6 +172,7 @@ void Ekf::fuseVelPosHeight(const float innov, const float innov_var, const int o
 	// if the covariance correction will result in a negative variance, then
 	// the covariance matrix is unhealthy and must be corrected
 	bool healthy = true;
+	setVelPosFaultStatus(obs_index, false);
```
bresch, line 175: Good point, this could have been causing issues if for some values P < KHP but not the last one...

EKF/mag_fusion.cpp
```diff
@@ -173,6 +173,10 @@ void Ekf::fuseMag()
 	// before they are used to constrain heading drift
 	const bool update_all_states = ((_imu_sample_delayed.time_us - _flt_mag_align_start_time) > (uint64_t)5e6);
 
+	_fault_status.flags.bad_mag_x = false;
```
bresch, line 176: Nice, that's something really hard to see with those huge function cluttered with auto-generated code...

bresch:
The commits were clean, keeping the history to help with potential bisect.

PR #857, opened 2020-06-28
Make KFusion of type Vector24f

kamilritz:
Previously the Kalman gain was stored and passed to the `fuse` function as a c-array. This PR is changing its type to be of Vector24f. This allows to write the `fuse` function to be written more compactly.

All the Kalman gain container were and are initialized to 0. In multiple places components of the Kalman gain were set to zero for a second time. This is now removed too.

The commit "Make Kfusion a Vector24f" was changing the output of the change indicator by a tiny bit. I do not really understand why.


kamilritz:
I just checked, and there seems to be no significant difference in performance. @bresch

PR #858, opened 2020-06-30
GPS Yaw: fix heading initialisation and add unit tests

bresch:
When the antennas are not parallel to the x body axis, the GPS message contains the angular offset but the data is already corrected in the driver. EKF2 should then not add this offset during the initialization.

I also added unit tests with different initial orientations and antenna offsets.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -518,40 +518,7 @@ void Ekf::controlGpsFusion()
 		// Detect if coming back after significant time without GPS data
 		bool gps_signal_was_lost = isTimedOut(_time_prev_gps_us, 1000000);
 
-
-		if (!(_params.fusion_mode & MASK_USE_GPSYAW)
-		    || _is_gps_yaw_faulty) {
-
-			stopGpsYawFusion();
-
-		} else {
-			if (ISFINITE(_gps_sample_delayed.yaw)) {
-
-				if (!_control_status.flags.gps_yaw
-				    && _control_status.flags.tilt_align
-				    && !_gps_hgt_intermittent) {
-
-					// Activate GPS yaw fusion
-					if (resetGpsAntYaw()) {
-						_control_status.flags.yaw_align = true;
-						startGpsYawFusion();
-					}
-				}
-
-				if (_control_status.flags.gps_yaw) {
-					fuseGpsAntYaw();
-				}
-			}
-
-			// Check if the data is constantly fused by the estimator,
-			// if not, declare the sensor faulty and stop the fusion
-			// By doing this, another source of yaw aiding is allowed to start
-			if (isTimedOut(_time_last_gps_yaw_fuse, (uint64_t)5e6)
-			    && _control_status.flags.gps_yaw) {
-				_is_gps_yaw_faulty = true;
-				stopGpsYawFusion();
-			}
-		}
+		controlGpsYawFusion();
 
 		// Determine if we should use GPS aiding for velocity and horizontal position
 		// To start using GPS we need angular alignment completed, the local NED origin set and GPS data that has not failed checks recently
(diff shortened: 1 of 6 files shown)


EKF/control.cpp
```diff
@@ -535,6 +535,7 @@ void Ekf::controlGpsFusion()
 					if (resetGpsAntYaw()) {
 						_control_status.flags.yaw_align = true;
 						startGpsYawFusion();
+						_time_last_gps_yaw_fuse = _time_last_imu;
```
kamilritz, line 538: Should `_time_last_gps_yaw_fuse` be better set inside the `resetGpsAntYaw`?
Maybe we could also come up with a better name for `resetGpsAntYaw`, e.g. `resetYawToGps` or `resetYawToDualGps`, ...

EKF/gps_yaw_fusion.cpp
```diff
@@ -277,8 +277,8 @@ bool Ekf::resetGpsAntYaw()
 		return false;
 	}
 
-	// get measurement and correct for antenna array yaw offset
-	const float measured_yaw = _gps_sample_delayed.yaw + _gps_yaw_offset;
+	// GPS yaw measurement is alreday compensated for antenna offset in the driver
```
kamilritz, line 280: Why do we then need the offset variable at all?

EKF/control.cpp
bresch: It kind of makes sense to move it inside the reset function, but I followed the pattern used for the other resets e.g.: when we call `resetHorizontalPositionToGps`, we set `_time_last_gps = _time_last_imu;` just after. Not sure what's better.
https://github.com/PX4/ecl/blob/97225fcb7431ddcb2ca1e2f72ea79043597ebea6/EKF/control.cpp#L581-L593

EKF/gps_yaw_fusion.cpp
bresch: We need it to compute the angle of the antenna array with the vertical -Down- axis: https://github.com/PX4/ecl/blob/97225fcb7431ddcb2ca1e2f72ea79043597ebea6/EKF/gps_yaw_fusion.cpp#L59-L66

Then, we're actually doing the fusion in the antenna frame, but there is no real advantage because the innovation is the same.

EKF/control.cpp
kamilritz: I see. I would argue when the resetFunction matches only to one last_sensor_fuse, then we should move it into the resetFunction. So nobody can forget to set the timestamp of the last fusion.

kamilritz: I am not sure if your example above is suitable. `_time_last_gps_yaw_fuse` stores the timestamp of the last fusion event. 
`_time_last_gps` is the timestamp of the newest measurement in the buffer. I think it is wrong to update this `_time_last_gps` outside of the `EstimatorInterface::setGpsData` function in the first place. This seems to be a bug and should be removed.

kamilritz: Apparently, we not update any of the [fusion timestamps](https://github.com/PX4/ecl/blob/master/EKF/ekf.h#L344-L352) always during a reset of the respective state. So should we even do it for the gps_yaw, or should we do it everywhere?

EKF/gps_yaw_fusion.cpp
kamilritz: So could we get rid of the offset in general?

EKF/control.cpp
bresch: >_time_last_gps is the timestamp of the newest measurement in the buffer. I think it is wrong to update this _time_last_gps outside of the EstimatorInterface::setGpsData function in the first place. This seems to be a bug and should be removed.

I agree, not sure why this is there

bresch: >I see. I would argue when the resetFunction matches only to one last_sensor_fuse, then we should move it into the resetFunction. So nobody can forget to set the timestamp of the last fusion.

Ok, you convinced me, I moved it to the reset function. I was also not happy to have that independent gps yaw fusion control logic in the `controlGpsFusion` function so I extracted it into `controlGpsYawFusion`.

EKF/gps_yaw_fusion.cpp
bresch: To compute the innovation, I would say yes. I also thing that we should be able to use the function `fuseYaw312` instead of having a new one here; we basically just fuse a yaw measurement.
However, before I do that, I would have the opinion of @priseborough .

bresch: See https://github.com/PX4/ecl/pull/860

bresch:
@kamilritz Thanks for the review. I tried to remove the offset in #860 and it seems to work exactly the same.

Issue #824, opened 2020-05-20
controlGpsFusion position and velocity reset error.


garlinplus:
because resetPosition and resetVelocity function only return true.so when yaw is aligned and gps is effective,it's unnecessary to judge return value of resetPosition and resetVelocity.
				if (_control_status.flags.yaw_align) {
					// if we are not already aiding with optical flow, then we need to reset the position and velocity
					// otherwise we only need to reset the position
					_control_status.flags.gps = true;

//					if (!_control_status.flags.opt_flow) {
//						if (!resetPosition() || !resetVelocity()) {
//							_control_status.flags.gps = false;
//						}
//
//					} else if (!resetPosition()) {
//						_control_status.flags.gps = false;
//
//					}
					if (!_control_status.flags.opt_flow) {
						resetPosition();
						resetVelocity();
					} else {
						resetPosition();
					}

					if (_control_status.flags.gps) {
						ECL_INFO_TIMESTAMPED("commencing GPS fusion");
						_time_last_gps = _time_last_imu;
					}
				}

bresch:
That's correct, thanks for reporting

PR #859, opened 2020-07-02
Return type of resets

kamilritz:
fixes https://github.com/PX4/ecl/issues/824


kamilritz:
Seems like the Firmware Build Check is still failing sometimes when PR child is a fork.

PR #860, opened 2020-07-02
GPS Yaw: use fuseYaw321/312 functions

bresch:
Saves 840 bytes of flash space

With the current implementation:
2020-07-02_13-03-29_01_plot
Screen Capture_select-area_20200702134057

Replayed with this PR:
2020-07-02_13-03-44_01_plot
Screen Capture_select-area_20200702133840

Original log file:
https://logs.px4.io/plot_app?log=46feda49-d644-4ed0-acbd-868ff70779eb

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -537,20 +537,23 @@ private:
 	// yaw : angle observation defined as the first rotation in a 321 Tait-Bryan rotation sequence (rad)
 	// yaw_variance : variance of the yaw angle observation (rad^2)
 	// zero_innovation : Fuse data with innovation set to zero
-	void fuseYaw321(const float yaw, const float yaw_variance, bool zero_innovation);
+	// return true if the fusion was successful
+	bool fuseYaw321(const float yaw, const float yaw_variance, bool zero_innovation = false);
 
 	// fuse the yaw angle defined as the first rotation in a 312 Tait-Bryan rotation sequence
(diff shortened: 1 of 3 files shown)


kamilritz:
For normal yaw fusion we have fusionYaw321 and fusionYaw312. Should we not consider both here?

bresch:
>For normal yaw fusion we have fusionYaw321 and fusionYaw312. Should we not consider both here?

Absolutely, I ignored it because I thought we were protected against gimbal lock by the antenna tilt check, but this is not the case if we have an antenna offset of 90 degrees.

bresch:
We could also move it to the fuseHeading function that already contains the `ev_yaw` and `mag_hdg` fusion logic, but it will require a bit of cleanup first. I think it's simpler if we do it step by step.

priseborough:
The GPS yaw fusion requires its own method because it supports an arbitrary antenna orientation in the XY body frame - see https://github.com/PX4/ecl/blob/46a5e6e4285df7f16ebc7b81d43149324d86133d/EKF/matlab/scripts/Inertial%20Nav%20EKF/GenerateNavFilterEquations.m#L422-L434

bresch:
>The GPS yaw fusion requires its own method because it supports an arbitrary antenna orientation in the XY body frame

@priseborough Yes, but how is it different from rotating the measurement by the antenna offset and then fuse it like a magnetometer (in heading fusion mode) ?

PR #862, opened 2020-07-02
Use correct literal

kamilritz:
Use 2.f instead of 2


dagar:
Shouldn't this be updated a step before in the generation?

kamilritz:
Indeed, but I try to stay away from those scripts.

bresch:
@kamilritz I talked about that with @priseborough and we'll make sure we get it right with the new python auto-generated code but I don't think it's worth changing now; the output of the compiler would still be the same.

PR #864, opened 2020-07-07
Negative distances go in the opposite direction

jkflying:
Previous 2D (and 3D) projection functions I've worked with have all responded to negative distance with as being in the opposite of the positive distance. Up for discussion whether this should be changed.

(part of my demo for the dev summit)

### DIFF
--- a/geo/geo.cpp
+++ b/geo/geo.cpp
@@ -304,7 +304,7 @@ void waypoint_from_heading_and_distance(double lat_start, double lon_start, floa
 					double *lat_target, double *lon_target)
 {
 	bearing = wrap_2pi(bearing);
-	double radius_ratio = (double)fabs((double)dist) / CONSTANTS_RADIUS_OF_EARTH;
+	double radius_ratio = dist / CONSTANTS_RADIUS_OF_EARTH;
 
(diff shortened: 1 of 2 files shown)


dagar:
I agree that definitely sounds like the right thing to do. We should check the existing usage quickly.

dagar:
I only saw one potentially problematic area in navigator where we need to be careful with the loiter_radius that's interpreted as negative for counter-clockwise.

https://github.com/PX4/Firmware/blob/b1641740379787f6fcc1ab2a0b9e361e6be0534e/src/modules/navigator/mission_block.cpp#L441-L444

kamilritz:
@dagar The Firmware test failed here for a reason. Please pay attention to them before merging.

dagar:
> @dagar The Firmware test failed here for a reason. Please pay attention to them before merging.

The danger with tolerating these builds being flaky at different periods is people stop looking at them. If things are expected to be working now I'd set them as required in PX4/ecl branch protection.

kamilritz:
Agree, it is not flaky anymore. There is also a mechanism in place that allows to perform update to the interface of ECL: create a branch with the same name as the ecl branch in Firmware.

PR #865, opened 2020-07-10
Fix firmware-build-test

kamilritz:
It seems that I finally got it.
There is a [race condition in the Github checkout action v1](https://github.com/actions/checkout/issues/23#issuecomment-553994987), which likely caused the issue.
This PR uses the newest version of the checkout action. But for creating a local git repository during the checkout a git version >2.18 is now necessary. Therefore the dev-base-focal container needs to be used.

### DIFF
--- a/.github/workflows/firmware_build_test.yml
+++ b/.github/workflows/firmware_build_test.yml
@@ -1,9 +1,6 @@
 name: Firmware Build Test
 
 on:
-  push:
-    branches:
-      - master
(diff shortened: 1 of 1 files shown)


kamilritz:
@dagar could you have a look?

kamilritz:
@bresch @dagar @priseborough Let get this in. The firmware build test is really failing constantly now, and I expect this PR to fix it.

PR #866, opened 2020-07-11
Variadic KHP computation

kamilritz:
I just learned about variadic functions and templates. It seems that they could be used to define a single function that performs a sparse computation of the KHP matrix, which can be reused in multiple places through the code base. 5 places without further modifications.
Since the measurement jacobian is often very sparse, it is beneficial to neglect the zero entries during the vector-matrix multiplication.

Variadic functions are usually implemented in a recursive manner, so is this proposed implementation. This has the drawback that we can not explicitly cache the KH[i] = K(row) * H(x) terms to reduce the amount of multiplications.

TODO:
-[] Add test to cover these changes. Add the moment it is not covered at all. Plus and minus can be exchanged without failure.
-[] Test performance issues
-[] Test numeric stability

### DIFF
--- a/EKF/EKFGSF_yaw.cpp
+++ b/EKF/EKFGSF_yaw.cpp
@@ -236,10 +236,10 @@ void EKFGSF_yaw::ahrsAlignYaw()
 
 		} else {
 			// Calculate the 312 Tait-Bryan rotation sequence that rotates from earth to body frame
-			Vector3f rot312;
-			rot312(0) = wrap_pi(_ekf_gsf[model_index].X(2)); // first rotation (yaw) taken from EKF model state
-			rot312(1) = asinf(_ahrs_ekf_gsf[model_index].R(2, 1)); // second rotation (roll)
-			rot312(2) = atan2f(-_ahrs_ekf_gsf[model_index].R(2, 0), _ahrs_ekf_gsf[model_index].R(2, 2));  // third rotation (pitch)
+			const Vector3f rot312(wrap_pi(_ekf_gsf[model_index].X(2)),  // yaw
+					      asinf(_ahrs_ekf_gsf[model_index].R(2, 1)),  // roll
+					      atan2f(-_ahrs_ekf_gsf[model_index].R(2, 0),
+						      _ahrs_ekf_gsf[model_index].R(2, 2)));  // pitch
 
 			// Calculate the body to earth frame rotation matrix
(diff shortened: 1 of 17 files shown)


kamilritz:
@jkflying Thank you very much for your feedback.

> This roughly doubles the number of multiplications.

It is not that bad, as the common factor `K(row)` only needs to be multiplied once.

I also thought about a function with the signature you mentioned. My thought was that if the compiler generates and optimizes code for each caller, that we might end up with faster code. Do you think that this would be the case?

Anyway, your arguments regarding flash space reduction and readability are very convincing.

jkflying:
> My thought was that if the compiler generates and optimizes code for each caller, that we might end up with faster code. Do you think that this would be the case?

I don't think there will be useable optimizations that can be achieved beyond the manual code written now, which is pretty close to what the signature I proposed above would be. Also each will have its own flash usage.

Also, calling can be something like 
`auto khp = computeKHP(K, H, {1, 2, 3, 4});`

kamilritz:
@jkflying The normal `matrix::Vector` does not seem to allow initialization from brace initalizer list. `Vector2` and `Vector3` does as they have a constructor(x,y(,z)). Is there a good way to add this as well for `matrix::Vector`?

jkflying:
The normal way is via `std::initializer_list` like:
```C++
    Vector(std::initializer_list<Type> list)
    {
        assert(list.size() == M);
        std::copy(list.begin(), list.end(), &MatrixM1::_data[0][0]);
    }
```

But (bad news), `<initializer_list>` isn't available in Nuttx builds. IMO some things like this should be brought into the Nuttx builds, they don't have any runtime cost.

With what is there now it can be achieved something like this:
```C++
template<size_t num_indices>
SquareMatrix24f computeKHP(const Vector24f& K, const Vector24f& H, size_t col_indices[num_indices] ){
...
}
...
auto khp = computeKHP(K, H, (size_t[]){1, 2, 3, 4});
```

But [this is technically only a C99 feature](https://stackoverflow.com/questions/28116467/are-compound-literals-standard-c), and C++ might not be happy in certain cases.

_This_ issue can be worked around by changing to
```C++
template<size_t num_indices>
SquareMatrix24f computeKHP(const Vector24f& K, const Vector24f& H, size_t (&col_indices)[num_indices] ){
...
}
...
auto khp = computeKHP(K, H, {1, 2, 3, 4});
```

Which has an ugly signature, but works great :joy: 


@dagar any chance of getting a few more things from the C++ std lib into Nuttx builds, eg. `std::array` and its dependencies?

PR #874, opened 2020-07-26
Make delayed output state non-member variable

kamilritz:
It helps to look at commit it by commit.
The `output_*_delayed` states are used to compute the correction that gets applied to the output filter, such that the `output_*_delayed` states are tracking the EKF delayed states.

It seems that it is not necessary to have a member variable to store this delayed output. Storing it only introduces the need to keep it in sync with the buffer itself, during a EKF state reset.

In addition, I notice that we get the delayed output by calling the buffers `get_oldest` function. Will the oldest entry in the buffer always be close to the delayed fusion horizon?
Plus, in the alignOutputFilter, the output_new's quaternion gets updated, but not its velocity or position. Bug?

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -426,8 +426,32 @@ void Ekf::calculateOutputStates()
 		const float vel_gain = _dt_ekf_avg / math::constrain(_params.vel_Tau, _dt_ekf_avg, 10.0f);
 		const float pos_gain = _dt_ekf_avg / math::constrain(_params.pos_Tau, _dt_ekf_avg, 10.0f);
 
-		applyCorrectionToVerticalOutputBuffer(vel_gain, pos_gain);
-		applyCorrectionToOutputBuffer(vel_gain, pos_gain);
+		// calculate down velocity and position tracking errors
+		const float vert_vel_err = (_state.vel(2) - _output_vert_delayed.vert_vel);
+		const float vert_vel_integ_err = (_state.pos(2) - _output_vert_delayed.vert_vel_integ);
+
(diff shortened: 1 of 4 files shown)


priseborough:
> In addition, I notice that we get the delayed output by calling the buffers `get_oldest` function. Will the oldest entry in the buffer always be close to the delayed fusion horizon?

It is by definition. This is the method used to extract the oldest IMU data from the buffer. The time stamp of that oldest IMU data defines the delayed fusion time horizon and that data is used for prediction only. Observation data is extracted using the pop_first_older_than function which finds the newest data that has fallen behind the fusion time horizon defined by the _imu_sample_delayed.time_us timestamp.

> Plus, in the alignOutputFilter, the output_new's quaternion gets updated, but not its velocity or position. Bug?

It gets updated the next time applyCorrectionToOutputBuffer is called, which does 

_output_new = _output_buffer.get_newest();

This is happens at the end of each Ekf::update() and the accessor functions for  _output_new.vel,pos are not called in ekf2_main.cpp until after the Ekf::update() has run so currently this is not causing a change in behaviour. However this is a potential trap for future development so would recommend that in alignOutputFilter we replace:

	_output_new.quat_nominal = q_delta * _output_new.quat_nominal;
	_output_new.quat_nominal.normalize();

	_output_sample_delayed.quat_nominal = q_delta * _output_sample_delayed.quat_nominal;
	_output_sample_delayed.quat_nominal.normalize();


with

	_output_new = _output_buffer.get_newest();

	_output_sample_delayed.quat_nominal = _state.quat;
	_output_sample_delayed.vel = _state.vel;
	_output_sample_delayed.pos = _state.pos;

PR #879, opened 2020-07-30
EKF: Convert magnetic field observation methods to use SymPy generated code

priseborough:
This PR replaces the Matlab generated equations with sympy generated equations. Continues on from #870.

Modifies the methods used for fusion of 3-axis magnetometer data, declination angle and yaw angle.
Includes comparison test scripts that compare the observation Jacobians and Kalman gain equations generated for the Matlab symbolic toolbox and SymPy.

**Testing**

3Dmag_fusion_generated_compare.cpp output:

Pass: Magnetomer X axis Hfusion max diff fraction = 0.000000e+00
Pass: Magnetomer X axis Kfusion max diff fraction = 7.846418e-08
Pass: Magnetomer Y axis Hfusion max diff fraction = 0.000000e+00
Pass: Magnetomer Y axis Kfusion max diff fraction = 1.035609e-07
Pass: Magnetomer Z axis Hfusion max diff fraction = 0.000000e+00
Pass: Magnetomer Z axis Kfusion max diff fraction = 0.000000e+00

mag_decl_fusion_generated_compare.cpp output:

Pass: Mag Declination Hfusion max diff fraction = 0.000000e+00
Pass: Mag Declination Kfusion max diff fraction = 1.228912e-07

yaw_fusion_generated_compare.cpp output:

Pass: 321 yaw option A Hfusion max diff fraction = 2.090426e-07
Pass: 321 yaw option B Hfusion max diff fraction = 1.139841e-07
Pass: 312 yaw option A Hfusion max diff fraction = 1.333159e-07
Pass: 312 yaw option B Hfusion max diff fraction = 1.998272e-07

Comparison of mag fusion innovations and field estimates before and after the change:

Before:

Screen Shot 2020-07-30 at 2 59 46 pm

After:

Screen Shot 2020-07-30 at 3 00 37 pm

Comparison of yaw innovations and innovation variances before and after the change:

Before:

Screen Shot 2020-07-30 at 3 35 12 pm

After:

Screen Shot 2020-07-30 at 3 33 40 pm


EKF/mag_fusion.cpp
```diff
@@ -142,6 +203,19 @@ void Ekf::fuseMag()
 
 	}
 
+	// rotate magnetometer earth field state into body frame
```
kamilritz, line 206: Wondering if it would make sense to auto generated the computation of `mag_innov` as well. I assume we could remove some duplicated computation.

EKF/mag_fusion.cpp
```diff
@@ -47,51 +48,77 @@
...
+	const float HK43 = HK29*P(1,3);
+	const float HK44 = HK30*P(0,3);
+	const float HK45 = HK21*P(3,16) + HK27*P(3,17) - HK28*P(3,18) - HK32*P(2,3) + HK34*P(3,3) + HK43 + HK44 + P(3,19);
+	const float HK46 = HK32*P(1,2);
+	const float HK47 = HK34*P(1,3);
+	const float HK48 = HK21*P(1,16) + HK27*P(1,17) - HK28*P(1,18) + HK29*P(1,1) + HK30*P(0,1) - HK46 + HK47 + P(1,19);
+	const float HK49 = HK21*P(16,19) + HK27*P(17,19) - HK28*P(18,19) + HK29*P(1,19) + HK30*P(0,19) - HK32*P(2,19) + HK34*P(3,19) + P(19,19);
+
+	_mag_innov_var(0) = (HK21*HK37 + HK27*HK42 - HK28*HK38 + HK29*HK48 + HK30*HK36 - HK32*HK41 + HK34*HK45 + HK49 + R_MAG);
+	float HK50;
```
kamilritz, line 117:
```suggestion
	if (_mag_innov_var(0) < R_MAG) {
		// the innovation variance contribution from the state covariances is negative which means the covariance matrix is badly conditioned
		_fault_status.flags.bad_mag_x = true;
		// we need to re-initialise covariances and abort this fusion step
		resetMagRelatedCovariances();
		ECL_ERR("magX %s", numerical_error_covariance_reset_string);
		return;
	}
	_fault_status.flags.bad_mag_x = false;
	const float HK50 = 1.0F/_mag_innov_var(0);
```
The same applies to the other axes

EKF/mag_fusion.cpp
```diff
@@ -170,7 +244,7 @@ void Ekf::fuseMag()
 	const bool update_all_states = ((_imu_sample_delayed.time_us - _flt_mag_align_start_time) > (uint64_t)5e6);
 
 	// Observation jacobian and Kalman gain vectors
-	Vector24f H_MAG;
+	float Hfusion[24];
```
kamilritz, line 247: What about making this a Vector8f with no zero elements by hand?

EKF/mag_fusion.cpp
```diff
@@ -179,99 +253,88 @@ void Ekf::fuseMag()
...
+
+				for (unsigned row = 4; row <= 15; row++) {
+					Kfusion(row) = HK50*(HK21*P(row,16) + HK27*P(row,17) - HK28*P(row,18) + HK29*P(1,row) + HK30*P(0,row) - HK32*P(2,row) + HK34*P(3,row) + P(row,19));
+				}
+
+				for (unsigned row = 22; row <= 23; row++) {
+					Kfusion(row) = HK50*(HK21*P(row,16) + HK27*P(row,17) - HK28*P(row,18) + HK29*P(1,row) + HK30*P(0,row) - HK32*P(2,row) + HK34*P(3,row) + P(row,19));
+				}
+			} else {
+				for (uint8_t i = 0; i < 16; i++) {
```
kamilritz, line 281: No setting to 0 necessary

EKF/mag_fusion.cpp
```diff
@@ -179,99 +253,88 @@ void Ekf::fuseMag()
...
+				for (unsigned row = 22; row <= 23; row++) {
+					Kfusion(row) = HK71*(HK29*P(2,row) - HK30*P(3,row) + HK32*P(1,row) + HK34*P(0,row) + HK53*P(row,17) + HK57*P(row,18) - HK58*P(row,16) + P(row,20));
+				}
+			} else {
+				for (uint8_t i = 0; i < 16; i++) {
+					Kfusion(i) = 0.0f;
+				}
+
+				Kfusion(22) = 0.0f;
+				Kfusion(23) = 0.0f;
```
kamilritz, line 329: no setting to 0 necessary

EKF/mag_fusion.cpp
```diff
@@ -859,81 +890,57 @@ void Ekf::fuseHeading()
...
 	// assign intermediate state variables
-	const float magN = _state.mag_I(0);
-	const float magE = _state.mag_I(1);
+	const float &magN = _state.mag_I(0);
+	const float &magE = _state.mag_I(1);
 
-	// minimum horizontal field strength before calculation becomes badly conditioned (T)
-	constexpr float h_field_min = 0.001f;
+	// minimum North field strength before calculation becomes badly conditioned (T)
+	const float N_field_min = 0.001f;
```
kamilritz, line 897:
```suggestion
	constexpr float N_field_min = 0.001f;
```

EKF/mag_fusion.cpp
```diff
@@ -859,81 +890,57 @@ void Ekf::fuseHeading()
...
+	const float HK7 = powf(HK1, -2);
+	const float HK8 = HK5*P(16,16) - P(16,17);
+	const float innovation_variance = -HK0*HK6*HK7 + HK7*HK8*magE/powf(magN, 3) + R_DECL;
+	float HK9;
+	if (innovation_variance > R_DECL) {
+		HK9 = HK4/innovation_variance;
 	} else {
+		// variance calculation is badly conditioned
 		return;
 	}
```
kamilritz, line 923:
```suggestion
	if (innovation_variance < R_DECL) {
		// variance calculation is badly conditioned
		return;
	}
	const float HK9 = HK4/innovation_variance;
```

EKF/mag_fusion.cpp
```diff
@@ -859,81 +890,57 @@ void Ekf::fuseHeading()
...
 
 	// Calculate the observation Jacobian
 	// Note only 2 terms are non-zero which can be used in matrix operations for calculation of Kalman gains and covariance update to significantly reduce cost
-	Vector24f H_DECL;
-	H_DECL(16) = -magE*t21;
-	H_DECL(17) = magN*t21;
+	// Note Hfusion indices do not match state indices
+	float Hfusion[2] = {};
```
kamilritz, line 928: What about using Vector2f?

kamilritz:
There are two ways how we compute the mag update. The normal sequential one and then the alternative. I am not sure if the alternative way is conceptually fully correct. 
We compute first the innovation variances for all axes with the help of some shared intermediate variables. Then we fuse sequentially all axes, but during this we reuse shared intermediate variables from the innovation variance computation for the computation of H and K. If we want to be fully correct we would not be allowed to reuse some of the shared intermediate variables as they are a function of P, which changes after each sequential fusion.
@priseborough Do you think this is critical?

priseborough:
I don't think it's an issue given they are orthogonal observations and i couldn't detect a visual difference either way when doing the early comparisons, but technically it is an approximation because the covariances will have changed as each axis is fused.

I will create an alternative version using the other method and lets see how much flash and operations we are saving. If the difference is small, then the more independent method is lower risk.

kamilritz:
@priseborough @jkflying I created a SparseVector class, that only stores the N nonzero elements in a Vector<Type, N>, but still allows to access them with the original index. E.g.
```
SparseVector<float, 16, 17> a;  // owns a Vector<float, 2>
a(16) = ...;
a(17) = ...;
a(1) = ...;  // will assert during compilation as index is invalid
```
When calling the operator() with a compile-time constant as above, finding the right entry in the Vector<Type, N> will happen at compile time if using at least -O1 optimization. See here for produced assembly code in a toy example: https://godbolt.org/z/f7f9f9.

I think this would be a nice solution for the measurement jacobian, as it helps to avoid modifying the auto generated code (except for removing the zero lines), but still avoiding a Vector with a lot of nonzero elements that are stored in memory.

I took the liberty to add the class to this PR and applying it to the mag related measurement jacobians. @jkflying could you please have a look at the SparseVector class and let me know what you think about this approach?

EKF/mag_fusion.cpp
kamilritz: Done

EKF/mag_fusion.cpp
kamilritz: Done

EKF/mag_fusion.cpp
kamilritz: Done

EKF/mag_fusion.cpp
kamilritz: Done

EKF/mag_fusion.cpp
kamilritz: Done

jkflying:
> I took the liberty to add the class to this PR and applying it to the mag related measurement jacobians. @jkflying could you please have a look at the SparseVector class and let me know what you think about this approach?

This looks really good, I'd actually like to get it into Matrix first. There's a few things I think should be improved (eg handling of non-dense indices should return 0, add a dot product with dense vector (or other sparse vector) functionality, etc), but the use of the variadic template values in this case makes a lot of sense. Also I don't think ECL/sympy PR is the right place to review this in depth :wink: Once we have it in Matrix with the functionality we want, we can then go back and reduce the memory consumption of the new code generation, otherwise I think as SparseMatrix undergoes some small API changes there might be a lot of churn in ECL.

kamilritz:
> This looks really good, I'd actually like to get it into Matrix first. There's a few things I think should be improved (eg handling of non-dense indices should return 0, add a dot product with dense vector (or other sparse vector) functionality, etc), but the use of the variadic template values in this case makes a lot of sense. Also I don't think ECL/sympy PR is the right place to review this in depth wink Once we have it in Matrix with the functionality we want, we can then go back and reduce the memory consumption of the new code generation, otherwise I think as SparseMatrix undergoes some small API changes there might be a lot of churn in ECL.

Ok, let me remove it from here. And open a PR on Matrix. Feel free to modify it as you like.

jkflying:
@kamilritz some inspiration, note that there are recursion inlining limits hit in `run2` unless you make the indices template parameters. If you make them template parameters, you can have compile-time checking that your indices are correct.

https://godbolt.org/z/zrzGMa

priseborough:
I've created a branch here that implements the fusion without the assumption of invariant covariance and states between fusion of orthogonal axes - https://github.com/PX4/ecl/tree/pr-ekfSymPyMagFusionAlt

The results from replay are visually identical:

Approximation method (this PR):
temporal_approximation

Exact method from https://github.com/PX4/ecl/tree/pr-ekfSymPyMagFusionAlt
temporal_exact

The next step is to see if i can generate some profiling comparison data using SITL - @dagar is there a way to do profiling with replay?

If the cost isn't much more, then I would prefer the 'exact' method. The previous sequential fusion implementation using the Matlab equations also had some built in assumptions of invariance, so the method implemented in https://github.com/PX4/ecl/tree/pr-ekfSymPyMagFusionAlt is theoretically better than what we had before.

priseborough:
Here are the callgrind comparisons generated using "HEADLESS=1 make px4_sitl jmavsim___callgrind"  to generate the callgrind.out file and using kcachegrind to generate the figures. These have been zoomed to show the 3d mag fusion cycle count estimates for the mag fusion processes as a % of the ekf update() function.

Approximate method (assumes invariant states and covariance between fusion of each axis:

approximation

Exact method from https://github.com/PX4/ecl/tree/pr-ekfSymPyMagFusionAlt branch:

exact

The increase in cycle count is negligible so i will pull in the patch for the exact method.

priseborough:
@kamilritz I got the following build error when I tried to test your changes;

In file included from ../../src/lib/ecl/EKF/drag_fusion.cpp:42:0:
../../src/lib/ecl/EKF/ekf.h:56:34: error: ‘SparseVectorf’ in namespace ‘matrix’ does not name a template type
  using SparseVector24f = matrix::SparseVectorf<24, Idxs...>;

kamilritz:
> @kamilritz I got the following build error when I tried to test your changes;
> 
> In file included from ../../src/lib/ecl/EKF/drag_fusion.cpp:42:0:
> ../../src/lib/ecl/EKF/ekf.h:56:34: error: ‘SparseVectorf’ in namespace ‘matrix’ does not name a template type
> using SparseVector24f = matrix::SparseVectorf<24, Idxs...>;
> 
> Edit: The Firmware build test is failing for the same reason.

I am on it. I need to bring the matrix update into Firmware first. It is a bit of a hurdle race, but I am soon there.

priseborough:
I've been through the SparseVector24 code changes and verified that attempting to access a 'blank' index for either read or write gives an unambiguous build error.

Will generate updated change indication.

EKF/mag_fusion.cpp
priseborough: It's just a coordinate rotation so doesn't buy us much.

kamilritz:
@priseborough It would be great if you could re-run the log to check if everything is still okay. After this we should be good to merge this.

priseborough:
Replay results before and after change shown below. They are visually the same.

Before:
before


After:
after

kamilritz:
@priseborough could update the change indication file again. It seems I made a mistake when merging it. I think if you just add it again in a new commit. it will be okay.

priseborough:
I will rebase and generate it again.

PR #880, opened 2020-07-30
EKF: Convert GPS yaw observation method to use SymPy generated code

priseborough:
This PR replaces the Matlab generated equations with sympy generated equations. Continues on from #870.

Modifies the methods used for fusion of GPS receiver yaw angle.
Includes comparison test scripts that compare the observation Jacobians and Kalman gain equations generated for the Matlab symbolic toolbox and SymPy.

**Testing**

gps_yaw_fusion_generated_compare.cpp output:

Pass: GPS yaw Hfusion max diff fraction = 1.796981e-07
Pass: GPS yaw Kfusion max diff fraction = 3.981467e-07

SITL testing TODO

### DIFF
--- a/EKF/gps_yaw_fusion.cpp
+++ b/EKF/gps_yaw_fusion.cpp
@@ -117,17 +117,12 @@ void Ekf::fuseGpsYaw()
 	const float HK29 = 4/ecl::powf(HK17, 2);
 	const float HK30 = -HK16*P(0,2) - HK24*P(1,2) - HK25*P(2,2) + HK26*P(2,3);
 	const float HK31 = -HK16*P(0,3) - HK24*P(1,3) - HK25*P(2,3) + HK26*P(3,3);
-	float HK32;
 
 	// check if the innovation variance calculation is badly conditioned
 	_heading_innov_var = (-HK16*HK27*HK29 - HK24*HK28*HK29 - HK25*HK29*HK30 + HK26*HK29*HK31 + R_YAW);
-	if (_heading_innov_var >= R_YAW) {
-		// the innovation variance contribution from the state covariances is not negative, no fault
-		_fault_status.flags.bad_hdg = false;
-		HK32 = HK18/_heading_innov_var;
 
-	} else {
-		// the innovation variance contribution from the state covariances is negative which means the covariance matrix is badly conditioned
+	if (_heading_innov_var < R_YAW) {
+	// the innovation variance contribution from the state covariances is negative which means the covariance matrix is badly conditioned
 		_fault_status.flags.bad_hdg = true;
 
 		// we reinitialise the covariance matrix and abort this fusion step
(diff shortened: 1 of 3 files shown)


priseborough:
If someone with a RTK yaw GPs setup can provide a log for replay, then that would be appreciated. TXS.

bresch:
@priseborough This is a log that I used in the past to replay GPS yaw data: https://logs.px4.io/plot_app?log=46feda49-d644-4ed0-acbd-868ff70779eb

EKF/python/ekf_derivation/generated/gps_yaw_fusion_generated_compare.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+float sq(float in) {
+    return in * in;
+}
+
+int main()
+{
+    // Compare calculation of observation Jacobians and Kalman gains for sympy and matlab generated equations
+
+	float Hfusion[24];
```
bresch, line 17: A few indentation issues in this file.

EKF/gps_yaw_fusion.cpp
```diff
@@ -68,88 +69,62 @@ void Ekf::fuseGpsYaw()
...
+	// TODO extend interface to use yaw uncertainty provided by GPS if available
+	const float R_YAW = sq(fmaxf(_params.mag_heading_noise, 1.0e-2f));
+
+	// calculate intermediate variables
+	const float HK0 = sinf(_gps_yaw_offset);
+	const float HK1 = q0*q3;
+	const float HK2 = q1*q2;
+	const float HK3 = 2*HK0*(HK1 - HK2);
+	const float HK4 = cosf(_gps_yaw_offset);
+	const float HK5 = ecl::powf(q1, 2);
```
bresch, line 82: What's the reason for using `ecl::powf(q1, 2)` instead of `q1*q1`?

jkflying: Sympy generates `powf(a,b)` a lot, so we added a custom `ecl::powf` which only accepts integer powers and just does the multiplication so that we don't accidentally do `powf(float,float)` style calculations which the compiler can't optimize to something like `x*x`. Then it's easy to just find/replace powf with ecl::powf and make sure everything is covered.

priseborough:
I was able to test this in SITL using a modification to ekf2_main.cpp that set the gps.heading value using the vehicle_attitude_ground_truth message and EKF2_AID_MASK = 129

Screen Shot 2020-08-11 at 8 52 32 am

A replay comparison using the log provided gave identical innovation and innovation variance sequences before:
before


and after the change:
after

priseborough:
@bresch and @kamilritz My understanding is that the review issues have been addressed and testing is compete.

kamilritz:
@priseborough I would prefer, when we do the mag one first and copy the new standard way of handling the measurement jacobian to here.

priseborough:
OK, can wait for you to get the matrix library changes required into firmware.

priseborough:
Here are the before at commit and after at commit https://github.com/PX4/ecl/pull/880/commits/8e52aedb48affb93235ebec0a375cecf183b3f8e results which are visually identical.

Before:
before

After:
after

priseborough:
> I just added the missing file with the auto generated code. Is there a reason why we now do not write `ecl::powf`?

We could incorporate a text find and replace, but it would have to check that the exponent was an integer.

EKF/python/ekf_derivation/generated/gps_yaw_fusion_generated_compare.cpp
priseborough: Fixed

PR #881, opened 2020-07-30
geo: split out geo helpers and move back to PX4/Firmware

dagar:
Only the map projection helpers are used in ecl, the rest we can move back to PX4/Firmware.


dagar:
To avoid conflicts/confusion I've renamed `geo` within ecl to `map_projection`. I'm happy to rename it again if someone has a better suggestion.

dagar:
Corresponding PX4/Firmware change. https://github.com/PX4/Firmware/pull/15454

jkflying:
Maybe get https://github.com/PX4/ecl/pull/864 in as part of this migration?

PR #883, opened 2020-08-09
Update matrix version

kamilritz:
Needed for updating matrix version in whole Firmware


dagar:
> Needed for updating matrix version in whole Firmware

I'm not sure if you're aware, but when ecl is included in PX4/Firmware it uses Matrix from the submodule here. https://github.com/PX4/Firmware/tree/master/src/lib

kamilritz:
> I'm not sure if you're aware, but when ecl is included in PX4/Firmware it uses Matrix from the submodule here. https://github.com/PX4/Firmware/tree/master/src/lib

I know. But ECL needs to comply with new version first. Additionally, I updated the matrix version for the standalone build.

PR #887, opened 2020-08-11
EKF: eliminate duplicate SymPy code_gen

priseborough:


priseborough:
> Could you include the auto generated code in the new location?

Done

PR #888, opened 2020-08-13
EKF: Convert optical flow fusion to use SymPy generated equations

priseborough:
Replaces the Matlab symbolic toolbox auto generated code used by the optical flow fusion method with code generated using the SymPy library.

Innovation and innovation variances before and after the change generated using replay of a SILT log are visually identical:

Before:
before

After:
after

Output from the test program comparing the Jacobians and Kalman gain values from the Matlab and SymPy method is shown below. Difference s are within the expect range for single precision operations.

Pass: Optical Flow X axis Hfusion max diff fraction = 7.477993e-08
Pass: Optical Flow X axis Kfusion max diff fraction = 3.627760e-06
Pass: Optical Flow Y axis Hfusion max diff fraction = 1.010612e-07
Pass: Optical Flow Y axis Kfusion max diff fraction = 1.309633e-06


EKF/optflow_fusion.cpp
```diff
@@ -111,294 +113,132 @@ void Ekf::fuseOptFlow()
...
+	_flow_innov_var(0) = (HK25*HK43*HK46 + HK33*HK43*HK45 + HK37*HK43*HK44 + HK38*HK42*HK43 + HK39*HK43*HK49 + HK40*HK43*HK47 + HK41*HK43*HK48 + R_LOS);
+
+	float HK50;
+	if (_flow_innov_var(0) > R_LOS) {
+		HK50 = HK4/_flow_innov_var(0);
+	} else {
+		// we need to reinitialise the covariance matrix and abort this fusion step
+		initialiseCovariance();
+		return;
+	}
```
kamilritz, line 185:
```suggestion
	if (_flow_innov_var(0) < R_LOS) {
		// we need to reinitialise the covariance matrix and abort this fusion step
		initialiseCovariance();
		return;
	}
	const float HK50 = HK4/_flow_innov_var(0);
```

EKF/optflow_fusion.cpp
```diff
@@ -111,294 +113,132 @@ void Ekf::fuseOptFlow()
 		return;
 	}
 
+	// The derivation allows for an arbitrary body to flow sensor frame rotation which is
+	// currently not supported by the EKF, so assume sensor frame is aligned with the
+	// body frame
+	Dcmf Tbs;
+	Tbs.identity();
```
kamilritz, line 120:
```suggestion
	const Dcmf Tbs = matrix::eye<float, 3>();
```

EKF/optflow_fusion.cpp
```diff
@@ -111,294 +113,132 @@ void Ekf::fuseOptFlow()
...
+	// calculate innovation variance for Y axis observation and protect against a badly conditioned calculation
+	_flow_innov_var(1) = (HK43*HK74*HK90 + HK43*HK77*HK89 + HK43*HK79*HK88 + HK43*HK80*HK87 + HK66*HK92*HK94 + HK68*HK91*HK92 + HK70*HK92*HK93 + R_LOS);
+	float HK95;
+	if (_flow_innov_var(1) >= R_LOS) {
+		HK95 = HK4/_flow_innov_var(1);
+	} else {
+		// we need to reinitialise the covariance matrix and abort this fusion step
+		initialiseCovariance();
+		return;
 	}
```
kamilritz, line 242:
```suggestion
	if (_flow_innov_var(1) < R_LOS) {
		// we need to reinitialise the covariance matrix and abort this fusion step
		initialiseCovariance();
		return;
	}
	const float HK95 = HK4/_flow_innov_var(1);

```

EKF/optflow_fusion.cpp
```diff
@@ -63,8 +65,8 @@ void Ekf::fuseOptFlow()
 	// calculate the optical flow observation variance
 	const float R_LOS = calcOptFlowMeasVar();
 
-	float H_LOS[2][24] = {}; // Optical flow observation Jacobians
-	float Kfusion[24][2] = {}; // Optical flow Kalman gains
+	SparseVector24f<0,1,2,3,4,5,6> Hfusion; // Optical flow observation Jacobians
```
kamilritz, line 68: Would move those declarations down to the place before it is used for the first time.

PR #890, opened 2020-08-13
EKF: Convert airspeed fusion to use SymPy generated auto-code

priseborough:
Replaces matlab symbolic toolbox generated equations with equivalents generated using the SymPy library.

The following innovation and innovation variance sequence was generated from replay of a gazebo_plane SITL log. The sequences are visually identical.

Before:

before

After:

after

Output from the test program comparing the Jacobians and Kalman gain values from the Matlab and SymPy method is shown below. Differences are within the expect range for single precision operations.

Pass: Airspeed Hfusion max diff fraction = 0.000000e+00
Pass: Airspeed Kfusion max diff fraction = 3.226521e-07

### DIFF
--- a/EKF/airspeed_fusion.cpp
+++ b/EKF/airspeed_fusion.cpp
@@ -47,9 +47,6 @@
 
 void Ekf::fuseAirspeed()
 {
-	SparseVector24f<4,5,6,22,23> Hfusion; // Observation Jacobians
-	Vector24f Kfusion; // Kalman gain vector
-
 	const float &vn = _state.vel(0); // Velocity in north direction
 	const float &ve = _state.vel(1); // Velocity in east direction
 	const float &vd = _state.vel(2); // Velocity in downwards direction
@@ -67,12 +64,12 @@ void Ekf::fuseAirspeed()
 	const float HK0 = vn - vwn;
 	const float HK1 = ve - vwe;
 	const float HK2 = ecl::powf(HK0, 2) + ecl::powf(HK1, 2) + ecl::powf(vd, 2);
-	if (HK2 < 1.0f) {
+	const float v_tas_pred = sqrtf(HK2); // predicted airspeed
+	//const float HK3 = powf(HK2, -1.0F/2.0F);
+	if (v_tas_pred < 1.0f) {
 		// calculation can be badly conditioned for very low airspeed values so don't fuse this time
 		return;
 	}
-	const float v_tas_pred = sqrtf(HK2); // predicted airspeed
-	//const float HK3 = powf(HK2, -1.0F/2.0F);
 	const float HK3 = 1.0f / v_tas_pred;
 	const float HK4 = HK0*HK3;
 	const float HK5 = HK1*HK3;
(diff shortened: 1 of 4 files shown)


EKF/airspeed_fusion.cpp
```diff
@@ -46,152 +47,137 @@
...
+	const float HK12 = HK0*P(4,4) - HK1*P(4,23) + HK1*P(4,5) - HK11 + P(4,6)*vd;
+	const float HK13 = HK0*HK6;
+	const float HK14 = -HK0*P(22,23) + HK0*P(4,23) - HK1*P(23,23) + HK8 + P(6,23)*vd;
+	const float HK15 = -HK0*P(22,22) - HK1*P(22,23) + HK1*P(5,22) + HK11 + P(6,22)*vd;
+	//const float HK16 = HK3/(-HK10*HK14 + HK10*HK9 + HK12*HK13 - HK13*HK15 + HK6*HK7*vd + R_TAS);
+	float HK16;
+
+	// innovation variance
+	_airspeed_innov_var = (-HK10*HK14 + HK10*HK9 + HK12*HK13 - HK13*HK15 + HK6*HK7*vd + R_TAS);
+	if (_airspeed_innov_var >= R_TAS) { // Check for badly conditioned calculation
```
kamilritz, line 94: It is a bit cleaner if there is only a single `if (_airspeed_innov_var < R_TAS)` case which returns.

EKF/airspeed_fusion.cpp
```diff
@@ -46,152 +47,137 @@
...
-
-			return;
-		}
+	// determine if we need the sideslip fusion to correct states other than wind
+	const bool update_wind_only = !_is_wind_dead_reckoning;
+
+	// Intermediate variables
+	const float HK0 = vn - vwn;
+	const float HK1 = ve - vwe;
+	const float HK2 = ecl::powf(HK0, 2) + ecl::powf(HK1, 2) + ecl::powf(vd, 2);
```
kamilritz, line 69: The if-statement would be a bit clearer when the variable `v_tas_speed` would be used

EKF/airspeed_fusion.cpp
```diff
@@ -46,152 +47,137 @@
...
 
-			fixCovarianceErrors(true);
+	} else {
+		_innov_check_fail_status.flags.reject_airspeed = false;
+	}
 
-			// apply the state corrections
-			fuse(Kfusion, _airspeed_innov);
+	// Airspeed measurement sample has passed check so record it
+	_time_last_arsp_fuse = _time_last_imu;
```
kamilritz, line 159: The  `_time_last_arsp_fuse ` is also set below. I argue that we should only update it once it passes also the covariance health checks.

EKF/airspeed_fusion.cpp
```diff
@@ -46,152 +47,137 @@
 
 void Ekf::fuseAirspeed()
 {
-	const float vn = _state.vel(0); // Velocity in north direction
-	const float ve = _state.vel(1); // Velocity in east direction
-	const float vd = _state.vel(2); // Velocity in downwards direction
-	const float vwn = _state.wind_vel(0); // Wind speed in north direction
-	const float vwe = _state.wind_vel(1); // Wind speed in east direction
+	SparseVector24f<4,5,6,22,23> Hfusion; // Observation Jacobians
```
kamilritz, line 50: This could be declared closer to the first usage.

priseborough:
The SITL log replay was repeated with the same results:

Screen Shot 2020-08-15 at 9 27 20 am

PR #892, opened 2020-08-15
Random cleanup

kamilritz:


priseborough:
The states for the main filter will get picked up in the change indicator, but I am concerned about the the yaw estimator test coverage for non-functional changes.

kamilritz:
@priseborough do you have a replay-able log, where a reset to the ekf-gsf filter is performed. Then I could add a similar change indicator for that.

priseborough:
> @priseborough do you have a replay-able log, where a reset to the ekf-gsf filter is performed. Then I could add a similar change indicator for that.

https://logs.px4.io/plot_app?log=e8f88236-3e39-4705-9d62-36a1072b0fd6

In the process of generating the log using Firmware and ecl master i noticed that the faster immediate post takeoff reset to the yaw estimator value is not activating with the result that the  commander  failsafe is activating before the yaw is reset and the EKF recovers. This change in behaviour has been introduced previously and is unrelated to this fix, so will fix in a separate PR.

priseborough:
After further investigation the issue of the delayed yaw reset appears to be due to the gazebo simulation's lower horizontal velocity during a bad yaw flyaway compared to a real vehicle. The log provided should be OK for testing the EKF-GSF operation and reset of the main filter to the EKF-GSF yaw.

priseborough:
Can we get this rebased so that the ekf-gsf change indicator test is included

PR #894, opened 2020-08-16
EKF: convert drag fusion from Matlab to SymPy generated auto-code

priseborough:
Before and after innovations, innovation variances and wind state estimates from replay of gazebo SITL log data with EKF2_AID_MASK set to 33 to activate drag fusion.

Before:
before

After:
after

Output from the auto-code comparison test program shows expected amount of difference between Matlab and SyPmy generated code:

Pass: Specific Force X axis Hfusion max diff fraction = 0.000000e+00
Pass: Specific Force X axis Kfusion max diff fraction = 1.161862e-06
Pass: Specific Force Y axis Hfusion max diff fraction = 0.000000e+00
Pass: Specific Force Y axis Kfusion max diff fraction = 2.331720e-06

### DIFF
--- a/EKF/drag_fusion.cpp
+++ b/EKF/drag_fusion.cpp
@@ -60,19 +60,19 @@ void Ekf::fuseDrag()
 	const Vector2f ballistic_coef_inv_xy(1.f / _params.bcoef_x, 1.f / _params.bcoef_y);
 
 	// get latest estimated orientation
-	const float q0 = _state.quat_nominal(0);
-	const float q1 = _state.quat_nominal(1);
-	const float q2 = _state.quat_nominal(2);
-	const float q3 = _state.quat_nominal(3);
+	const float &q0 = _state.quat_nominal(0);
+	const float &q1 = _state.quat_nominal(1);
+	const float &q2 = _state.quat_nominal(2);
+	const float &q3 = _state.quat_nominal(3);
 
 	// get latest velocity in earth frame
-	const float vn = _state.vel(0);
-	const float ve = _state.vel(1);
-	const float vd = _state.vel(2);
+	const float &vn = _state.vel(0);
+	const float &ve = _state.vel(1);
+	const float &vd = _state.vel(2);
 
 	// get latest wind velocity in earth frame
-	const float vwn = _state.wind_vel(0);
-	const float vwe = _state.wind_vel(1);
+	const float &vwn = _state.wind_vel(0);
+	const float &vwe = _state.wind_vel(1);
 
 	// predicted specific forces
 	// calculate relative wind velocity in earth frame and rotate into body frame
--- a/EKF/drag_fusion.cpp
+++ b/EKF/drag_fusion.cpp
@@ -48,8 +48,8 @@ void Ekf::fuseDrag()
 {
 	SparseVector24f<0,1,2,3,4,5,6,22,23> Hfusion;  // Observation Jacobians
 	Vector24f Kfusion; // Kalman gain vector
-	const float R_ACC = sq(_params.drag_noise); // observation noise variance in specific force drag (m/sec**2)**2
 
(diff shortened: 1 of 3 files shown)


EKF/drag_fusion.cpp
```diff
@@ -88,114 +86,173 @@ void Ekf::fuseDrag()
 		const float mea_acc = _drag_sample_delayed.accelXY(axis_index)  - _state.delta_vel_bias(axis_index) / _dt_ekf_avg;
 		const float airSpd = sqrtf((2.0f * fabsf(mea_acc)) / (ballistic_coef_inv_xy(axis_index) * rho));
 
+		// Estimate the derivative of specific force wrt airspeed along the X axis
```
kamilritz, line 89: Do you want to make the above introduced variables vn,ve,vd,vwn,vwe a reference as with the other fusion types?

EKF/drag_fusion.cpp
```diff
@@ -45,12 +46,9 @@
...
 void Ekf::fuseDrag()
 {
-	float SH_ACC[4] = {}; // Variable used to optimise calculations of measurement jacobian
-	Vector24f H_ACC; // Observation Jacobian
-	float SK_ACC[9] = {}; // Variable used to optimise calculations of the Kalman gain vector
+	SparseVector24f<0,1,2,3,4,5,6,22,23> Hfusion;  // Observation Jacobians
 	Vector24f Kfusion; // Kalman gain vector
-	// TODO: resolve variance vs stdDev bug
-	const float R_ACC = _params.drag_noise; // observation noise variance in specific force drag (m/sec**2)**2
+	const float R_ACC = sq(_params.drag_noise); // observation noise variance in specific force drag (m/sec**2)**2
```
kamilritz, line 51: Variables with the ending `_noise` are usually standard deviations, so taking the square of it for the variance is the right thing to do. But when adding the square, we should adapt the description and the default value of `drag_noise` in common.h, which is currently
`	float drag_noise{2.5f};			///< observation noise variance for drag specific force measurements (m/sec**2)**2
` to
`	float drag_noise{1.6f};			///< observation noise for drag specific force measurements (m/sec**2)
`

priseborough: Yes, that was a error carried over from previous. Will fix.

priseborough: Done

priseborough:
Here's the replay result with the fix to the R_ACC value. The slightly slower wind state convergence for the SymPy vs Matlab result has been eliminated.

Screen Shot 2020-08-16 at 5 01 36 pm

priseborough:
> We should change the `drag_noise` param description in the common.h file as well

Description has correct units. Variable name could be better, eg drag_noise_variance, but that would also require a change to the ekf2 module and the declaration comments and parameter description are consistent and correct so would prefer to leave alone.

kamilritz:
drag_noise (as standard deviation) described [here](https://github.com/PX4/ecl/blob/master/EKF/common.h#L349) can not have the same units as sq(drag_noise) (as variance). I would argue it should be adapted in common.h and ekf2 params.

priseborough:
> drag_noise (as standard deviation) described [here](https://github.com/PX4/ecl/blob/master/EKF/common.h#L349) can not have the same units as sq(drag_noise) (as variance). I would argue it should be adapted in common.h and ekf2 params.

It's not being squared (it is copied across to R_ACC), the term 'standard deviation' is not mentioned anywhere in relation to this parameter and it is consistently described as a 'observation noise variance' in the module and ecl library so I don't understand what the issue is.

kamilritz:
> It's not being squared (it is copied across to R_ACC), the term 'standard deviation' is not mentioned anywhere in relation to this parameter and it is consistently described as a 'observation noise variance' in the module and ecl library so I don't understand what the issue is.

Ohh, my apologies. I was glimpsing over it too fast. I think we do the square on all other noise variables.

PR #895, opened 2020-08-17
Add change indication for ekf_gsf_yaw_reset

kamilritz:
This following plot shows the quaternion state in the change indication output. The reset can be seen clearly.
Screenshot from 2020-08-17 18-12-54


priseborough:
Why is the change indicator failing?

kamilritz:
> Why is the change indicator failing?

The change indication is comparing the change indication output between current master and this PR. Since we added a new output file in this PR, there is a diff. So it is okay to merge.
Screenshot from 2020-08-18 09-06-24

PR #900, opened 2020-08-22
Make yaw_jacobian of type SparseVector24f<0,1,2,3>

kamilritz:
so we can use the generic computeKHP function. Review best commit by commit.

### DIFF
--- a/EKF/mag_fusion.cpp
+++ b/EKF/mag_fusion.cpp
@@ -639,24 +639,9 @@ void Ekf::updateQuaternion(const float innovation, const float measurement_varia
 		_heading_innov = innovation;
 	}
 
-	// apply covariance correction via P_new = (I -K*H)*P
-	// first calculate expression for KHP
-	// then calculate P - KHP
-	const SquareMatrix24f KHP = computeKHP(Kfusion, yaw_jacobian);
+	const bool is_fused = measurementUpdate(Kfusion, yaw_jacobian, _heading_innov);
 
-	const bool healthy = checkAndFixCovarianceUpdate(KHP);
-
-	_fault_status.flags.bad_hdg = !healthy;
-
-	if (healthy) {
-		// apply the covariance corrections
-		P -= KHP;
-
-		fixCovarianceErrors(true);
-
-		// apply the state corrections
-		fuse(Kfusion, _heading_innov);
-	}
+	_fault_status.flags.bad_hdg = !is_fused;
 }
 
 void Ekf::fuseHeading()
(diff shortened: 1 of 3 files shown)


EKF/mag_fusion.cpp
```diff
@@ -601,21 +601,14 @@ void Ekf::updateQuaternion(const float innovation, const float variance, const f
 		float tmp = 0.0f;
 
 		for (uint8_t col = 0; col <= 3; col++) {
-			tmp += P(row,col) * yaw_jacobian(col);
+			tmp += P(row,col) * yaw_jacobian.atCompressedIndex(col);
```
jkflying, line 604: The `atCompressedIndex` is mixing sparse and dense addressing. It technically works _here_, but only because the indices are [0,1,2,3]. Better to not get into this pattern. I see 2 options:

1. Express this nested loop as a matrix operation instead. I think that is the direction we should go for this kind of thing, and add any needed matrix-style operations to the SparseVector class. This would be something like `_heading_innov_var += yaw_jacobian.transpose() * P * yaw_jacobian;` , although maybe we want to make a helper function called something like `mult_btAb(Matrix A, SparseVector b)` to avoid re-implementing Eigen's entire compile-time temporary system.

2. If it really can't be a matrix operation (or we can't implement that operation for some reason), then we need to correctly translate between sparse and dense indices. The loops should iterate over the compressed indices, eg: `for (unsigned compressed_row = 0; compressed_row < yaw_jacobian.non_zeros(); compressed_row++)`, and accesses to P need to be translated into non-compressed indices:
```suggestion
			tmp += P(yaw_jacobian.index(compressed_row), yaw_jacobian.index(compressed_col)) * yaw_jacobian.atCompressedIndex(compressed_col);
```

jkflying: Note that option 2 is basically what would be inside of the helper function described in option 1.

EKF/mag_fusion.cpp
```diff
@@ -578,30 +578,14 @@ void Ekf::fuseYaw312(float yaw, float yaw_variance, bool zero_innovation)
...
+	_heading_innov_var = matrix::quadraticForm(P, yaw_jacobian) + measurement_variance;
 
 	// check if the innovation variance calculation is badly conditioned
-	if (_heading_innov_var >= variance) {
-		// the innovation variance contribution from the state covariances is not negative, no fault
-		_fault_status.flags.bad_hdg = false;
-		heading_innov_var_inv = 1.0f / _heading_innov_var;
-
-	} else {
+	if (_heading_innov_var < measurement_variance) {
```
jkflying, line 588: If seems this is a functional change?

jkflying (dismissed):

bresch (dismissed):
Nice

PR #906, opened 2020-09-02
WindEstimator: Fix incorrect _state accessing, clarify enums

rjgritter:
The WindEstimator class uses an enum to index into its _state vector. However, in a few places (lines 166, 170, 171, 239 and 241), it mistakenly uses those enums as if they are variables representing the states, entirely corrupting the validity of the math.  

This change fixes that and also renames the enums using a style less likely to be mistaken for local variables in the future.


kamilritz (approved):
Thanks.

sfuhrer (approved):

PR #908, opened 2020-09-21
EKF: use GPS to lookup declination from WMM before full GPS checks pass

dagar:
Long before the full GPS checks pass the fix is sufficient to lookup the magnetic declination from the geomagnetic tables. The tables only have data every 10 degrees (~1000 km) and the change is gradual.

This PR adds a fallback in GPS collection before both NED origin is initialized and GPS checks pass to set the declination (_mag_declination_gps) as soon as there's at least a 2D fix with eph < 1 km. When GPS checks finally do pass the declination is set again from that lat/lon.


priseborough (dismissed):
Tested on SITL gazebo.

dagar:
Thanks for the review, there are a small number of test failures to resolve.

Screenshot from 2020-09-24 08-19-39

Screenshot from 2020-09-24 08-20-11

dagar:
The GPS heading test failures look real, but I haven't yet figured out why having a mag declination sooner results in a (small) yaw estimate error in this scenario.

dagar:
The GPS heading convergence tests start passing again when the sensor simulator time is extended.

dagar:
Tests are now passing. I've added a mag yaw reset request the first time we get a mag declination and also skip the reset once GPS checks are passing if there's already a valid mag declination.

bresch (dismissed):
I went through the modification and ran the unit tests without `[DO NOT MERGE] test EKF gps yaw changes for discussion` and It's all good. Can you remove this commit?

dagar:
> I went through the modification and ran the unit tests without `[DO NOT MERGE] test EKF gps yaw changes for discussion` and It's all good. Can you remove this commit?

It was reverted in the followup, but I'll squash it in place now for simplicity.

bresch (approved):

PR #912, opened 2020-09-25
ekf: fix variable length array error

bresch:
Even if the non_zeros() function is static constexpr, gcc (10.2) does not like it.
Using sizeof...(Idxs) fixes the error.

Thanks @jkflying for the help.


jkflying (approved):
FYI @kamilritz

kamilritz (approved):

PR #913, opened 2020-09-28
range_finder: parametrize range sensor quality hysteresis time

cmic0:
This will allow to adjust the hysteresis time based on the distance sensor performances.

For distance sensors that can't read high distances (i.e. >10m) waiting 1s or more before fusing the data might lead to hard landings during auto descends.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -124,12 +124,13 @@ void Ekf::controlFusionModes()
 	// Get range data from buffer and check validity
 	const bool is_rng_data_ready = _range_buffer.pop_first_older_than(_imu_sample_delayed.time_us, _range_sensor.getSampleAddress());
 	_range_sensor.setDataReadiness(is_rng_data_ready);
-	_range_sensor.runChecks(_imu_sample_delayed.time_us, _R_to_earth);
 
 	// update range sensor angle parameters in case they have changed
 	_range_sensor.setPitchOffset(_params.rng_sens_pitch);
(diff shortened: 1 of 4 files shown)


dagar:
How low do you want to go? I'm wondering if the initial value of 1 second is arbitrary in the first place.

Another thing you could consider is adjusting it automatically, the EKF already carries the sensor metadata (min/max distance).

Wherever possible I'd like to try and avoid parameter bloat if we can get away with it.

bresch:
The problem is that the quality of the sensor data does not correlate with its range. To solve @cmic0 's issue, we should lower the threshold given the short maximum distance, but there are also a lot of crappy range finders with short range that would need a longer hysteresis time. On the opposite side, there are also good range finder with a large max distance that don't require a large hysteresis.
So I think the other alternative to having a parameter for this in ekf2 is to move the check in the driver of each range finder, such that it is tuned for a specific sensor model.

dagar (approved):

EKF/control.cpp
```diff
@@ -129,6 +129,7 @@ void Ekf::controlFusionModes()
 	// update range sensor angle parameters in case they have changed
 	_range_sensor.setPitchOffset(_params.rng_sens_pitch);
 	_range_sensor.setCosMaxTilt(_params.range_cos_max_tilt);
+	_range_sensor.setQualityHysteresis(_params.range_valid_quality_s);
```
dagar, line 132: These should probably be updated before the checks are run (_range_sensor.runChecks() above).

bresch (approved):

PR #914, opened 2020-10-04
EKF: Improve robustness of yaw reset to bad inertial data

priseborough:
Preliminary response to the corner case highlighted by https://discuss.px4.io/t/estimator-failure/18790

Testing to date limited to a basic test in SITL with the mag data rotated 180 deg to induce a flyaway, yaw reset and recovery sequence - https://review.px4.io/plot_app?log=b9381141-21c1-4401-af2f-4639690dea49

EKF-GSF convergence
Screen Shot 2020-10-04 at 8 38 05 pm

Further testing is required including a way to simulate the clipping behaviour case encountered.

Another PR forthe PX4/Firmware ekf2 module to change setting of  _ekf.set_in_air_status(!_vehicle_land_detected.landed) is recommended to ensure the EKF-GSF activates earlier.


priseborough:
TODO - update comparison file for reversion diff test.

EKF/control.cpp
```diff
@@ -580,16 +581,28 @@ void Ekf::controlGpsFusion()
...
 
+			if ((recent_takeoff_nav_failure || inflight_nav_failure) && _time_last_hor_vel_fuse > 0) {
+				// Do sanity check to see if the innovation failures is likely caused by a yaw angle error
+				// A 180 deg yaw error would (in the absence of other errors) generate a velocity innovation
+				// magnitude of twice the measured speed so allow up to 3 times to allow for additional errors.
+				const float hvel_obs_sq = sq(_last_vel_obs(0)) + sq(_last_vel_obs(1));
+				const float hvel_innov_sq = sq(_last_fail_hvel_innov(0)) + sq(_last_fail_hvel_innov(1));
+				if (hvel_innov_sq > 9.0f * hvel_obs_sq) {
+					recent_takeoff_nav_failure = false;
+					inflight_nav_failure = false;
```
bresch, line 602: As described in the comment, this check is the answer to "is the nav issues due to a yaw error?", so IMO, instead of changing `recent_takeoff_nav_failure` and `inflight_nav_failure`, we should rather add a new boolean `is_yaw_failure` and replace the check below:
```c++
const bool do_yaw_vel_pos_reset = (_do_ekfgsf_yaw_reset || recent_takeoff_nav_failure || inflight_nav_failure) && ...
```
So the meaning of the variables isn't changed.
by 
```c++
const bool do_yaw_vel_pos_reset = (_do_ekfgsf_yaw_reset || is_yaw_failure) && ...
```

EKF/EKFGSF_yaw.cpp
```diff
@@ -79,14 +79,23 @@ void EKFGSF_yaw::update(const imuSample& imu_sample,
 			if (!bad_update) {
 				float total_weight = 0.0f;
 				// calculate weighting for each model assuming a normal distribution
+				const float min_weight = 1E-5f;
```
bresch, line 82:
```suggestion
				const float min_weight = 1e-5f;
```

bresch (changes requested):

EKF/control.cpp
priseborough: Done

EKF/EKFGSF_yaw.cpp
priseborough: Changed in multiple places for consistency

bresch (approved):
Ok, looks good now. However, why is this changing the output of the standard change indicator? If there is no emergency yaw reset, nothing should change.

priseborough:
I can't explain that. Maybe it's the compiler version. I'm building from a Ubuntu 20.04 machine.

priseborough:
There is a time delay between the landed flag going false and the  EKF-GSF logging starting, see below. Thinking about this some more, some of the delay can be explained by the fact that the EKF-GSF yaw estimator is running on the same delayed fusion time horizon as the main navigation EKF loop which can introduce a delay of up to 200msec. A further 200msec can be explained by the 200Hz logging rate for EKF-GSF data and a further 200msec by the GPS update rate as the EKF-GSF won't initialise until the next fresh GPS sample is available, then has to wait a further 200msec for the next GPS measurement which explains the majority of the delay.

No further  investigation of options for setting theEKF's  in-air status earlier is warranted at this time. 

Screen Shot 2020-10-06 at 8 15 59 pm

Issue #909, opened 2020-09-21
EKF minimize console output


dagar:
Can we work on minimizing console output for the estimator? It gets a little absurd when running multiple (https://github.com/PX4/Firmware/pull/14650).

``` Console
INFO  [ecl/EKF] reset position to last known position
INFO  [ecl/EKF] reset velocity to zero
INFO  [ecl/EKF] reset position to last known position
INFO  [ecl/EKF] reset velocity to zero
INFO  [ecl/EKF] reset position to last known position
INFO  [ecl/EKF] reset velocity to zero
INFO  [ecl/EKF] reset position to last known position
INFO  [ecl/EKF] reset velocity to zero
INFO  [ecl/EKF] reset position to last known position
INFO  [ecl/EKF] reset velocity to zero
INFO  [ecl/EKF] reset position to last known position
INFO  [ecl/EKF] reset velocity to zero
INFO  [ecl/EKF] 11544067: EKF aligned, (baro hgt, IMU buf: 18, OBS buf: 14)
INFO  [ecl/EKF] 11644085: EKF aligned, (baro hgt, IMU buf: 18, OBS buf: 14)
INFO  [ecl/EKF] 11751436: EKF aligned, (baro hgt, IMU buf: 18, OBS buf: 14)
INFO  [ecl/EKF] 11813268: EKF aligned, (baro hgt, IMU buf: 18, OBS buf: 14)
INFO  [ecl/EKF] 11951770: EKF aligned, (baro hgt, IMU buf: 18, OBS buf: 14)
INFO  [ecl/EKF] 11973531: EKF aligned, (baro hgt, IMU buf: 18, OBS buf: 14)
```

I would propose we (incrementally) eliminate all console output by default, ensuring the equivalent information is trivial to find in the logging. Critical in flight errors and resets could still make sense if used sparingly, but even there we'll want to make sure it's again captured by logging and perhaps even handled higher level (ekf2_main) to produce a mavlink critical message rather than console noise. 

FYI @priseborough @bresch

dagar:
We also need to be careful with messages that have the potential to flood the console.

``` Console
ERROR [ecl/EKF] airspeed badly conditioned - wind covariance reset
ERROR [ecl/EKF] airspeed badly conditioned - wind covariance reset
ERROR [ecl/EKF] airspeed badly conditioned - wind covariance reset
ERROR [ecl/EKF] airspeed badly conditioned - wind covariance reset
[...]

PR #917, opened 2020-10-21
EKF: add fault status bit for clipping

dagar:
Does this make sense as a fault status bit? We don't necessarily have all the input data logged at full rate so I wanted to capture this somewhere within ecl/EKF status.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -401,7 +401,8 @@ union fault_status_u {
 		bool bad_pos_E: 1;	///< 13 - true if fusion of the East position has encountered a numerical error
 		bool bad_pos_D: 1;	///< 14 - true if fusion of the Down position has encountered a numerical error
 		bool bad_acc_bias: 1;	///< 15 - true if bad delta velocity bias estimates have been detected
-		bool bad_acc_clipping: 1; ///< 16 - true if delta velocity data contains clipping (asymmetric railing)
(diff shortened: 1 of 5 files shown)


dagar:
Relevant for multi-ekf, where we could switch the primary estimator immediately rather than waiting for the effects to accumulate in the comparison metrics (eg https://github.com/PX4/Firmware/pull/14650#issuecomment-713292307).

bresch:
Sounds like a good idea. Since you're already changing the status flags, can you also add `_bad_vert_accel_detected` in it? https://github.com/PX4/ecl/blob/f666ebb99552eed796e82a20bb60e813d45eee03/EKF/control.cpp#L902

dagar:
Yes I'll add that as well. 

We're already capturing `_bad_vert_accel_detected` as ekf_solution_status.accel_error, but that entire message seems to only exist for being passed through to Mavlink. 

https://github.com/PX4/ecl/blob/dd3ffc419238bfc7d9b7e3be92a178c2cf7756f1/EKF/common.h#L495

http://mavlink.io/en/messages/common.html#ESTIMATOR_STATUS

Screenshot from 2020-10-26 10-25-14

We should really give this an overall review at some point, including killing off most of the console messages (https://github.com/PX4/ecl/issues/909).

dagar:
Added `bad_acc_vert` to fault status.

bresch (approved):
Thanks!

PR #918, opened 2020-10-22
allow reset heading for initializing with no mag fusion

moreba1:


priseborough:
Please provide us with some context for this change:

What problem are you trying to solve.
How have you tested this and can you provide test evidence.

PR #920, opened 2020-10-26
Optical flow: compute velociy from corrected flow data for logging

bresch:
This is something I always needed when working with optical flow sensors. It allows to easily fine tune the delay and scale factor of the optical flow measurements to get the best possible gyro compensation.
It's also extremely useful to see if the measurements make sense because looking at flow innovations isn't really intuitive.

Here's an example of what you can get in the log (I'll make the corresponding firmware PR to add the logging part):

2020-10-23_12-30-04_01_plot

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -107,6 +107,9 @@ public:
 
 	void getFlowVelBody(float flow_vel_body[2]) const override;
 	void getFlowVelNE(float flow_vel_ne[2]) const override;
+	void getFlowCompensated(float flow_compensated[2]) const override;
+	void getFlowUncompensated(float flow_uncompensated[2]) const override;
+	void getFlowGyro(float flow_gyro[3]) const override;
 
 	void getHeadingInnov(float &heading_innov) const override;
 
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -105,6 +105,9 @@ public:
 
 	void getFlowInnovRatio(float &flow_innov_ratio) const override;
 
+	void getFlowVelBody(float flow_vel_body[2]) const override;
+	void getFlowVelNE(float flow_vel_ne[2]) const override;
+
 	void getHeadingInnov(float &heading_innov) const override;
 
 	void getHeadingInnovVar(float &heading_innov_var) const override;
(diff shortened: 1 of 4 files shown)


EKF/ekf.h
```diff
@@ -105,6 +105,9 @@ class Ekf : public EstimatorInterface
 
 	void getFlowInnovRatio(float &flow_innov_ratio) const override;
 
+	void getFlowVelBody(float flow_vel_body[2]) const override;
```
kamilritz, line 108: I would prefer to return a `Vector2f` and do the copyTo on the Firmware side. This way if we want to use these functions in unit tests, they are easier to handle.

kamilritz (changes requested):
Cool functionality. Thanks a lot!

EKF/ekf.h
bresch: Ok, yes, it makes sense; I just copied the existing pattern

bresch: @kamilritz Done

EKF/ekf_helper.cpp
```diff
@@ -708,6 +708,31 @@ void Ekf::getFlowInnovRatio(float &flow_innov_ratio) const
...
+}
+
+Vector2f Ekf::getFlowCompensated() const
+{
+	return _flow_compensated_XY_rad;
+}
+
+Vector2f Ekf::getFlowUncompensated() const
+{
+	return _flow_sample_delayed.flow_xy_rad;
```
kamilritz, line 728: Is it necessary to expose the incoming sample?

EKF/estimator_interface.h
```diff
@@ -91,6 +91,11 @@ class EstimatorInterface
 	virtual void getFlowInnov(float flow_innov[2]) const = 0;
 	virtual void getFlowInnovVar(float flow_innov_var[2]) const = 0;
 	virtual void getFlowInnovRatio(float &flow_innov_ratio) const = 0;
+	virtual Vector2f getFlowVelBody() const = 0;
+	virtual Vector2f getFlowVelNE() const = 0;
+	virtual Vector2f getFlowCompensated() const = 0;
```
kamilritz, line 96: Are those all really needed?

EKF/ekf_helper.cpp
bresch: Well, this is the "delta angle" synchronized and integrated to the optical flow update period. It is not just a pure gyro measurement and is only an "incoming sample" if the optical flow message provides gyro data, which is often not the case.

EKF/estimator_interface.h
bresch: I don't think so; the getters could be directly in the estimator interface, but I preferred having them next to the `getFlowInnov...`. We could then move them to the estimator interface altogether in a separate PR.

EKF/ekf_helper.cpp
bresch: Actually, thanks to that logging, I found out that the optical flow scaling is off by quite a bit.

kamilritz (approved):

EKF/ekf_helper.cpp
```diff
@@ -708,6 +708,31 @@ void Ekf::getFlowInnovRatio(float &flow_innov_ratio) const
 	flow_innov_ratio = _optflow_test_ratio;
 }
 
+Vector2f Ekf::getFlowVelBody() const
+{
+	return _flow_vel_body;
```
dagar, line 713: Simple one liners like this can be kept in the header to allow for potential inlining across translation units (ie ekf2 calling ecl/EKF).

bresch: Yes, will do that next time

PR #922, opened 2020-10-26
EKF inline simple getters and respect MAX_CUSTOM_OPT_LEVEL if set by parent

dagar:


EKF/estimator_interface.h
```diff
@@ -91,11 +91,11 @@ class EstimatorInterface
 	virtual void getFlowInnov(float flow_innov[2]) const = 0;
 	virtual void getFlowInnovVar(float flow_innov_var[2]) const = 0;
 	virtual void getFlowInnovRatio(float &flow_innov_ratio) const = 0;
-	virtual Vector2f getFlowVelBody() const = 0;
-	virtual Vector2f getFlowVelNE() const = 0;
-	virtual Vector2f getFlowCompensated() const = 0;
-	virtual Vector2f getFlowUncompensated() const = 0;
-	virtual Vector3f getFlowGyro() const = 0;
+	virtual const Vector2f &getFlowVelBody() const = 0;
```
dagar, line 94: @bresch / @kamilritz / @jkflying is the high level intended design of ecl/EKF with `EstimatorInterface` clear to any of you? At the moment it mainly seems like a good mechanism for introducing a huge number of unnecessary virtual calls.

bresch: I guess the initial intent is clear and yes, most variables are already declared in the interface and the getters/setters would not need to be pure virtual and then implemented in EKF

dagar: Currently there's no need for any of it because the consumer (eg PX4 ekf2) has an `Ekf` object, it's never dealing with an `EstimatorInterface`.

dagar: PR is good to go and separate from this question, although I'd like to consider gutting most of these next.

bresch (approved):

PR #923, opened 2020-10-29
EKF: pass imuSample by const ref

dagar:


EKF/ekf.cpp
```diff
@@ -132,7 +132,7 @@ bool Ekf::update()
 
 	// the output observer always runs
 	// Use full rate IMU data at the current time horizon
-	calculateOutputStates();
+	calculateOutputStates(_newest_high_rate_imu_sample);
```
jkflying, line 135: Any way _newest_high_rate_imu_sample could also be made private so we don't end up in a situation where it's a mix of members and arguments being used?

jkflying (approved):

bresch (approved):

EKF/ekf.cpp
dagar: Yes I'm hoping to get rid of it entirely, I left this as an eyesore reminder.

PR #924, opened 2020-10-29
EKF: add mag bias reset helper and update IMU bias reset to match

dagar:


EKF/ekf.h
```diff
@@ -168,12 +168,13 @@ class Ekf final : public EstimatorInterface
...
 	void get_ekf_ctrl_limits(float *vxy_max, float *vz_max, float *hagl_min, float *hagl_max) override;
 
-	/*
-	Reset all IMU bias states and covariances to initial alignment values.
-	Use when the IMU sensor has changed.
-	Returns true if reset performed, false if rejected due to less than 10 seconds lapsed since last reset.
-	*/
-	bool reset_imu_bias() override;
+	// Reset all IMU bias states and covariances to initial alignment values.
+	void resetIMUBias();
```
kamilritz, line 172:
```suggestion
	void resetImuBias();
```

EKF/ekf_helper.cpp
```diff
@@ -824,27 +824,43 @@ void Ekf::get_ekf_ctrl_limits(float *vxy_max, float *vz_max, float *hagl_min, fl
 
 }
 
-bool Ekf::reset_imu_bias()
+void Ekf::resetIMUBias()
 {
-	if (_imu_sample_delayed.time_us - _last_imu_bias_cov_reset_us < (uint64_t)10e6) {
```
kamilritz, line 829: was this not needed?

EKF/ekf_helper.cpp
```diff
@@ -824,27 +824,43 @@ void Ekf::get_ekf_ctrl_limits(float *vxy_max, float *vz_max, float *hagl_min, fl
...
 
-	return true;
+void Ekf::resetMagBias()
+{
+	// Zero the magnetometer bias states
+	_state.mag_B.zero();
+
+	// Zero the corresponding covariances and set
+	// variances to the values use for initial alignment
+	P.uncorrelateCovarianceSetVariance<3>(19, sq(_params.mag_noise));
```
kamilritz, line 863: Would we need to reset `_saved_mag_bf_variance` as well?

dagar: Sounds like it. 

FYI the use case is either magnetometer calibration updated or the device completely changed mis-use (failured).

EKF/ekf_helper.cpp
dagar: I'd rather be in control of it firmware side and prevent rapid switching.

EKF/ekf_helper.cpp
dagar: Why does `_saved_mag_bf_variance` include the D earth axis?

dagar: https://github.com/PX4/PX4-ECL/blob/48a8992caf7a95e09a5e17235133894c072b96bd/EKF/ekf_helper.cpp#L1346-L1349

dagar: Updated.

EKF/ekf_helper.cpp
kamilritz: ok

EKF/ekf_helper.cpp
kamilritz: I am honestly not sure, what is the right thing to do here. Should we reset it to 0, mag_noise or let it be. @bresch can you help us here?

dagar:
The other reason for this change is to get rid of the cases in the ekf frontend (eg PX4 ekf2) where it's poking into the covariance and states by magic index.

https://github.com/PX4/PX4-Autopilot/blob/817285ec64a56951b8d8db295596bb2ff09171e9/src/modules/ekf2/EKF2.cpp#L982-L995

Screenshot from 2020-11-02 13-03-50


https://github.com/PX4/PX4-Autopilot/blob/817285ec64a56951b8d8db295596bb2ff09171e9/src/modules/ekf2/EKF2.cpp#L912-L920

Screenshot from 2020-11-02 13-03-34

dagar:
Can we move forward with this?

EKF/ekf_helper.cpp
kamilritz: @bresch can you check this?

dagar:
This is harmless as the IMU bias reset logic is equivalent and we aren't yet resetting the mag bias anywhere. Let's go forward and revisit the mag bias reset specifics in detail with actual usage.

PR #928, opened 2020-10-30
flow: restructure optical flow control logic

bresch:
I tried to untangle a bit the flow selection logic to make it more readable. At the same time the `resetToFlowVelocityInAir` unit test got fixed.

fixes https://github.com/PX4/PX4-ECL/issues/926

Currently no SITL/real tests

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -373,7 +373,7 @@ void Ekf::controlOpticalFlowFusion()
 	// Only fuse optical flow if valid body rate compensation data is available
 	if (_flow_data_ready && calcOptFlowBodyRateComp()) {
 
-		bool flow_quality_good = (_flow_sample_delayed.quality >= _params.flow_qual_min);
+		const bool flow_quality_good = (_flow_sample_delayed.quality >= _params.flow_qual_min);
 
 		if (!flow_quality_good && !_control_status.flags.in_air) {
 			// when on the ground with poor flow quality, assume zero ground relative velocity and LOS rate
@@ -404,8 +404,8 @@ void Ekf::controlOpticalFlowFusion()
 
 		// inhibit use of optical flow if motion is unsuitable and we are not reliant on it for flight navigation
 		const bool preflight_motion_not_ok = !_control_status.flags.in_air
-		                                     && (((_imu_sample_delayed.time_us - _time_good_motion_us) > (uint64_t)1E5)
-		                                         || ((_imu_sample_delayed.time_us - _time_bad_motion_us) < (uint64_t)5E6));
+		                                     && ((_imu_sample_delayed.time_us > (_time_good_motion_us + (uint64_t)1E5))
+		                                         || (_imu_sample_delayed.time_us < (_time_bad_motion_us + (uint64_t)5E6)));
 		const bool flight_condition_not_ok = _control_status.flags.in_air && !isTerrainEstimateValid();
 
 		_inhibit_flow_use = ((preflight_motion_not_ok || flight_condition_not_ok) && !is_flow_required)
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -360,14 +360,12 @@ void Ekf::controlOpticalFlowFusion()
 		const bool is_quality_good = (_flow_sample_delayed.quality >= _params.flow_qual_min);
 		const bool is_magnitude_good = !_flow_sample_delayed.flow_xy_rad.longerThan(_flow_sample_delayed.dt * _flow_max_rate);
 		const bool is_tilt_good = (_R_to_earth(2, 2) > _params.range_cos_max_tilt);
-		const bool is_body_rate_comp_available = calcOptFlowBodyRateComp();
 
-		// check if enough integration time and fail if integration time is less than 50%
-		// of min arrival interval because too much data is being lost
-		const float delta_time_min = 0.5e-6f * (float)_min_obs_interval_us;
-		const float delta_time_max = 0.2f;
+		const float delta_time_min = fmaxf(0.8f * _delta_time_of, 0.001f);
+		const float delta_time_max = fminf(1.2f * _delta_time_of, 0.2f);
 		const bool is_delta_time_good = _flow_sample_delayed.dt >= delta_time_min
 		                                && _flow_sample_delayed.dt <= delta_time_max;
+		const bool is_body_rate_comp_available = calcOptFlowBodyRateComp();
 
 		if (is_quality_good
 		    && is_magnitude_good
@@ -378,7 +376,15 @@ void Ekf::controlOpticalFlowFusion()
 			_flow_compensated_XY_rad = _flow_sample_delayed.flow_xy_rad - _flow_sample_delayed.gyro_xyz.xy();
 
 		} else if (!_control_status.flags.in_air && is_body_rate_comp_available) {
-			// when on the ground with poor flow quality, assume zero ground relative velocity and LOS rate
+
+			if (!is_delta_time_good) {
+				// handle special case of SITL and PX4Flow where dt is forced to
+				// zero when the quaity is 0
+				_flow_sample_delayed.dt = delta_time_min;
+			}
+
+			// when on the ground with poor flow quality,
+			// assume zero ground relative velocity and LOS rate
 			_flow_compensated_XY_rad.setZero();
 
 		} else {
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -144,12 +144,13 @@ void Ekf::controlFusionModes()
 	// This means we stop looking for new data until the old data has been fused, unless we are not fusing optical flow,
 	// in this case we need to empty the buffer
 	if (!_flow_data_ready || !_control_status.flags.opt_flow) {
-		_flow_data_ready = _flow_buffer.pop_first_older_than(_imu_sample_delayed.time_us, &_flow_sample_delayed)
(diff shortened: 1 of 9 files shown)


EKF/control.cpp
```diff
@@ -368,54 +368,66 @@ void Ekf::controlOpticalFlowFusion()
...
+		                              && (_is_dead_reckoning // is doing inertial dead-reckoning so must constrain drift urgently
+		                                  || isOnlyActiveSourceOfHorizontalAiding(_control_status.flags.opt_flow)
+		                                  || (_control_status.flags.gps && (_gps_error_norm > gps_err_norm_lim))); // is using GPS, but GPS is bad
+
 
-		// Handle cases where we are using optical flow but are no longer able to because data is old
-		// or its use has been inhibited.
+		// inhibit use of optical flow if motion is unsuitable and we are not reliant on it for flight navigation
+		const bool preflight_motion_not_ok = !_control_status.flags.in_air
+		                                     && (((_imu_sample_delayed.time_us - _time_good_motion_us) > (uint64_t)1E5)
```
kamilritz, line 405: a > c + b instead of a - b > c

EKF/ekf_helper.cpp
```diff
@@ -79,7 +79,7 @@ void Ekf::resetHorizontalVelocityToOpticalFlow()
 	// calculate absolute distance from focal point to centre of frame assuming a flat earth
 	const float range = heightAboveGndEst / _range_sensor.getCosTilt();
 
-	if ((range - _params.rng_gnd_clearance) > 0.3f && _flow_sample_delayed.dt > 0.05f) {
+	if ((range - _params.rng_gnd_clearance) > 0.3f && _flow_sample_delayed.dt < 0.05f) {
```
kamilritz, line 82: Was this always like this? So we reset always to 0?
Is this test really necessary? If yes, shouldn't we use it also for the normal fusion?

kamilritz (commented):
This PR is really nice. The optical flow fusion code has had by far the most complex logic of all sensors. This is really not necessary, and this PR is a good step in the right direction. Now it is much easier to understand.

We will need extensive flight testing as the logic changes quite a bit with this PR.

EKF/ekf_helper.cpp
bresch: Yes, apparently this never worked :D https://github.com/PX4/PX4-ECL/pull/74/files#diff-141e70f9277c0572024e165ddc7b56fffd1b604a06d9239c7410015f3a9ad0c7R89

bresch: > Is this test really necessary? If yes, shouldn't we use it also for the normal fusion?

I would say that it is especially important during initialization to avoid resetting on a bad sample, but yes, this is something that can be done during normal fusion too.

EKF/control.cpp
```diff
@@ -353,23 +354,25 @@ void Ekf::controlOpticalFlowFusion()
 	_imu_del_ang_of += _imu_sample_delayed.delta_ang - _state.delta_ang_bias;
 	_delta_time_of += _imu_sample_delayed.delta_ang_dt;
 
-	// Only fuse optical flow if valid body rate compensation data is available
-	if (_flow_data_ready && calcOptFlowBodyRateComp()) {
+	if (_flow_data_ready) {
+		const bool is_quality_good = (_flow_sample_delayed.quality >= _params.flow_qual_min);
+		const bool is_magnitude_good = !_flow_sample_delayed.flow_xy_rad.longerThan(_flow_sample_delayed.dt * _flow_max_rate);
```
kamilritz, line 359: nice

EKF/ekf_helper.cpp
```diff
@@ -1479,6 +1479,7 @@ void Ekf::stopFlowFusion()
 	_flow_innov.setZero();
 	_flow_innov_var.setZero();
 	_optflow_test_ratio = 0.0f;
+	_flow_data_ready = false;
```
kamilritz, line 1482: This seems a bit problematic to me. We should use the control_status.flag.opt_control where needed.

test/test_EKF_flow.cpp
```diff
@@ -142,3 +142,52 @@ TEST_F(EkfFlowTest, resetToFlowVelocityOnGround)
...
+	_ekf->set_in_air_status(true);
+	_sensor_simulator.runSeconds(5.f);
+
+	const float estimated_distance_to_ground = _ekf->getTerrainVertPos();
+
+	// WHEN: start fusing flow data
+	Vector2f simulated_horz_velocity(0.5f, -0.2f);
+	flowSample flow_sample = _sensor_simulator._flow.dataAtRest();
+	flow_sample.flow_xy_rad =
+		Vector2f( simulated_horz_velocity(1) * flow_sample.dt / estimated_distance_to_ground,
```
kamilritz, line 162: We should create a helper function for this

kamilritz (commented):
Very nice work!

EKF/ekf_helper.cpp
bresch: Yes, that's right. In fact, we already check for the status flag so this should not be needed

test/sensor_simulator/flow.cpp
```diff
@@ -15,7 +15,6 @@ Flow::~Flow()
 
 void Flow::send(uint64_t time)
 {
-	_flow_data.dt = static_cast<float>(time - _time_last_data_sent) * 1e-6f;
```
kamilritz, line 18: why are you removing this?

bresch: Because this is the "integration timespan" of the optical flow sensor and not the publishing rate. My actual issue was that the first sample had a really large dt (a few seconds).

kamilritz: Ok. But then we should set it to 1 / sensor_rate

test/sensor_simulator/flow.cpp
```diff
@@ -29,7 +28,7 @@ void Flow::setData(const flowSample& flow)
 flowSample Flow::dataAtRest()
 {
 	flowSample _flow_at_rest;
-	_flow_at_rest.dt = 0.02f;
+	_flow_at_rest.dt = static_cast<float>(_update_period) * 1e-6f;
```
bresch, line 31: @kamilritz Would it be fine that way? So the user can get the `dataAtRest`, modify `dt` if he wants (generate errors), but the default value is always correct

kamilritz: yes, that is a good option

kamilritz (commented):
Looks good. The helper function for simulating the flow would be a nice add-on. And one of the flow fusion logic test is failing on macOs.

test/sensor_simulator/sensor_simulator.cpp
```diff
@@ -12,8 +12,8 @@ _rng(ekf),
 _vio(ekf),
 _airspeed(ekf)
 {
-	setSensorDataToDefault();
 	setSensorRateToDefault();
+	setSensorDataToDefault();
```
bresch, line 16: @kamilritz This was the issue making macOS unit test fail. In fact, the unit test failed because dt was set to 0 but on Linux, the value was corrected to 0.02 in one of the tests above (-> removing the tests above the flow one triggered the issue). On macOS, it seems that the variables are reset between each unit test (which is good).

test/test_EKF_fusionLogic.cpp
```diff
@@ -198,10 +202,10 @@ TEST_F(EkfFusionLogicTest, doFlowFusion)
 
 	// WHEN: Stop sending flow data
 	_sensor_simulator.stopFlow();
-	_sensor_simulator.runSeconds(10);
+	_sensor_simulator.runSeconds(11);
 
 	// THEN: EKF should not intend to fuse flow measurements
-	EXPECT_TRUE(_ekf_wrapper.isIntendingFlowFusion()); // TODO: change to false
+	EXPECT_FALSE(_ekf_wrapper.isIntendingFlowFusion());
```
bresch, line 208: @kamilritz this is also fixed now (the timeout is 10s)

kamilritz:
This looks good now. Since the logic changed quite a bit, we should have rather extensive flight testing.

EKF/control.cpp
```diff
@@ -358,14 +358,12 @@ void Ekf::controlOpticalFlowFusion()
 		const bool is_quality_good = (_flow_sample_delayed.quality >= _params.flow_qual_min);
 		const bool is_magnitude_good = !_flow_sample_delayed.flow_xy_rad.longerThan(_flow_sample_delayed.dt * _flow_max_rate);
 		const bool is_tilt_good = (_R_to_earth(2, 2) > _params.range_cos_max_tilt);
-		const bool is_body_rate_comp_available = calcOptFlowBodyRateComp();
 
-		// check if enough integration time and fail if integration time is less than 50%
-		// of min arrival interval because too much data is being lost
-		const float delta_time_min = 0.5e-6f * (float)_min_obs_interval_us;
-		const float delta_time_max = 0.2f;
+		const float delta_time_min = fmaxf(0.8f * _delta_time_of, 0.001f);
```
kamilritz, line 362: I feel that we should add unit test for these time checks. So that we exactly capture the behavior that we want.

bresch: Yes, I will add tests for that

priseborough (approved):

bresch:
Status: need flight testing

bresch:
I rebased on master and made a few test flights on a x500; with and without GPS aiding. Everything went well, I think it's safe to merge

jkflying (approved):

PR #929, opened 2020-11-01
EKF: add const reference getters for status flags

dagar:
This is for convenient usage outside of ECL.

### Before
``` C++
		filter_control_status_u control_status;
		_ekf.get_control_mode(&control_status.value);

		_preflt_checker.setUsingGpsAiding(control_status.flags.gps);
```

### After
``` C++
		_preflt_checker.setUsingGpsAiding(_ekf.control_status_flags().gps);
```

### DIFF
--- a/EKF/estimator_interface.h
+++ b/EKF/estimator_interface.h
@@ -311,9 +311,14 @@ public:
 
 	// get EKF mode status
 	void get_control_mode(uint32_t *val) { *val = _control_status.value; }
+	const decltype(filter_control_status_u::flags) &control_status_flags() const { return _control_status.flags; }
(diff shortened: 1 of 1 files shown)


jkflying (dismissed):
I'm a little hesitant to make it easier to expose potentially implementation-specific internals of ECL, but since we seem to be moving in that direction anyways...

dagar:
> I'm a little hesitant to make it easier to expose potentially implementation-specific internals of ECL, but since we seem to be moving in that direction anyways...

That's a fair point, but PX4 ekf2 already crossed that bridge years ago. I'd be happy to discuss walking some of that back.

PR #931, opened 2020-11-02
EKF allow initializing without baro or mag depending on configuration

dagar:
- ~~baro only required if EKF2_HGT_MODE barometric pressure~~ LATER
 - mag only required if EKF2_MAG_TYPE Automatic/Magnetic heading/3-axis
 - don't require 2 buffers worth of baro or mag data before initializing
 - relax accel norm requirements


EKF/ekf.cpp
```diff
@@ -235,8 +256,8 @@ bool Ekf::initialiseTilt()
 	const float accel_norm = _accel_lpf.getState().norm();
 	const float gyro_norm = _gyro_lpf.getState().norm();
 
-	if (accel_norm < 0.9f * CONSTANTS_ONE_G ||
-	    accel_norm > 1.1f * CONSTANTS_ONE_G ||
+	if (accel_norm < 0.8f * CONSTANTS_ONE_G ||
+	    accel_norm > 1.2f * CONSTANTS_ONE_G ||
```
bresch, line 260: Did you really have an issue with that?

dagar: Not personally, but a few times different users have reported problems/confusion that turned out to the the estimator not initializing due to how bad the lsm303d (original pixhawk, pixhawk 2, mindracer, etc) can be.

dagar: To be clear, the solution in those cases was re-calibration, the problem is without the estimator running there's nothing that really indicates what's wrong or what a user should do.

priseborough: Why don't we use the gravity vector to check the orientations for accel calibration? It only needs to know which side it is on which doesn't require running an ahrs.

EKF/ekf.cpp
```diff
@@ -160,42 +160,63 @@ bool Ekf::initialiseFilter()
...
 		}
 	}
 
-	const bool not_enough_baro_samples_accumulated = _baro_counter <= 2u * _obs_buffer_length;
-	const bool not_enough_mag_samples_accumulated = _mag_counter <= 2u * _obs_buffer_length;
-
-	if (not_enough_baro_samples_accumulated || not_enough_mag_samples_accumulated) {
-		return false;
+	if (_params.mag_fusion_type <= MAG_FUSE_TYPE_3D) {
+		if (_mag_counter < _obs_buffer_length) {
```
priseborough, line 189: @dagar can you confirm that the derivers for these sensors will not publish until valid data is available. In the past we have had the situation where some sensors could publish zero before the sensor had properly started. From memory the 2xbuffer was a hack to work around this.

dagar: That's the goal, and if there are still any offenders I'd like to find and fix them.

We're now very careful to not just fully reset a sensor, but also to give them their full reset time as per the datasheet. Then on transfer wherever possible sanity check the returned data before publishing (status bits, data ready, overflow, etc).

 - https://github.com/PX4/PX4-Autopilot/blob/149ac16bb442fc64eef320794a306b755758b5d4/src/drivers/magnetometer/isentek/ist8308/IST8308.cpp#L111
 - https://github.com/PX4/PX4-Autopilot/blob/149ac16bb442fc64eef320794a306b755758b5d4/src/drivers/magnetometer/isentek/ist8308/IST8308.cpp#L176

EKF/ekf.cpp
dagar: > Why don't we use the gravity vector to check the orientations for accel calibration?

We do.
 - https://github.com/PX4/PX4-Autopilot/blob/149ac16bb442fc64eef320794a306b755758b5d4/src/modules/commander/accelerometer_calibration.cpp#L257-L308


The question here is more about the filter silently failing (not initializing) vs starting and then the system likely complaining about high accel bias or one of the preflight innovation failures. Alternatively we could think about adding something simpler between sensors and preflight checks that catch this and tell the user to re-calibrate (maybe move vehicle_at_reset out of ecl). This isn't a widespread problem, but it does come up.

dagar:
Conflicts resolved and rebased on current master.

EKF/ekf.cpp
priseborough: Opening this up to a range of 1.2 to 0.8 would allow up to acos(1/1.2) = 33 deg of initial tilt error due to to a horizontal acceleration at the moment of alignment. I will assess via replay what this does to the EKF, but a warning every 10 seconds that this check fails that says 'keep vehicle or check accel calibration' could be a better option.

priseborough: I ran a test with adding a 33.5 deg pitch offset to the ekf initialiseTilt function and the EKF handled it OK and both attitude and gyro bias states had stabilised within 20 seconds.

Screen Shot 2020-12-16 at 5 42 00 pm

EKF/ekf.cpp
```diff
@@ -161,42 +161,63 @@ bool Ekf::initialiseFilter()
...
+			// not enough mag samples accumulated
+			return false;
+		}
 	}
 
-	// we use baro height initially and switch to GPS/range/EV finder later when it passes checks.
```
priseborough, line 198: What happens if those other sensors are not providing data at the time of alignment, eg delayed GPS lock? i on't think the EKF will handle this well. We cannot merge this until the implications of these sensor not being available at alignment has been tested. We should be delaying alignment until the nominated height sensor is returning data that passes quality checks.

dagar: That makes sense, I'll see if that's a viable change.

dagar:
Let's start with reducing the baro and mag samples and skipping the mag requirement depending on MAG_TYPE. We can come back to `vdist_sensor_type` later.

priseborough (approved):

PR #932, opened 2020-11-03
EKF: update gnd_effect flag automatically if terrain valid

dagar:
This internalizes what the PX4-Autopilot frontend is already doing to update the gnd_effect flag.

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -256,7 +256,7 @@ struct parameters {
 	float baro_innov_gate{5.0f};		///< barometric and GPS height innovation consistency gate size (STD)
 	float gps_pos_innov_gate{5.0f};		///< GPS horizontal position innovation consistency gate size (STD)
 	float gps_vel_innov_gate{5.0f};		///< GPS velocity innovation consistency gate size (STD)
-	float gnd_effect_deadzone{5.0f};	///< Size of deadzone applied to negative baro innovations when ground effect compensation is active (m)
+	float gnd_effect_deadzone{0.0f};	///< Size of deadzone applied to negative baro innovations when ground effect compensation is active (m)
 	float gnd_effect_max_hgt{0.5f};		///< Height above ground at which baro ground effect becomes insignificant (m)
 
 	// magnetometer fusion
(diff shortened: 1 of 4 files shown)


priseborough:
> This internalizes what the PX4-Autopilot frontend is already doing to update the gnd_effect flag.

That's a concern then. This functionality is only supposed to be activated when close to ground and about to land or about to takeoff, whereas with long range range finders the terrain could be valid for the entire flight. I'm not sure the EKF can have the contextual knowledge to make this decision correctly. I'll have a look at PX4-Autopilot and see how it's being set.

dagar:
It's not a major issue, but I didn't particularly like how we were making the decision (set_gnd_effect_flag) during the assembly of the `vehicle_local_position` publication. 

https://github.com/PX4/PX4-Autopilot/blob/ddc1f964d22b2a20bf02dda930b7ff8cf9a06c93/src/modules/ekf2/EKF2.cpp#L721-L729

It also interacts with the land detector until there's valid terrain data. https://github.com/PX4/PX4-Autopilot/blob/ddc1f964d22b2a20bf02dda930b7ff8cf9a06c93/src/modules/ekf2/EKF2.cpp#L393-L401

At a minimum I wanted to consolidate the scattered state and updates so it was a little easier to follow (and harder to get wrong).

dagar:
The land detector ground effect state also seems questionable. https://github.com/PX4/PX4-Autopilot/blob/ddc1f964d22b2a20bf02dda930b7ff8cf9a06c93/src/modules/land_detector/MulticopterLandDetector.cpp#L324

> I'm not sure the EKF can have the contextual knowledge to make this decision correctly.

Maybe it's time we finally talk about the architecture we need to get in place to handle all of these contextual decisions (eg vehicle model module). This also includes things like validity thresholds for different phases of flight and perhaps handling things a bit differently for VTOL (MC vs FW mode).

priseborough:
I've had a look at the multi copter land detector module and it needs to explicitly handle the 'in ground effect' decision using a height above ground threshold parameter to handle different vehicle characteristics. I will start work creating a new function to explicitly handle that decision.

PR #935, opened 2020-11-17
EKF: dynamically allocate ring buffers

dagar:
Saves a little bit of ram, flash, and can even be slightly faster by skipping unnecessary work.

Work in progress.

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -124,16 +124,17 @@ bool EstimatorInterface::checkIfVehicleAtRest(float dt, const imuSample &imu)
 
 void EstimatorInterface::setMagData(const magSample &mag_sample)
 {
-	if (!_initialised || _mag_buffer_fail) {
+	if (!_initialised) {
 		return;
 	}
 
 	// Allocate the required buffer size if not previously done
-	// Do not retry if allocation has failed previously
-	if (_mag_buffer.get_length() < _obs_buffer_length) {
-		_mag_buffer_fail = !_mag_buffer.allocate(_obs_buffer_length);
+	if (_mag_buffer == nullptr) {
+		_mag_buffer = new RingBuffer<magSample>(_obs_buffer_length);
 
-		if (_mag_buffer_fail) {
+		if (_mag_buffer == nullptr || !_mag_buffer->valid()) {
+			delete _mag_buffer;
+			_mag_buffer = nullptr;
 			printBufferAllocationFailed("mag");
 			return;
 		}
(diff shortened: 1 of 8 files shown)


EKF/estimator_interface.cpp
```diff
@@ -125,16 +138,17 @@ bool EstimatorInterface::checkIfVehicleAtRest(float dt, const imuSample &imu)
...
-	if (!_initialised || _mag_buffer_fail) {
+	if (!_initialised) {
 		return;
 	}
 
 	// Allocate the required buffer size if not previously done
-	// Do not retry if allocation has failed previously
-	if (_mag_buffer.get_length() < _obs_buffer_length) {
-		_mag_buffer_fail = !_mag_buffer.allocate(_obs_buffer_length);
+	if (_mag_buffer == nullptr && !_control_status.flags.in_air) {
```
bresch, line 146: So we do retry to allocate if is previously failed? Why didn't we do that before?

EKF/estimator_interface.cpp
```diff
@@ -512,34 +534,46 @@ void EstimatorInterface::setDragData(const imuSample &imu)
...
+	float max_time_delay_ms = math::max(_params.baro_delay_ms,
+					math::max(_params.auxvel_delay_ms,
+						_params.airspeed_delay_ms));
+
+	// mag mode
+	if (_params.mag_fusion_type != MAG_FUSE_TYPE_NONE) {
+		max_time_delay_ms = math::max(_params.mag_delay_ms, max_time_delay_ms);
+	}
+
+	// range aide or range height
```
bresch, line 546:
```suggestion
	// range aid or range height
```

EKF/imu_down_sampler.hpp
```diff
@@ -50,9 +50,9 @@ class ImuDownSampler
 	explicit ImuDownSampler(float target_dt_sec);
 	~ImuDownSampler() = default;
 
-	bool update(const imuSample &imu_sample_new);
+	bool update(const imuSample& imu_sample_new);
```
bresch, line 53:
```suggestion
	bool update(const imuSample &imu_sample_new);
```
I think we're usually using this convention, no?

EKF/estimator_interface.cpp
dagar: Spamming the error and potential performance issues if you get stuck trying (and failing) to allocate in flight. This allows it to re-try as needed up until flight. I'm also going to add a flag to rate limit the allocation failure print message.

Note this really isn't a common case we need to worry about  (especially past stm32f4).

EKF/imu_down_sampler.hpp
dagar: Yes, but apparently I have bad muscle memory. I usually let the tool handle it.

dagar:
A few remaining test failures.

Screenshot from 2020-12-07 13-20-12

dagar:
Resurrected in https://github.com/PX4/PX4-Autopilot/pull/19006.

PR #936, opened 2020-11-22
Auto update change indication

kamilritz:
The change indication is a nice tool to check if a proposed change does not introduce a functional change. It only covers limited scenarios, so it is by no means sufficient proof that there is no functional change introduced. The change indication requires to update certain csv files each time a functional change is made. This is a very delicate process, as the local build environment (compiler version, ...) has to match the CI build environment closely. Using docker containers is an alternative to produce correct change indication files. However, this involves many steps and is laborious. 

This PR adds another Github action, that updates the change indication on push, commits it, and pushes it to the remote branch.  The commit has the name `[AUTO COMMIT] update change indication`.  As it is triggered by push, when creating a pull request from a fork, it will run the check on the fork and update the fork branch with the new change indication. The original change indication in the `functional_change_indicator.yaml` will only be run on the pull request and does not update the change indication.

In case a functional change happens and the pull request is created almost immediately after pushing the branch, the pull request change_indicator check can fail, as it does not have the change indication update. In this case the check can be rerun after the automatic commit is pushed and will then pass.

### DIFF
--- a/.github/workflows/functional_change_indicator.yml
+++ b/.github/workflows/functional_change_indicator.yml
@@ -1,21 +1,19 @@
 name: Change Indicator
 
-on:
-  push:
-    branches:
-      - master
-  pull_request:
-    branches:
-      - '*'
+on: pull_request
 
 jobs:
   unit_tests:
     runs-on: ubuntu-latest
     container: px4io/px4-dev-base-focal:2020-09-14
     steps:
-    - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


bresch:
I'll try to have a look at it next week

kamilritz:
> I'll try to have a look at it next week

@bresch May be this week?

dagar (approved):
Looks good, let's give this a try.

PR #937, opened 2020-12-02
Update geo_lookup WMM  to latest Tue Dec  1 11:48:41 UTC 2020

dagar:


dagar:
I updated `iris_gps.csv` and squashed into the commit.

PR #938, opened 2020-12-03
EKF: RingBuffer minor improvements

dagar:
- allocate IMU and output buffers on construction according to defaults
 - determine buffer max time delay based on configuration parameters
 - reorder flowSample and extVisionSample to minimize padding
 - adjust parameter defaults to match PX4-Autopilot

This was broken out of https://github.com/PX4/PX4-ECL/pull/935.

Saves a bit of memory (depending on configuration) and about 2 kB of flash (surprisingly).

### DIFF
--- a/EKF/RingBuffer.h
+++ b/EKF/RingBuffer.h
@@ -45,14 +45,8 @@ template <typename data_type>
 class RingBuffer
 {
 public:
-	RingBuffer()
-	{
-		if (allocate(1)) {
-			// initialize with one empty sample
-			data_type d = {};
-			push(d);
-		}
-	}
+	explicit RingBuffer(size_t size) { allocate(size); }
+	RingBuffer() { allocate(1); }
 	~RingBuffer() { delete[] _buffer; }
 
 	// no copy, assignment, move, move assignment
@@ -63,12 +57,20 @@ public:
 
 	bool allocate(uint8_t size)
 	{
+		if (valid() && (size == _size)) {
+			// no change
+			return true;
+		}
+
+		if (size == 0) {
+			return false;
+		}
 
 		if (_buffer != nullptr) {
 			delete[] _buffer;
 		}
 
-		_buffer = new data_type[size];
+		_buffer = new data_type[size]{};
 
 		if (_buffer == nullptr) {
 			return false;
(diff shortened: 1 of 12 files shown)


EKF/RingBuffer.h
```diff
@@ -63,12 +57,20 @@ class RingBuffer
...
+		if (size == 0) {
+			return false;
+		}
 
 		if (_buffer != nullptr) {
 			delete[] _buffer;
 		}
 
-		_buffer = new data_type[size];
+		_buffer = new data_type[size]{};
```
jkflying, line 73: I guess it's just once, but we shouldn't be using the data anyways. For these kind of things I like to initialize with 0xDEADBEEF or something similar, at least for debug builds.

EKF/estimator_interface.cpp
```diff
@@ -512,34 +512,46 @@ void EstimatorInterface::setDragData(const imuSample &imu)
...
+		max_time_delay_ms = math::max(_params.flow_delay_ms, max_time_delay_ms);
+	}
+
+	if (_params.fusion_mode & (MASK_USE_EVPOS | MASK_USE_EVYAW | MASK_USE_EVVEL)) {
+		max_time_delay_ms = math::max(_params.ev_delay_ms, max_time_delay_ms);
+	}
 
 	// calculate the IMU buffer length required to accomodate the maximum delay with some allowance for jitter
-	_imu_buffer_length = (max_time_delay_ms / FILTER_UPDATE_PERIOD_MS) + 1;
+	_imu_buffer_length = ceilf(max_time_delay_ms / FILTER_UPDATE_PERIOD_MS) + 1;
```
jkflying, line 542: Does it need the +1 if you have `ceilf`?

EKF/RingBuffer.h
```diff
@@ -79,26 +81,15 @@ class RingBuffer
...
-		for (uint8_t index = 0; index < _size; index++) {
-			_buffer[index] = {};
-		}
-
 		_first_write = true;
 
 		return true;
 	}
 
-	void unallocate()
```
jkflying, line 93: This will probably make lots of noise for valgrind, even if it's ok in the real world.

jkflying (approved):
Looks ok, I'm a little hesitant about removing the cleanup though, more thinking forward to if it gets used in a different use case.

dagar:
> I'm a little hesitant about removing the cleanup though

You mean the `unallocate()`? Why aren't the destructors sufficient?

EKF/estimator_interface.cpp
dagar: Perhaps not, but the tolerances in the unit tests need to be adjusted slightly if it changes.

EKF/RingBuffer.h
dagar: What I discovered in https://github.com/PX4/PX4-ECL/pull/935 (the larger cleanup this was broken out of) is that there seems to be a few minor (hopefully) cases where it matters. I'd still like to fix that entirely and really ensure we're not using anything before there's valid data, but it needs some debug time.

PR #939, opened 2020-12-03
Fix GPS loss dead reckoning behaviour

priseborough:
During SITL testing with gazebo_plane it was observed that if sideslip fusion was disabled by setting EKF2_FUSE_BETA=0 and the GPS stopped by first setting the parameter SYS_FAILURE_EN = 1 followed by pxh> failure gps off, then the following occurred:

The vehicle_local_position.valid_xy flag periodically goes true true and wind states experience large variations after the EKF had reverted back to a non-position estimation mode - see below.

Screen Shot 2020-12-03 at 9 03 26 pm
 
These changes fix this by ensuring that a valid local position can never be reported true and the wind states and covariances are deactivated  if the EKF is in a non-position mode - see below
 
Screen Shot 2020-12-03 at 9 01 13 pm

Note that setting EKF2_FUSE_BETA = 1 is recommended for fixed wing flight in the majority of applications due to the extended dead-reckoning in excess of 5 minutes achievable after GPS loss.

@sfuhrer FYI

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1215,7 +1215,7 @@ void Ekf::controlBetaFusion()
 
 void Ekf::controlDragFusion()
 {
-	if (_params.fusion_mode & MASK_USE_DRAG &&
+	if ((_params.fusion_mode & MASK_USE_DRAG) &&
 	    !_using_synthetic_position &&
 	    _control_status.flags.in_air &&
 	    !_mag_inhibit_yaw_reset_req) {
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1161,8 +1161,8 @@ void Ekf::controlAirDataFusion()
 	// If both airspeed and sideslip fusion have timed out and we are not using a drag observation model then we no longer have valid wind estimates
 	const bool airspeed_timed_out = isTimedOut(_time_last_arsp_fuse, (uint64_t)10e6);
 	const bool sideslip_timed_out = isTimedOut(_time_last_beta_fuse, (uint64_t)10e6);
-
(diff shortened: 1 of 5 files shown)


EKF/estimator_interface.h
```diff
@@ -159,7 +159,7 @@ class EstimatorInterface
 	int getNumberOfActiveHorizontalAidingSources() const;
 
 	// return true if the local position estimate is valid
-	bool local_position_is_valid() const { return !_deadreckon_time_exceeded; }
+	bool local_position_is_valid() const { return local_position_is_valid(); }
```
dagar, line 162: ?

dagar, line 162: This could be deleted entirely if you also delete the call in print_status().

https://github.com/PX4/PX4-ECL/blob/67f13f7ededd4eb21e10abb6445f3f0fc4b7a6e5/EKF/estimator_interface.cpp#L605

priseborough: I've removed it.

priseborough: Done

EKF/control.cpp
```diff
@@ -1229,9 +1215,10 @@ void Ekf::controlBetaFusion()
 
 void Ekf::controlDragFusion()
 {
-	if (_params.fusion_mode & MASK_USE_DRAG) {
-		if (_control_status.flags.in_air
-				&& !_mag_inhibit_yaw_reset_req) {
+	if (_params.fusion_mode & MASK_USE_DRAG &&
```
dagar, line 1218:
```suggestion
	if ((_params.fusion_mode & MASK_USE_DRAG_ &&
```

dagar (approved):

sfuhrer (approved):
Looks good to me, and SITL yields good results.

PR #942, opened 2020-12-10
Fix GPS altitude variance computation

bresch:
Review commit-by-commit.

For some reason, the GPS position variance was fused with an accuracy of `1.5 x reported vertical position accuracy` (introduced here: https://github.com/PX4/PX4-ECL/commit/6b2e2dba909a44430dc3879e73e990231fcc2206#diff-4fbcbe3d9e3bd3fad671899aab33b57ee060d253a71a7c45e9acb78c2d297d22R125-R128).

To me, it seems to be from a copy-pasta of the GPS vertical speed fusion, where the vertical speed accuracy needs to be computed as `1.5 x reported horizontal speed accuracy`. (https://github.com/PX4/PX4-ECL/blob/fbf67bdef99f6033ce6c1fb6736cd7bf67020602/EKF/control.cpp#L698-L699).

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1103,11 +1103,7 @@ void Ekf::controlHeightFusion()
 			Vector3f gps_hgt_obs_var;
 			// vertical position innovation - gps measurement has opposite sign to earth z axis
 			_gps_pos_innov(2) = _state.pos(2) + _gps_sample_delayed.hgt - _gps_alt_ref - _hgt_sensor_offset;
-			// observation variance - receiver defined and parameter limited
-			// use scaled horizontal position accuracy assuming typical ratio of VDOP/HDOP
-			const float lower_limit = fmaxf(_params.gps_pos_noise, 0.01f);
-			const float upper_limit = fmaxf(_params.pos_noaid_noise, lower_limit);
-			gps_hgt_obs_var(2) = sq(1.5f * math::constrain(_gps_sample_delayed.vacc, lower_limit, upper_limit));
+			gps_hgt_obs_var(2) = getGpsAltVar();
 			// innovation gate size
 			gps_hgt_innov_gate(1) = fmaxf(_params.baro_innov_gate, 1.0f);
 			// fuse height information
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1103,7 +1103,7 @@ void Ekf::controlHeightFusion()
 			Vector3f gps_hgt_obs_var;
 			// vertical position innovation - gps measurement has opposite sign to earth z axis
 			_gps_pos_innov(2) = _state.pos(2) + _gps_sample_delayed.hgt - _gps_alt_ref - _hgt_sensor_offset;
-			gps_hgt_obs_var(2) = getGpsAltVar();
+			gps_hgt_obs_var(2) = getGpsHeightVariance();
 			// innovation gate size
 			gps_hgt_innov_gate(1) = fmaxf(_params.baro_innov_gate, 1.0f);
 			// fuse height information
--- a/EKF/covariance.cpp
+++ b/EKF/covariance.cpp
@@ -74,7 +74,7 @@ void Ekf::initialiseCovariance()
 		P(9,9) = sq(fmaxf(_params.range_noise, 0.01f));
 
 	} else if (_control_status.flags.gps_hgt) {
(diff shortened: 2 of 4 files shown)


EKF/ekf_helper.cpp
```diff
@@ -271,6 +271,9 @@ void Ekf::resetHeight()
 			_state.pos(2) = _hgt_sensor_offset - gps_newest.hgt + _gps_alt_ref;
 
 			// the state variance is the same as the observation
+			// TODO: the vacc is not constrained using the parameters
+			// do we really want to use the latest and not delayed measurement?
```
bresch, line 275: @priseborough Since the main EKF runs on delayed-time, why are we using the newest measurement here?

priseborough: If the reset to a sensor is guaranteed to be actioned on the same frame the sensor data is pulled from the buffer, the using the delayed data makes more sense. The exception to this is when the reset can be triggered at any time, in which case the data from the buffer can be old by up to the measurement interval, in which case using the latest data from the buffer input can result in smaller worst case timing errors. For example with GPS at 5Hz (200msec) and a buffer 100msec long, taking data from the input to the buffer results in a timing error between -100 and +100 msec, whereas if the output data was used, the timing error would be between 0 and 200msec.

EKF/control.cpp
```diff
@@ -1103,11 +1103,7 @@ void Ekf::controlHeightFusion()
 			Vector3f gps_hgt_obs_var;
 			// vertical position innovation - gps measurement has opposite sign to earth z axis
 			_gps_pos_innov(2) = _state.pos(2) + _gps_sample_delayed.hgt - _gps_alt_ref - _hgt_sensor_offset;
-			// observation variance - receiver defined and parameter limited
-			// use scaled horizontal position accuracy assuming typical ratio of VDOP/HDOP
-			const float lower_limit = fmaxf(_params.gps_pos_noise, 0.01f);
-			const float upper_limit = fmaxf(_params.pos_noaid_noise, lower_limit);
-			gps_hgt_obs_var(2) = sq(1.5f * math::constrain(_gps_sample_delayed.vacc, lower_limit, upper_limit));
+			gps_hgt_obs_var(2) = getGpsAltVar();
```
kamilritz, line 1106: at the moment we call it in a lot of places height. I would name it `getGpsHeightVariance`.

EKF/ekf_helper.cpp
```diff
@@ -1244,6 +1247,15 @@ void Ekf::updateBaroHgtOffset()
 	}
 }
 
+float Ekf::getGpsAltVar()
+{
+	// observation variance - receiver defined and parameter limited
+	const float lower_limit = fmaxf(_params.gps_pos_noise, 0.01f);
```
priseborough, line 1253: Due to difference in GPS vertical vs horizontal error characteristics, I think this should be

const float lower_limit = fmaxf(1.5f*_params.gps_pos_noise, 0.01f);

Will set up a call to discuss.

EKF/covariance.cpp
```diff
@@ -74,9 +74,7 @@ void Ekf::initialiseCovariance()
 		P(9,9) = sq(fmaxf(_params.range_noise, 0.01f));
 
 	} else if (_control_status.flags.gps_hgt) {
-		float lower_limit = fmaxf(_params.gps_pos_noise, 0.01f);
-		float upper_limit = fmaxf(_params.pos_noaid_noise, lower_limit);
-		P(9,9) = sq(1.5f * math::constrain(_gps_sample_delayed.vacc, lower_limit, upper_limit));
+		P(9,9) = getGpsAltVar();
```
kamilritz, line 77: nice helper function

EKF/ekf_helper.cpp
bresch: Yes, that makes sense

EKF/control.cpp
bresch: done in 30bf5f5

EKF/ekf_helper.cpp
bresch: Ok, thanks for the explanation, that makes sense now

kamilritz (commented):
From a formal perspektiv this looks good. Can't judge the 1.5 factor.

priseborough (approved):

PR #943, opened 2020-12-14
fakePosFusion: reset _last_known_posNE to current state when starting

bresch:
When starting fake position fusion, we need to set `_last_known_posNE` to the current estimate to get small innovations. This was the case before but got lost at some point.

fixes https://github.com/PX4/PX4-ECL/issues/934


kamilritz (approved):
Yes, this might have gone missing.

priseborough (approved):

PR #945, opened 2020-12-16
AlphaFilter add constructor that takes alpha and use in EKF

dagar:


AlphaFilter/AlphaFilter.hpp
```diff
@@ -85,7 +87,11 @@ class AlphaFilter {
 	 *
 	 * @return retrieve the filtered result
 	 */
-	void update(const T &sample) { _filter_state = updateCalculation(sample); }
+	const T& update(const T &sample)
```
MaEtUgR, line 90:
```suggestion
	const T &update(const T &sample)
```
or adapt the getter just below depending on the desired style, I have not hard feelings on this one.

MaEtUgR (dismissed):
Great! I should have done that from the beginning.

AlphaFilter/AlphaFilter.hpp
MaEtUgR: We're doing it the other way for parameters:
e.g. `void setVelocityGains(const matrix::Vector3f &P, const matrix::Vector3f &I, const matrix::Vector3f &D);`
Or is that just me going against the stream?
No also here `Subscription &operator=(const Subscription &other)` 🤷‍♂️

dagar:
Formatted with PX4/Firmware astyle.

AlphaFilter/AlphaFilter.hpp
dagar: I'm happy with whatever can be done automatically (astyle or clang-format).

PR #948, opened 2020-12-19
Removing Matlab EKF derivation

kamilritz:
With the newer python script, the autogeneration of the EKF code is easy to run. On the other hand the Matlab script is outdated in terms of the current Matrix::SparseVector syntax. Additionally, multiple people including me had issues when running the Matlab script https://github.com/PX4/PX4-ECL/issues/903.

I suggest therefore that we remove the EKF derivation Matlab scripts from the current codebase. The are still part of the history of the repo.


dagar (approved):
Looks good to me, but I'll let @priseborough do the honours.

priseborough (approved):

PR #950, opened 2020-12-21
flow: add unit test for yaw motion

bresch:
Because unit tests are fun!

### DIFF
--- a/test/sensor_simulator/ekf_wrapper.cpp
+++ b/test/sensor_simulator/ekf_wrapper.cpp
@@ -95,6 +95,11 @@ bool EkfWrapper::isIntendingFlowFusion() const
 	return _ekf->control_status_flags().opt_flow;
 }
 
+void EkfWrapper::setFlowOffset(const Vector3f &offset)
+{
+	_ekf_params->flow_pos_body = offset;
+}
+
 void EkfWrapper::enableExternalVisionPositionFusion()
 {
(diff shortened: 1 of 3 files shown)


kamilritz (dismissed):
It is fun, especially if you find the code not working as expected.
Looks good in general. Added some suggestions. How sure are you about the placement of a minus or plus sign in front of the computed body_rate?

bresch:
Thanks @kamilritz for the review. I addressed your comments.

> especially if you find the code not working as expected

Yes, I actually started that because the drone is never really locked-in when yawing in opt flow fusion only. So probably the issue is not on the EKF side, but due to the sensor.

> How sure are you about the placement of a minus or plus sign in front of the computed body_rate?

Really sure: https://github.com/PX4/PX4-ECL/blob/df7f261e038aa80b07a91857f2f952280938bfc6/EKF/common.h#L139
Rotation of the image = - rotation of the body
See also in https://github.com/PX4/PX4-ECL/blob/df7f261e038aa80b07a91857f2f952280938bfc6/EKF/optflow_fusion.cpp#L363

kamilritz:
@bresch  I do not see any update to the PR. If this is intentional, please go ahead and merge it.

bresch:
@kamilritz I forgot to push... :|

kamilritz (approved):
Thought so. Looks good to me.

PR #952, opened 2020-12-29
EKF: Treat combined sideslip and airspeed as a horizontal aiding source

priseborough:
The combination of airspeed and sideslip fusion acts to constrain horizontal velocity drift with wind relative dead reckoning, however this was not being recognised in the code.  This change treats the combination of airspeed and sideslip fusion as an equivalent horizontal velocity aiding source and ensures the solution status velocity_horiz flag will remain true if a fixed wing aircraft with airspeed and sideslip fusion loses GPS.

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -548,9 +548,9 @@ int EstimatorInterface::getNumberOfActiveHorizontalAidingSources() const
 	       + int(_control_status.flags.opt_flow)
 	       + int(_control_status.flags.ev_pos)
 	       + int(_control_status.flags.ev_vel)
-		   // Combined airspeed and sideslip fusion allows sustained wind relative dead reckoning
-		   // and so is treated as a single aiding source.
-		   + int(_control_status.flags.fuse_aspd && _control_status.flags.fuse_beta);
+	       // Combined airspeed and sideslip fusion allows sustained wind relative dead reckoning
+	       // and so is treated as a single aiding source.
+	       + int(_control_status.flags.fuse_aspd && _control_status.flags.fuse_beta);
 }
 
 bool EstimatorInterface::isHorizontalAidingActive() const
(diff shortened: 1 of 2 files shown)


EKF/estimator_interface.cpp
```diff
@@ -547,7 +547,10 @@ int EstimatorInterface::getNumberOfActiveHorizontalAidingSources() const
 	return int(_control_status.flags.gps)
 	       + int(_control_status.flags.opt_flow)
 	       + int(_control_status.flags.ev_pos)
-	       + int(_control_status.flags.ev_vel);
+	       + int(_control_status.flags.ev_vel)
+		   // Combined airspeed and sideslip fusion allows sustained wind relative dead reckoning
+		   // and so is treated as a single aiding source.
+		   + int(_control_status.flags.fuse_aspd && _control_status.flags.fuse_beta);
```
bresch, line 553:
```suggestion
	       + int(_control_status.flags.ev_vel)
		// Combined airspeed and sideslip fusion allows sustained wind relative dead reckoning
		// and so is treated as a single aiding source.
		+ int(_control_status.flags.fuse_aspd && _control_status.flags.fuse_beta);
```

bresch (approved):

priseborough:
This was tested in SITL using gazebo_plane with takeoff initiated using 'commander takeoff' and GPs failed once in loiter using 'failure gps on'. The EKF reported the  loss of GPS, eg:

WARN  [ecl/EKF] 589068000: GPS data stopped, using only EV, OF or air data

and continued to dead reckon with a steadily increasing reported position uncertainty for several minutes before the commander triggered a failsafe when the reported horizontal position uncertainty exceeded 45m = 2.5 x COM_POS_FS_EPH (default of 15).

WARN  [commander] Failsafe enabled: no global position

Screen Shot 2021-01-05 at 8 43 36 pm

PR #960, opened 2021-01-07
Fix mag fusion error flag

priseborough:
The defect this fixes was detected during testing of https://github.com/PX4/PX4-Autopilot/pull/16417.

During high angular rate manoeuvres, there are transient magnetometer fusion fault flags caused when the _accel_bias_inhibit is true and the delta velocity bias state variances and covariances are being set to 0 in the covariance prediction. This results in the checkAndFixCovarianceUpdate function returning false because P(i,i) is 0 and KHP(i,i) is a very small positive number (eg 1E-16). This has been missed in the past because of the slow logging rate for the fault status flags and the requirement for high angular rates to trigger the fault. Also it had no material effect on the filter kinematic state estimates.

Screen Shot 2021-01-07 at 8 36 43 pm

Screen Shot 2021-01-07 at 8 32 17 pm

This is fixed by ensuring the delta velocity bias state variances are lept at the last value before _accel_bias_inhibit was set to true instead of zero and by ensuring that the Kalman gains for the delta velocity bias states are forced to zero when the corresponding _accel_bias_inhibit is true.

Replay of the log shows that the fusion fault flag transients are eliminated:

Screen Shot 2021-01-07 at 8 35 13 pm


priseborough:
@dagar FYI

EKF/ekf.h
```diff
@@ -700,8 +699,16 @@ class Ekf final : public EstimatorInterface
...
+	bool measurementUpdate(Vector24f &K, const SparseVector24f<Idxs...> &H, float innovation)
 	{
+		if (_accel_bias_inhibit[0]) {
+			K(13) = 0.0f;
+		} else if (_accel_bias_inhibit[1]) {
+			K(14) = 0.0f;
+		} else if (_accel_bias_inhibit[2]) {
+			K(15) = 0.0f;
+		}
+
```
bresch, line 711: Two axes can be inhibited at the same time
```suggestion
		for (unsigned i = 0; i < 3; i++) {
			if (_accel_bias_inhibit[i]) {
				K(13 + i) = 0.0f;
			}
		}
```

bresch (changes requested):

EKF/ekf.h
dagar: Squashed into the original commit.

dagar (dismissed):

dagar:
Unit test failures.

Screenshot from 2021-01-07 12-27-04

Failing due to the expectation that the delta velocity variance is zero, but it's now being preserved (4.e-06 instead of 0).

https://github.com/PX4/PX4-ECL/blob/c212975abe83bf462bc512288ba30358f9a17865/test/test_EKF_initialization.cpp#L132

dagar:
EkfExternalVisionTest as well.

Screen Shot 2021-01-07 at 6 50 44 PM

EKF/ekf.h
priseborough: TXS

priseborough:
Test logs with Multi instance EKF enabled via EKF2_MULTI_IMU and EKF2_MULTI_MAG = 2 and SENS_IMU_SEL and SENS_MAG_SEL = 0.

Test 1: Two lane switch events were observed early inf flight, no change in flight characteristics noticed: https://logs.px4.io/plot_app?log=fa75a9f3-635d-4840-9a21-1307f58d34f6 

This was caused by transient fusion numerical health events being reported for the first instance. The filter_fault_flags bitmask value of 512 for this lane indicates that it was a numerical health condition for the North axis velocity fusion. This will be investigated further. 
Screen Shot 2021-01-08 at 2 54 42 pm

Test 2: Wouldn't arm due to a high accelerometer bias error  check failing: https://review.px4.io/plot_app?log=aee5be10-e613-49f5-b2b2-eddcea6ac521. This was due to the Z axis bias value being on the bottom limit for a sustained period for all instances. This was starting to recover when the vehicle was re-powered and will be investigated further
Screen Shot 2021-01-08 at 3 03 50 pm

Test 3: No anomalies detected: https://review.px4.io/plot_app?log=6c8ddf8e-1915-450a-84c6-c17757d33fdd

priseborough:
I found a bug in the code that detects against badly conditioned delta velocity bias state estimates by looking for a saturated bias in the vertical direction that is opposite in sign to the vertical velocity and position innovations. This was introduced when we added different innovation variables for each vertical position source. The code does not handle the case where vertical position is from a baro.

https://github.com/PX4/PX4-ECL/blob/c212975abe83bf462bc512288ba30358f9a17865/EKF/covariance.cpp#L948-L952

I'll add a patch for that condition.

priseborough:
I'll try and reproduce some non-zero filter_fault_flags bitmask events with a single EKF instance running. Replay doesn't work on logs generated with multiple EKF instances.

EKF/covariance.cpp
```diff
@@ -949,7 +949,8 @@ void Ekf::fixCovarianceErrors(bool force_symmetry)
 		bool bad_acc_bias = (fabsf(down_dvel_bias) > dVel_bias_lim
 				     && ( (down_dvel_bias * _gps_vel_innov(2) < 0.0f && _control_status.flags.gps)
 				     ||   (down_dvel_bias * _ev_vel_innov(2) < 0.0f && _control_status.flags.ev_vel) )
-				     && down_dvel_bias * _gps_pos_innov(2) < 0.0f);
+				     && ( (down_dvel_bias * _gps_pos_innov(2) < 0.0f  && _control_status.flags.gps_hgt)
+				     ||   (down_dvel_bias * _baro_hgt_innov(2) < 0.0f  && _control_status.flags.baro_hgt) ) );
```
dagar, line 953: What about `ev_hgt`?

priseborough: I'll add support for EV and range finder height.

priseborough:
I did 6 flights/power cycles with this today running a single ekf instance and was unable to replicate any of the anomalies observed during yesterdays testing with multi-ekf's enabled.

dagar (approved):

PR #961, opened 2021-01-08
remove ECL_{INFO,WARN,ERN}_TIMESTAMPED macros

dagar:
@kamilritz can we revisit what you wanted here? We already have mechanisms to print these messages timestamped (debug builds or similar). The mix of timestamped and non-timestamped messages is a bit weird. Also note that in log files the messages are timestamped.

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -339,8 +339,7 @@ void Ekf::controlExternalVisionFusion()
 
 		// Turn off EV fusion mode if no data has been received
 		stopEvFusion();
-		ECL_INFO_TIMESTAMPED("vision data stopped");
-
+		ECL_INFO("vision data stopped");
 	}
 }
 
@@ -555,7 +554,7 @@ void Ekf::controlGpsFusion()
 			if (_control_status.flags.ev_pos && !(_params.fusion_mode & MASK_ROTATE_EV)) {
 				resetHorizontalPosition();
 			}
-			ECL_WARN_TIMESTAMPED("GPS quality poor - stopping use");
+			ECL_WARN("GPS quality poor - stopping use");
 		}
 
 		// handle case where we are not currently using GPS, but need to align yaw angle using EKF-GSF before
@@ -656,7 +655,7 @@ void Ekf::controlGpsFusion()
 				resetVelocity();
 				resetHorizontalPosition();
 				_velpos_reset_request = false;
-				ECL_WARN_TIMESTAMPED("GPS fusion timeout - reset to GPS");
+				ECL_WARN("GPS fusion timeout - reset to GPS");
 
 				// Reset the timeout counters
 				_time_last_hor_pos_fuse = _time_last_imu;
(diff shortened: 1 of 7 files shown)


kamilritz:
Will have a look

kamilritz:
@dagar The purpose of these print statements was to have timestamped messages in the console during EKF replay. This helped to get a good understanding of what is going on during a log replay (e.g. position reset after x seconds) without having to look at multiple plots. If I remember correctly, not all of those messages were displayed in plot juggler pop window.

dagar:
I'd like to get the output under control because it's a bit absurd with multi-ekf. https://github.com/PX4/PX4-ECL/issues/909

As a start can we limit many of these messages to replay builds? If we had that in place you'd actually be free to add significantly more console logging, with no impact for embedded usage.

kamilritz:
@dagar I understand. But then I would start by removing some of the print statements completely, and make sure that there is a good indicator in the log file for this as you suggested in #909.

priseborough:
I would like to only have print statements when enabled by parameter for development work provided we capture these events with status flags. We could add an EKF2_OPTIONS bitmask parameter to control developer options like this. Thoughts??

priseborough:
Could we make them a build time option?

dagar:
> Could we make them a build time option? Is it as simple as [priseborough@5559274](https://github.com/priseborough/PX4-ECL/commit/5559274ce0a91fc3bdca6a6c4b9d65e56fd4a7ee)

I'm fine with either build time or runtime options, whatever's convenient for your workflow.

Can we do a quick pass to review the existing messages within ECL and make sure any noteworthy events are going to be captured in flags or counters? 
 
For example, at a glance on flight review things like a midflight position or velocity reset stand out immediately. 

https://github.com/PX4/PX4-ECL/blob/602e65146c975c12482560fafab7e05bf43b0029/EKF/ekf_helper.cpp#L75

I was thinking we could add simple counters for each unique event like this that's worth tracking.

dagar:
> Could we make them a build time option? Is it as simple as [priseborough@5559274](https://github.com/priseborough/PX4-ECL/commit/5559274ce0a91fc3bdca6a6c4b9d65e56fd4a7ee)

My only minor nitpick with that in particular is the desire to by default keep things as consistent as possible across NuttX and Linux (both SITL and real flights).

priseborough:
> For example, at a glance on flight review things like a midflight position or velocity reset stand out immediately.
> 
> https://github.com/PX4/PX4-ECL/blob/602e65146c975c12482560fafab7e05bf43b0029/EKF/ekf_helper.cpp#L75
> 
> I was thinking we could add simple counters for each unique event like this that's worth tracking.

Most of these are already captured with the state_reset_status counters, eg _state_reset_status.velNE_counter is incremented for the example mentioned. One drawback is that these counters don't identify the source of the reset, however that information is contained in the _control_status.flags data. The other drawback is that these reset counters are logged as part of the vehicle _local_position message, not the estimator status message, but the information is logged.

I've been through the ECL_INFO prints, excluding the one in EstimatorInterface::print_status(), and they have corresponding changes  to _state_reset_status and _control_status.flags.

The ECL_WARN print statements are more problematic. For example ECL_WARN("GPS quality poor - stopping use"); would be indicated in the logs by _control_status.flags.gps going false but we wouldn't know what caused it.

A bitmask capturing these ECL_WARN (and possibly ECL_INFO) events would be easy to add, but these are single frame events so logging rate is an issue unless we create a separate message that is published only when the mask changes or a separate counter is used for each event.

dagar:
> The other drawback is that these reset counters are logged as part of the vehicle _local_position message, not the estimator status message, but the information is logged.

Additionally adding this information to `estimator_status` could be an easy next step, the tiny amount of duplication is fine. We now publish `estimator_status_flags` at a minimal rate or immediately on change, so maybe we can get in the habit (and add tooling) of looking at the reset count change + change in control status flags. 

> A bitmask capturing these ECL_WARN (and possibly ECL_INFO) events would be easy to add, but these are single frame events so logging rate is an issue unless we create a separate message that is published only when the mask changes or a separate counter is used for each event.

A small counter (uint8_t) for each event combined with a message we can log 100% (like estimator_status_flags) is what I was thinking.

Once this settles we can then go back to the ecl analysis script and make sure these are clear.

priseborough:
I'm happy for this change to go in as is and then work on the messaging issues raised.

priseborough (approved):

dagar:
Finished in https://github.com/PX4/PX4-ECL/pull/987.

PR #962, opened 2021-01-13
EKF set global origin

dagar:
- new `bool setGlobalOrigin(const double latitude, const double longitude, const float altitude);` intended to be used Firmware side with things like Mavlink SET_GPS_GLOBAL_ORIGIN
 - replace `bool get_ekf_origin(uint64_t *origin_time, map_projection_reference_s *origin_pos, float *origin_alt) const;` with const reference inlined getters
 - new vertical position reset helper `void resetVerticalPositionTo(const float &new_vert_pos);`

FYI @mcsauder

### DIFF
--- a/EKF/ekf.h
+++ b/EKF/ekf.h
@@ -151,7 +151,12 @@ public:
 
 	// get the ekf WGS-84 origin position and height and the system time it was last set
 	// return true if the origin is valid
-	bool get_ekf_origin(uint64_t *origin_time, map_projection_reference_s *origin_pos, float *origin_alt) const;
+	bool getEkfGlobalOrigin(uint64_t &origin_time, double &latitude, double &longitude, float &origin_alt) const;
+	bool setEkfGlobalOrigin(const double latitude, const double longitude, const float altitude);
+
+	float getEkfGlobalOriginAltitude() const { return _gps_alt_ref; }
+	bool setEkfGlobalOriginAltitude(const float altitude);
+
 
 	// get the 1-sigma horizontal and vertical position uncertainty of the ekf WGS-84 position
 	void get_ekf_gpos_accuracy(float *ekf_eph, float *ekf_epv) const;
@@ -617,6 +622,8 @@ private:
 	inline void resetHorizontalPositionToVision();
 
 	inline void resetHorizontalPositionTo(const Vector2f &new_horz_pos);
+
+	inline void resetVerticalPositionTo(const float &new_vert_pos);
 
 	void resetHeight();
 
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -688,7 +688,7 @@ bool Ekf::setEkfGlobalOrigin(const double latitude, const double longitude, cons
 	bool current_pos_available = false;
 	double current_lat = static_cast<double>(NAN);
 	double current_lon = static_cast<double>(NAN);
-	float current_alt  = NAN;
+	float current_alt  = 0.f;
 
 	// if we are already doing aiding, correct for the change in position since the EKF started navigating
 	if (map_projection_initialized(&_pos_ref) && isHorizontalAidingActive()) {
@@ -699,7 +699,6 @@ bool Ekf::setEkfGlobalOrigin(const double latitude, const double longitude, cons
 
 	// reinitialize map projection to latitude, longitude, altitude, and reset position
 	if (map_projection_init_timestamped(&_pos_ref, latitude, longitude, _time_last_imu) == 0) {
-
 		if (current_pos_available) {
 			// reset horizontal position
 			Vector2f position;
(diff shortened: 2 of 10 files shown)


mcsauder:
This PR requires PX4 PR [#16544](https://github.com/PX4/PX4-Autopilot/pull/16544), but together everything appears to be in working order:
```
ekf2 status
ekf2 instance:		0
attitude valid:		true
local position valid:	true
global position (LLA):	40.5841001, -111.8316491, 1473.5
ekf2: ECL update: 2069 events, 14463us elapsed, 6.99us avg, min 1us max 75us 5.133us rms
ekf2: ECL full update: 1973 events, 550771us elapsed, 279.15us avg, min 155us max 1264us 109.347us rms
nsh> 
nsh> commander origin 45.0000001 111.0000000001 150.125
nsh> ecl: new vert pos: -1323.379272
nsh> Get NED origin (LLA): 45.0000001000, 111.0000000001, 150.125
```

Thanks for all of your work @dagar !

EKF/ekf_helper.cpp
```diff
@@ -223,28 +223,44 @@ void Ekf::resetHorizontalPositionTo(const Vector2f &new_horz_pos)
 	_state_reset_status.posNE_counter++;
 }
 
+void Ekf::resetVerticalPositionTo(const float &new_vert_pos)
+{
+	printf("ecl: new vert pos: %f", (double)new_vert_pos);
```
dagar, line 228: Change to ECL_INFO or remove.

mcsauder: Thanks for spotting that printf() degbugging, it' wasn't meant to stay in. :}

mcsauder: Fixed and pushed up.

bresch (changes requested):
Good, I didn't spot anything wrong, however:
The test is enough to show that the origin has moved, however, it does not show if the rest of the states are adjusted properly. This could be done to check some innovations just after the origin update.

bresch:
I added a check for the GPS innov test ratio after reset and I get
```
Value of: matrix::isEqual(Vector2f(hpos), zero, 0.05f)
  Actual: false
Expected: true
pos = 3345244225536, 0
```
So something must be wrong

dagar:
Higher level the cases that come to mind for testing are...

1. GPS checks pass (EKF sets origin) then origin is manually set
2. origin manually set before GPS checks pass

mcsauder:
> I added a check for the GPS innov test ratio after reset and I get
> 
> ```
> Value of: matrix::isEqual(Vector2f(hpos), zero, 0.05f)
>   Actual: false
> Expected: true
> pos = 3345244225536, 0
> ```
> 
> So something must be wrong

The call to `_sensor_simulator.runSeconds(1);` was returning immediately, not blocking, so the values were not updated in the ekf before the test comparisons were being run.  I also noticed that the definition for `getGpsVelPosInnovRatio()` was `getGpsVelPosInnovRatio(float, float, float, float)` (no vectors like the other similarly named functions contain), so I updated the test according.  Tests pass now.

Thanks a ton for your help @bresch !

EKF/ekf.h
```diff
@@ -261,6 +262,8 @@ class Ekf final : public EstimatorInterface
 		*counter = _state_reset_status.quat_counter;
 	}
 
+	const float &global_origin_altitude() const { return _gps_alt_ref; }
```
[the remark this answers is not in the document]
mcsauder: Good question. I'm not sure, but it's fixed up now.

mcsauder:
> Higher level the cases that come to mind for testing are...
> 
>     1. GPS checks pass (EKF sets origin) then origin is manually set
> 
>     2. origin manually set before GPS checks pass

Thanks @dagar .  I've pushed up a new commit to address those cases.

mcsauder (dismissed):
Unit tests refactored and passing.  Functional bench tests passing on a pixhawk 4 before and after GPS lock.

@bresch , let us know if there is anything else you'd like to see with this PR.  Thanks!

mcsauder:
Rebased against current ecl master branch.

dagar:
Looks good from my perspective, but I'd like @bresch to verify.

test/sensor_simulator/gps.h
```diff
@@ -51,27 +51,26 @@ class Gps: public Sensor
 	~Gps();
 
 	void setData(const gps_message& gps);
-	void stepHeightByMeters(float hgt_change);
-	void stepHorizontalPositionByMeters(Vector2f hpos_change);
-	void setPositionRateNED(const Vector3f& rate) { _gps_pos_rate = rate; }
-	void setAltitude(int32_t alt);
-	void setLatitude(int32_t lat);
-	void setLongitude(int32_t lon);
+	void stepHeightByMeters(const float hgt_change);
```
bresch, line 54: I don't think using the `const` keyword in front of a native type in the function declaration is a good thing. It's anyway passing a copy, so the user doesn't care. You can however leave it in the definition to make sure you're not changing it in the function.

bresch (approved):

PR #969, opened 2021-01-31
Heading reset to magnetometer from GPS or EV

isidroas:
**Issue found**
I have been flying an UAV with an intermittent external vision and I found that heading state did not reset to magnetometer when external vision failed. 

**Solution**
In _mag_control.cpp_ I added a logic which is aware of other sources of heading may stop contributing measures. Also, I extended the tests to catch that bug.

### DIFF
--- a/EKF/mag_control.cpp
+++ b/EKF/mag_control.cpp
@@ -331,12 +331,10 @@ void Ekf::run3DMagAndDeclFusions()
 
 bool Ekf::otherHeadingSourcesHaveFinished()
 {
-    bool no_other_yaw_aiding_than_mag = noOtherYawAidingThanMag();
-
     // detect rising edge
-    bool result = no_other_yaw_aiding_than_mag && !_no_other_yaw_aiding_than_mag;
+    bool result = noOtherYawAidingThanMag() && !_no_non_mag_heading_aiding_running;
 
-    _no_other_yaw_aiding_than_mag = no_other_yaw_aiding_than_mag;
+    _no_non_mag_heading_aiding_running = noOtherYawAidingThanMag();
 
     return  result;
 }
--- a/EKF/mag_control.cpp
+++ b/EKF/mag_control.cpp
@@ -72,7 +72,7 @@ void Ekf::controlMagFusion()
 		return;
 	}
 
-	_mag_yaw_reset_req |= otherHeadingSourcesHaveFinished();
+	_mag_yaw_reset_req |= otherHeadingSourcesHaveStopped();
 
 	if (noOtherYawAidingThanMag() && _mag_data_ready) {
 		if (_control_status.flags.in_air) {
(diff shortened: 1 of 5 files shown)


EKF/mag_control.cpp
```diff
@@ -72,6 +72,8 @@ void Ekf::controlMagFusion()
 		return;
 	}
 
+	_mag_yaw_reset_req |= otherHeadingSourcesHaveFinished();
```
kamilritz, line 75: This returns true only at the one time point where the other sensor is lost. If in the following lines, the reset is not happening due to MagFieldDisturbance or another reason. It will still try to reset, as the `_mag_yaw_reset_req` stays true. So that is okay.

test/test_EKF_fusionLogic.cpp
```diff
@@ -281,19 +282,23 @@ TEST_F(EkfFusionLogicTest, doVisionHeadingFusion)
...
-	_sensor_simulator.runSeconds(6);
+	_sensor_simulator.runSeconds(7.1);
 
 	// THEN: EKF should stop to intend to fuse vision position estimate
 	//       and EKF should not have a valid local position estimate anymore
 	EXPECT_FALSE(_ekf_wrapper.isIntendingExternalVisionPositionFusion());
 	EXPECT_FALSE(_ekf_wrapper.isIntendingExternalVisionVelocityFusion());
-	EXPECT_TRUE(_ekf_wrapper.isIntendingExternalVisionHeadingFusion());
-	// TODO: it is still intending to fuse ev_yaw. There should be some fallback to mag if possible
+	EXPECT_FALSE(_ekf_wrapper.isIntendingExternalVisionHeadingFusion());
```
kamilritz, line 296: Thanks for eliminating this TODO.

EKF/mag_control.cpp
```diff
@@ -327,3 +329,14 @@ void Ekf::run3DMagAndDeclFusions()
 	}
 }
 
+bool Ekf::otherHeadingSourcesHaveFinished()
+{
+    bool no_other_yaw_aiding_than_mag = noOtherYawAidingThanMag();
+
+    // detect rising edge
+    bool result = no_other_yaw_aiding_than_mag && !_no_other_yaw_aiding_than_mag;
```
kamilritz, line 337: I am not so happy with the naming here. 
1. The local and global variable are easy to confound. Maybe just call `noOtherYawAidingThanMag()` in this line directly.
2. `no_other_yaw_aiding_than_mag` is a bit a misleading name, as it suggests that mag is already running. Could we change it to something like `no_non_mag_heading_aiding_running`? The same would apply for the function name.

kamilritz (approved):

isidroas:
Hi @kamilritz, @bresch. Do you want to change something else of this branch?
I believe that without these modifications, the magnetometer innovations would be too high so its fusion would be blocked (fortunately, I discovered it this without flying).

kamilritz:
@bresch for me this looks good

EKF/mag_control.cpp
```diff
@@ -327,3 +329,12 @@ void Ekf::run3DMagAndDeclFusions()
 	}
 }
 
+bool Ekf::otherHeadingSourcesHaveFinished()
+{
+    // detect rising edge
+    bool result = noOtherYawAidingThanMag() && !_no_non_mag_heading_aiding_running;
+
+    _no_non_mag_heading_aiding_running = noOtherYawAidingThanMag();
```
bresch, line 337:
```suggestion
    bool result = noOtherYawAidingThanMag() && _non_mag_yaw_aiding_running_prev;

    _non_mag_yaw_aiding_running_prev = !noOtherYawAidingThanMag();
```

EKF/mag_control.cpp
```diff
@@ -327,3 +329,12 @@ void Ekf::run3DMagAndDeclFusions()
 	}
 }
 
+bool Ekf::otherHeadingSourcesHaveFinished()
```
bresch, line 332:
```suggestion
bool Ekf::otherHeadingSourcesHaveStopped()
```

EKF/mag_control.cpp
bresch: using the same name as the function and not using double negation makes it simpler to read IMO. (also, "heading_aiding" is a bit strange to pronounce :D)

bresch (changes requested):

bresch (approved):

PR #971, opened 2021-02-05
Height test ratio status

NicolasM0:
Reported height test ratio was always equal to the position test ratio.

I fixed that using the correct index in _gps_pos_test_ratio vector.

I also add all height sources in this status.

Test: On the ground with baro as main height source, when I blow on my autopilot, the height test ratio increase.

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -910,7 +910,19 @@ void Ekf::get_innovation_test_status(uint16_t &status, float &mag, float &vel, f
 	pos = math::max(sqrtf(_gps_pos_test_ratio(0)), sqrtf(_ev_pos_test_ratio(0)));
 
 	// return the vertical position innovation test ratio
-	hgt = sqrtf(_gps_pos_test_ratio(1));
+	if (_control_status.flags.baro_hgt) {
+		hgt = sqrtf(_baro_hgt_test_ratio(1));
+
+	} else if (_control_status.flags.gps_hgt) {
+		hgt = sqrtf(_gps_pos_test_ratio(1));
+
+	} else if (_control_status.flags.rng_hgt) {
+		hgt = sqrtf(_rng_hgt_test_ratio(1));
+
+	} else if (_control_status.flags.ev_hgt) {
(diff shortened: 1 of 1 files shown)


EKF/control.cpp
```diff
@@ -1071,6 +1071,13 @@ void Ekf::controlHeightFusion()
 	}
 
 	if (fuse_height) {
+
+		// Reset all height tests ratio
+		_gps_pos_test_ratio(1) = 0;
+		_ev_pos_test_ratio(1) = 0;
+		_baro_hgt_test_ratio(1) = 0;
+		_rng_hgt_test_ratio(1) = 0;
```
bresch, line 1079:
```suggestion
		resetHeightTestRatios();
```
Could you please put them in a small function like that?

bresch (changes requested):

EKF/ekf_helper.cpp
```diff
@@ -910,7 +910,11 @@ void Ekf::get_innovation_test_status(uint16_t &status, float &mag, float &vel, f
 	pos = math::max(sqrtf(_gps_pos_test_ratio(0)), sqrtf(_ev_pos_test_ratio(0)));
 
 	// return the vertical position innovation test ratio
-	hgt = sqrtf(_gps_pos_test_ratio(0));
+	hgt = sqrtf(math::max(math::max(_gps_pos_test_ratio(1),
```
dagar, line 913: Shouldn't we be returning the corresponding test ratio for the current height mode?

NicolasM0: To be sure this return the current height ratio, I added a reset of all height ratios before calculating the current one .

I can use `ratio._control_status.flags.*_hgt` here to find the correct height ratio if you think that is better but I thought that is easier to read like that and more robust to future changes like that.

dagar: I have a minor preference for being lazier (not blanket resetting everything every iteration regardless of use), but I think it's fine either way here.

NicolasM0: here is a new implementation without useless resets

bresch (approved):

dagar (approved):

PR #974, opened 2021-02-08
flow terrain: use dedicated variable for last fusion time

bresch:
In many places, the main EKF assumes that a recent `_time_last_of_fuse` means that the horizontal velocity is constrained by flow fusion. This is not the case as this variable was also used in the terrain estimator.
Also, even if in most places the flow fusion flag is checked in addition to `_time_last_of_fuse`, there is a time window where this condition can be true since the flag can change instantaneously.

This PR adds a new `_time_last_flow_terrain_fuse` to avoid confusion and bugs related to this multiple use of the same variable.


dagar (approved):

dagar:
Thanks @bresch

kamilritz (approved):

PR #975, opened 2021-02-10
yaw emergency: check angle between vel observation and estimate

bresch:
In https://github.com/PX4/PX4-ECL/pull/914 we tried to robustify the yaw failure detection against false positives due to a divergent velocity observation after takeoff.

Unfortunately, the condition is now too strict and is difficult to pass even with a large yaw error (in jMAVSim with an initial yaw error of 90 or even 180 degrees, the check does not pass!). For example, with a yaw error of 180 degrees, a sitl test shows that going forward generates an innovation not more than double the observation (as expected). Even worse, an error of 90 degrees produces an innovation equal to the observation.

To address this issue, the current PR computes the angle between the observation and the current XY velocity estimate to determine if there is a yaw error.

SITL tests (jMAVSim) using this PR:
- 180 degrees initial yaw error:
https://logs.px4.io/plot_app?log=e524b573-538b-43f8-be4c-4d67e41d1ae3

- 90 degrees initial yaw error:
https://logs.px4.io/plot_app?log=a435ddbe-8a9a-4fd5-b82a-ec1f665e3f7b

To reproduce, simply swap the axes or hardcode some values here: https://github.com/PX4/jMAVSim/blob/358b6cca4093646eb96e0cb075e45efa8f4a0c48/src/me/drton/jmavsim/SimpleSensors.java#L140


EKF/control.cpp
```diff
@@ -617,12 +617,18 @@ void Ekf::controlGpsFusion()
...
-				const float hvel_obs_sq = sq(_last_vel_obs(0)) + sq(_last_vel_obs(1));
-				const float hvel_innov_sq = sq(_last_fail_hvel_innov(0)) + sq(_last_fail_hvel_innov(1));
-				if (hvel_innov_sq > 9.0f * hvel_obs_sq) {
-					is_yaw_failure = true;
+				const float norm_sq_vel_obs_xy = _last_vel_obs.xy().norm_squared();
+				const float norm_sq_vel_xy = _state.vel.xy().norm_squared();
+				const float obs_dot_vel = Vector2f(_last_vel_obs).dot(_state.vel.xy());
+				float cos_sq = 1.f;
+
+				if (norm_sq_vel_obs_xy > 4.f && norm_sq_vel_xy > 4.f) {
```
priseborough, line 625: Instead of using a fixed threshold of sq(2 ) could we use a TBD sigma threshold based on the reported GPS accuracy to better tolerate GPs glitches and variations in accuracy between receivers.

sq(TBD * fmaxf(_gps_sample_delayed.hacc, fmaxf(_params.gps_pos_noise, 0.01f))

bresch: Good idea! I added it here in f5a417a

priseborough (approved):

PR #978, opened 2021-02-16
Yaw on ground fusion fix

bresch:
While a yaw test ration > 1 rejects the mag data in air, the fusion is forced with a limited innovation on ground to fix gyro bias and local magnetic effects problems pre-takeoff. However, the same logic is applied after landing and immediately after the land detector detects landing. In some cases, this can disturb the other EKF states which can lead to a state change of the land detector. 

To avoid such a loop, the start of this forced yaw fusion is delayed by 5 seconds after landing.

fixes https://github.com/PX4/PX4-Autopilot/issues/16801


bresch:
The change indicator triggered because dcaa89d slightly changes the timing (it's only updated at the land detector rate). I would need to double check if this is the only difference.

EKF/estimator_interface.h
```diff
@@ -320,6 +329,8 @@ class EstimatorInterface
 					// [1] Vertical position drift rate (m/s)
 					// [2] Filtered horizontal velocity (m/s)
 	uint64_t _time_last_move_detect_us{0};	// timestamp of last movement detection event in microseconds
+	uint64_t _time_last_on_ground_us{0};	///< last time we were on the ground (uSec)
```
julianoes, line 332: Set but unused on purpose?

bresch: it's used somewhere, I just moved it.

EKF/mag_fusion.cpp
```diff
@@ -649,8 +649,10 @@ void Ekf::updateQuaternion(const float innovation, const float variance, const f
 		if (_control_status.flags.in_air) {
 			return;
 
-		} else {
+		} else if (isTimedOut(_time_last_in_air, (uint64_t)5e6)) {
 			// constrain the innovation to the maximum set by the gate
+			// we need to delay this forced fusion to avoid starting it
+			// immediately after touchdown, when the drone is still armed
```
priseborough, line 655: We need to return if not in air and not timed out

priseborough (approved):

EKF/mag_fusion.cpp
priseborough: Fixed with https://github.com/PX4/PX4-ECL/commit/d7459602c84b7faf615d9dd4ed5387a33aa34838

priseborough (dismissed):

priseborough (approved):

PR #979, opened 2021-02-17
Stop large yaw errors interfering with detection of landing

priseborough:
Fixes the issue observed in this log https://review.px4.io/plot_app?log=3c5f784f-7bd8-4a51-b5ff-cf7edf1ac709 

A large yaw error added in flight causes the yaw sensor  to be rejected and the EKF tolerates the loss of yaw due to the regular direction changes which keep the yaw observable. Immediately on landing when the in_air status goes false, the EKF starts fusing the yaw measurement and the large innovation which causes a degradation in states estimates which prevents the motor stop logic in the autopilot from activating due to the is landed check being affected by the state estimates.

Screen Shot 2021-02-17 at 8 06 33 pm

Screen Shot 2021-02-17 at 8 08 31 pm

This fix adds a 5 second delay on landing before yaw fusion will be forced, and if yaw fusion is failing prior, will perform a yaw reset to the sensor reading.

TODO - a way of reliably replicating this fault in SITL.


EKF/mag_fusion.cpp
```diff
@@ -643,19 +643,35 @@ void Ekf::updateQuaternion(const float innovation, const float variance, const f
...
-		if (_control_status.flags.in_air) {
-			return;
-
+		if (!_control_status.flags.in_air) {
+			if (isTimedOut(_time_last_in_air, 5E6)) {
+				if (_reset_yaw_after_landing) {
+					// If landed and yaw measurements were being rejected then we
+					// need to reset the yaw before recommencing fusion to prevent
+					// large innovations corrupting the state estimates
+					resetMagHeading(_mag_lpf.getState());
```
bresch, line 658: We shouldn't call a `resetMagHeading` here but rather set the `_mag_yaw_reset_req` flag that will be processed in `mag_control` when possible.

I'll add that to my PR which also includes a small cleanup of the `_time_last_in_air` and `_time_last_on_ground` variables : https://github.com/PX4/PX4-ECL/pull/978

bresch (changes requested):

PR #981, opened 2021-02-18
allow range aid without horizontal aiding

dagar:
- skip range aid velocity check when horizontal aiding isn't active
 - added new tilt limit check when horizontal aiding isn't active (currently 15 degrees somewhat arbitrarily)

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1168,24 +1168,25 @@ void Ekf::checkRangeAidSuitability()
 {
 	if (_control_status.flags.in_air
 	    && _range_sensor.isHealthy()
-	    && isTerrainEstimateValid()
-	    && isHorizontalAidingActive()) {
+	    && isTerrainEstimateValid()) {
 		// check if we can use range finder measurements to estimate height, use hysteresis to avoid rapid switching
 		// Note that the 0.7 coefficients and the innovation check are arbitrary values but work well in practice
-		const bool is_in_range = _is_range_aid_suitable
-					 ? (_terrain_vpos - _state.pos(2) < _params.max_hagl_for_range_aid)
-					 : (_terrain_vpos - _state.pos(2) < _params.max_hagl_for_range_aid * 0.7f);
+		const float range_hagl = _terrain_vpos - _state.pos(2);
+		const float range_hagl_max = _is_range_aid_suitable ? _params.max_hagl_for_range_aid : (_params.max_hagl_for_range_aid * 0.7f);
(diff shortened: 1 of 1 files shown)


dagar:
I was using this previously with only the horizontal aiding requirement dropped. The tilt limit was added for upstreaming, but I'm open to suggestions. I'll test the entire thing together next week.

EKF/control.cpp
```diff
@@ -1168,24 +1168,28 @@ void Ekf::checkRangeAidSuitability()
...
 
-		const bool is_hagl_stable = _is_range_aid_suitable
-					    ? ((_hagl_innov * _hagl_innov / (sq(_params.range_aid_innov_gate) * _hagl_innov_var)) < 1.0f)
-					    : ((_hagl_innov * _hagl_innov / (sq(_params.range_aid_innov_gate) * _hagl_innov_var)) < 0.01f);
+		if (isHorizontalAidingActive()) {
+			const float ground_vel = sqrtf(_state.vel(0) * _state.vel(0) + _state.vel(1) * _state.vel(1));
+			const float max_vel = _is_range_aid_suitable ? _params.max_vel_for_range_aid : (_params.max_vel_for_range_aid * 0.7f);
 
-		_is_range_aid_suitable = is_in_range && is_below_max_speed && is_hagl_stable;
+			const bool is_below_max_speed = ground_vel < max_vel;
```
bresch, line 1185:
```suggestion
			const bool is_below_max_speed = !_state.vel.xy().longerThan(max_vel);
```
So you can save a `sqrtf`

EKF/control.cpp
```diff
@@ -1168,24 +1168,28 @@ void Ekf::checkRangeAidSuitability()
...
+			_is_range_aid_suitable = is_in_range && is_hagl_stable && is_below_max_speed;
+
+		} else {
+			// otherwise tilt less than 15 degrees
+			const bool is_below_tilt_limit = _R_to_earth(2,2) < cosf(math::radians(15.f));
```
bresch, line 1190: I don't think the tilt check if useful here, there is already one in the terrain estimator and should be enough. Let's just skip the check?

bresch (changes requested):

EKF/control.cpp
dagar: Fine from my perspective, that's actually what I was already using.

EKF/control.cpp
dagar: Nice.

bresch (approved):

PR #987, opened 2021-03-03
Enable access to EKF information and warning events

priseborough:
This is a first step in enabling the various printed event messages to be removed 

See https://github.com/PX4/PX4-Autopilot/pull/17012 for PX4-Autopilot changes required to utilise this.

See https://github.com/PX4/PX4-ECL/pull/961 for addtional context.

### DIFF
--- a/ecl.h
+++ b/ecl.h
@@ -51,16 +51,6 @@ using ecl_abstime = hrt_abstime;
 #define ECL_WARN PX4_WARN
 #define ECL_ERR	 PX4_ERR
 
-#if defined(__PX4_POSIX)
-#define ECL_INFO_TIMESTAMPED(X) PX4_INFO("%llu: " X, (unsigned long long)_imu_sample_delayed.time_us)
-#define ECL_WARN_TIMESTAMPED(X) PX4_WARN("%llu: " X, (unsigned long long)_imu_sample_delayed.time_us)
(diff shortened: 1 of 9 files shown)


ecl.h
```diff
@@ -47,18 +47,14 @@
...
-#define ECL_INFO PX4_INFO
-#define ECL_WARN PX4_WARN
-#define ECL_ERR	 PX4_ERR
-
-#if defined(__PX4_POSIX)
-#define ECL_INFO_TIMESTAMPED(X) PX4_INFO("%llu: " X, (unsigned long long)_imu_sample_delayed.time_us)
-#define ECL_WARN_TIMESTAMPED(X) PX4_WARN("%llu: " X, (unsigned long long)_imu_sample_delayed.time_us)
-#define ECL_ERR_TIMESTAMPED(X) PX4_ERR("%llu: " X, (unsigned long long)_imu_sample_delayed.time_us)
+#if defined(__PX4_NUTTX)
+#  define ECL_INFO PX4_DEBUG
```
dagar, line 51: I'd still like to figure out a way to the SITL experience to be as close to identical as possible, but let's give this a try.

dagar (approved):

dagar:
Follow up idea - we could handle all the prints in the elf frontend based on flags alone, then it would be easy to print the messages per instance or add functionality to enable/disable output at runtime.

PR #989, opened 2021-03-12
EKF: use flow for vel test ratio if only active source of horizontal aiding

dagar:
This is intended to improve multi-EKF selection in scenarios where you only have flow.

I'm open to suggestions if there's a better way to do this.


bresch (approved):
Yes, that makes sense to me

PR #991, opened 2021-03-16
Terrain estimator reset counter

mortenfyhn:
Hi all!

This doesn't change any behaviour, but it keeps track of and exposes a reset counter for the terrain estimator. 

The reset counter can be used on the autopilot side to reset the terrain height setpoints in sync, which in turn can make it possible to make the terrain following / holding altitude modes more robust over uneven terrain, especially manmade and indoor environments. I intend to open a follow-up pull request on the PX4-Autopilot repo for that.


bresch (approved):

mortenfyhn:
Thank you!

PR #992, opened 2021-03-18
gps sacc: apply same minimum for EKF2 and yaw estimator

bresch:
The yaw estimator should use the same maximum GNSS speed accuracy controlled by the user-defined parameter.


priseborough:
How do we know if 0.01 is an appropriate minimum for the bank of 5 EKF yaw estimators. I would like to see some testing with RTK GPS or simulated RTK GPS accuracies using SITL.

bresch:
I tried a few times in STIL jMAVSim and did not find any issue. The bank of yaw estimators converges properly and the variances don't seem to have any numerical issue.
example log: https://logs.px4.io/plot_app?log=78810a15-04ba-4645-882e-2669a15ce968
DeepinScreenshot_select-area_20210323162607

priseborough (approved):

PR #994, opened 2021-03-29
Height source vision: Fallback to rangefinder if available

ecmnet:
This implements a fallback mechanism to RangeFinder (if existing) and then to Baro, if vision is selected as primary EKF2 height source in case of vision timeouts.

fixes https://github.com/PX4/PX4-Autopilot/issues/16858

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -787,7 +787,9 @@ void Ekf::controlHeightSensorTimeouts()
 	const bool continuous_bad_accel_hgt = isTimedOut(_time_good_vert_accel, (uint64_t)_params.bad_acc_reset_delay_us);
 
 	// check if height has been inertial deadreckoning for too long
-	const bool hgt_fusion_timeout = isTimedOut(_time_last_hgt_fuse, (uint64_t)5e6);
+	// in vision hgt mode check for vision data
+	const bool hgt_fusion_timeout = isTimedOut(_time_last_hgt_fuse, (uint64_t)5e6)  ||
+			 (_control_status.flags.ev_hgt && !isRecent(_time_last_ext_vision, 5 * EV_MAX_INTERVAL));
 
 	if (hgt_fusion_timeout || continuous_bad_accel_hgt) {
 
@@ -875,6 +877,19 @@ void Ekf::controlHeightSensorTimeouts()
 				request_height_reset = true;
 				failing_height_source = "ev";
 				new_height_source = "ev";
+
+			// Fallback to rangefinder data if available
+			} else if (_range_sensor.isHealthy()) {
+				setControlRangeHeight();
+				if (_control_status.flags.in_air && isTerrainEstimateValid()) {
+					_hgt_sensor_offset = _terrain_vpos;
+				} else if (_control_status.flags.in_air) {
+				    _hgt_sensor_offset = _range_sensor.getDistBottom() + _state.pos(2);
+			    } else {
+					_hgt_sensor_offset = _params.rng_gnd_clearance;
+				}
+				failing_height_source = "ev";
+				new_height_source = "rng";
 
 			} else if (!_baro_hgt_faulty) {
 				startBaroHgtFusion();
@@ -1061,11 +1076,13 @@ void Ekf::controlHeightFusion()
 
 	case VDIST_SENSOR_EV:
 
-		// don't start using EV data unless data is arriving frequently
+		// don't start using EV data unless data is arriving frequently, do not reset if pref mode was height
 		if (!_control_status.flags.ev_hgt && isRecent(_time_last_ext_vision, 2 * EV_MAX_INTERVAL)) {
 			fuse_height = true;
 			setControlEVHeight();
-			resetHeight();
+			if (!_control_status_prev.flags.rng_hgt) {
+				 resetHeight();
+			}
 		}
 
 		if (_control_status.flags.baro_hgt && _baro_data_ready && !_baro_hgt_faulty) {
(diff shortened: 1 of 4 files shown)


EKF/control.cpp
```diff
@@ -876,6 +878,19 @@ void Ekf::controlHeightSensorTimeouts()
...
+			// Fallback to rangefinder data if available
+			} else if (_range_sensor.isHealthy()) {
+				setControlRangeHeight();
+				if (_control_status.flags.in_air && isTerrainEstimateValid()) {
+					_hgt_sensor_offset = _terrain_vpos;
+				} else if (_control_status.flags.in_air) {
+				    _hgt_sensor_offset = _range_sensor.getDistBottom() + _state.pos(2);
+			    } else {
+					_hgt_sensor_offset = _params.rng_gnd_clearance;
+				}
```
bresch, line 890: This part of the code shouldn't be there. I'm not sure if you can add it cleanly in `resetHeight` or somewhere else, but this function should only make the reset decision, nothing more complicated.

EKF/control.cpp
```diff
@@ -787,7 +787,9 @@ void Ekf::controlHeightSensorTimeouts()
 	const bool continuous_bad_accel_hgt = isTimedOut(_time_good_vert_accel, (uint64_t)_params.bad_acc_reset_delay_us);
 
 	// check if height has been inertial deadreckoning for too long
-	const bool hgt_fusion_timeout = isTimedOut(_time_last_hgt_fuse, (uint64_t)5e6);
+	// in vision hgt mode check for vision data
+	const bool hgt_fusion_timeout = isTimedOut(_time_last_hgt_fuse, (uint64_t)5e6)  ||
+			 (_control_status.flags.ev_hgt && !isRecent(_time_last_ext_vision, 5 * EV_MAX_INTERVAL));
```
bresch, line 792: IMO, ev_hgt fusion should simply update `_time_last_hgt_fuse` as it is also another source of height fusion.

bresch (changes requested):

bresch:
It is generally a good idea and something needed but I asked a few modification because I want it to fit cleanly in the estimator. Vision is "just" another source of aiding and we should be able to follow the standard implementation. If the implementation is not possible because the estimator is too much oriented towards GNSS/baro aiding, let's rework that to make it more universal.

EKF/control.cpp
ecmnet: I tried that, but I could not get any vision timeouts then

EKF/control.cpp
ecmnet: I absolutely agree.

ecmnet:
One additional observation while testing: Terrain estimation does not work properly after falling back to rangefinder in vision based height mode. 

As an experimental workaround I added to ekf.cpp (122):

`if(!(_control_status.flags.rng_hgt && _params.vdist_sensor_type == VDIST_SENSOR_EV))
		   runTerrainEstimator();`

to switch off terrain estimation and keep the estimate constant during fallback time (which is an assumption, but works for altitude control and flow measurements).

This is the current solution:

Bildschirmfoto 2021-03-30 um 16 22 03

EKF/control.cpp
ecmnet: _hgt_sensor_offset calculation moved to resetHeight with [4602349](https://github.com/PX4/PX4-ECL/pull/994/commits/4602349b289af0c448299dd4a14b3051a68c73f6).

ecmnet:
Unit test issue addressed: https://github.com/PX4/PX4-ECL/issues/997#issue-846448344

bresch:
> Unit test issue addressed: #997 (comment)

If your PR makes a unit test fail because it actually fixes something, you can directly modify it to reflect the changes and make it pass.

EKF/control.cpp
bresch: I think the difference is that `isTimedOut(_time_last_hgt_fuse, (uint64_t)5e6)` is a timeout of 5 seconds while `5 * EV_MAX_INTERVAL` is 1 second. Is that the issue? Do you just need it to be more reactive?

EKF/vel_pos_fusion.cpp
```diff
@@ -166,10 +166,9 @@ bool Ekf::fuseVerticalPosition(const Vector3f &innov, const Vector2f &innov_gate
 	}
 
 	if (innov_check_pass) {
-		_time_last_hgt_fuse = _time_last_imu;
 		_innov_check_fail_status.flags.reject_ver_pos = false;
 		fuseVelPosHeight(innovation, innov_var(2), 5);
-
+		_time_last_hgt_fuse = _time_last_imu;
```
bresch, line 171: Please just put it back where it was to avoid unnecessary diff

ecmnet: done

bresch:
> One additional observation while testing: Terrain estimation does not work properly after falling back to rangefinder in vision based height mode.

Have you found out why it doesn't work properly anymore?

ecmnet:
> > One additional observation while testing: Terrain estimation does not work properly after falling back to rangefinder in vision based height mode.
> 
> Have you found out why it doesn't work properly anymore?

I have not investigated it deeper, but in mode `rng_hgt` terrain cannot be estimated properly at all - as far as I understood as the rangefinder is the source of distance to bottom and the source of height at the same time. I noticed, that, when falling back to range finder, terrain follows height.

EKF/control.cpp
ecmnet: This is definitely the case - but to be honest, I think I did not get a timeout at all, that's also why I played around with `_time_last_hgt_fuse` a bit.

bresch (approved):

bresch:
Thanks for your contribution @ecmnet !

EKF/control.cpp
mhkabir: I think I found why this was needed. There is a bug causing height fusion to not stop correctly (`_time_last_hgt_fuse` keeps getting updated long after data stops)

mhkabir: There was no check for `_ev_data_ready` before fusing the height meas, leading to a stale measurement being fused continuously.

PR #995, opened 2021-03-29
EKF feature options to disable features at compile time

dagar:
The EKF when used within PX4 (ekf2) is a relatively enormous 124 kB to 161 kB of flash depending on optimization level (-Os vs -O2). This can become problematic for inclusion on relatively flash constrained boards that might be found in FPV quadcopters (with GPS) or CAN peripheral nodes.

Are there certain features we can completely strip out at build time for simpler cases? Everything would still be on by default, but specific builds could then start to specialize.

Ideas
 - GPS heading fusion (requires higher end GPS most don't have)
 - airspeed fusion (maybe could be captured in something more general for fixed-wing usage?)
 - drag fusion
 - vision

### DIFF
--- a/.github/workflows/build_ekf_flags.yml
+++ b/.github/workflows/build_ekf_flags.yml
@@ -1,0 +1,31 @@
+
+name: Feature Build Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - CMAKE_OPTIONS: "-DECL_EKF_AIRSPEED_FUSION=OFF"
+          - CMAKE_OPTIONS: "-DECL_EKF_DRAG_FUSION=OFF"
+          - CMAKE_OPTIONS: "-DECL_EKF_GPS_YAW_FUSION=OFF"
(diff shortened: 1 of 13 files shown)


bresch:
We could also remove the whole terrain estimator and yaw emergency estimator on some platforms

dagar:
> We could also remove the whole terrain estimator and yaw emergency estimator on some platforms

I'll add those as options.

I was imagining the simplest case would probably still have a GPS and might benefit from the yaw estimator.

dagar:
If we move forward with this I'll extend the testing within this repository to additionally build and run testing with all flags disabled, and then all enabled individually.

AlexKlimaj:
This will be very useful on the ARK Flow and ARK GPS boards to run the ekf.

priseborough (commented):
Are we still exposing the relevant parameters in the ekf2 module? How does the user know this feature is not available if the parameter shows in the GCS?

priseborough:
Are we still exposing the relevant parameters in the ekf2 module? How does the user know this feature is not available if the parameter shows in the GCS?

dagar:
> Are we still exposing the relevant parameters in the ekf2 module?

At the moment yes, but the defines could be extended around those parameters as well (ekf2 side). Then they wouldn't even be in the build, show in QGC, etc.

bresch:
Having all those ifdefs adds a lot of noise in the code. If we want to disable functionalities at compile time, I think we should first refactor all the fusion type in separate classes/files that we can then en-/dis-able cleanly

PR #996, opened 2021-03-30
EKF: allow initialising without baro depending on configuration

dagar:
Attempted followup to https://github.com/PX4/PX4-ECL/pull/931.

 - every potential height source maintains a filtered value at the delayed time horizon
 - in Ekf::initialiseFilter() only the configured height source is used, but valid height data is required at init time
 - height resets use filtered delayed time horizon data rather than most recent measurement

### DIFF
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -164,6 +164,24 @@ void EstimatorInterface::setMagData(const magSample &mag_sample)
 		_mag_data_sum.setZero();
 		_mag_timestamp_sum = 0;
 	}
+
+	{
+		magSample sample_delayed;
+
+		if (_mag_buffer.peek_first_older_than(_imu_sample_delayed.time_us, &sample_delayed)) {
+			if ((sample_delayed.time_us != 0) && (sample_delayed.time_us > _mag_sample_delayed_time_us_last)) {
+				if (_mag_counter == 0) {
+					_mag_lpf.reset(sample_delayed.mag);
+
+				} else {
+					_mag_lpf.update(sample_delayed.mag);
+				}
+
+				_mag_counter++;
+				_mag_sample_delayed_time_us_last = sample_delayed.time_us;
+			}
+		}
+	}
 }
 
 void EstimatorInterface::setGpsData(const gps_message &gps)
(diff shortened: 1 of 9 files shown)


EKF/estimator_interface.h
```diff
@@ -394,6 +394,26 @@ class EstimatorInterface
 	warning_event_status_u _warning_events{};
 	information_event_status_u _information_events{};
 
+	uint64_t _baro_sample_delayed_time_us_last{0};
+	uint32_t _baro_hgt_counter{0};
+	AlphaFilter<float> _baro_hgt_delayed{0.1f};
```
bresch, line 399:
```suggestion
	AlphaFilter<float> _baro_hgt_delayed_lpf{0.1f};
```
I would like to have the `_lpf` prefix to make it obvious that this is not just holding "baro_sample_delayed"

bresch: same for the other ones

EKF/estimator_interface.h
```diff
@@ -394,6 +394,26 @@ class EstimatorInterface
...
+	uint32_t _range_hgt_counter{0};
+	AlphaFilter<float> _range_hgt_delayed{0.1f};
+
+	uint64_t _ev_sample_delayed_time_us_last{0};
+	uint32_t _ev_hgt_counter{0};
+	AlphaFilter<float> _ev_hgt_delayed{0.1f};
+
+	uint64_t _mag_sample_delayed_time_us_last{0};
+	uint32_t _mag_counter{0};
+	AlphaFilter<Vector3f> _mag_lpf{0.1f}; ///< filtered magnetometer measurement for instant reset (Gauss)
```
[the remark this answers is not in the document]
bresch: or remove the "delayed" from the other ones. Since the whole EKF is delayed, I'm not sure if it's useful to specify it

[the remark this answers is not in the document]
dagar: It was due to initial confusion with some areas of the code base grabbing the newest sample. Should be fixed now.

priseborough:
Can we do a test where the board is started without the baro driver?

EKF/estimator_interface.cpp
```diff
@@ -332,6 +388,25 @@ void EstimatorInterface::setRangeData(const rangeSample &range_sample)
...
+		// accumulate enough height measurements to be confident in the quality of the data
+		rangeSample sample_delayed;
+
+		if (_range_buffer.peek_first_older_than(_imu_sample_delayed.time_us, &sample_delayed)) {
+			if ((sample_delayed.time_us != 0) && (sample_delayed.time_us > _range_sample_delayed_time_us_last)) {
+				if (_range_hgt_counter == 0) {
+					_range_hgt_lpf.reset(sample_delayed.rng);
+
+				} else {
+					_range_hgt_lpf.update(sample_delayed.rng);
```
priseborough, line 402: This is being used anywhere at the moment. The reset function is using the latest measurement which is good for in flight when climbing or descending quickly. We could use the LPF value on startup, but for this type of sensor the best option is a median filter to get rid of spikes.

dagar: I'll add a median filter.

dagar: I've updated `resetHeight()` to use the median filtered range data.

priseborough:
There's a lot of code here for the ability to initialise height without a baro. Can we split this into one PR into what is required for initialisation without baro and a second PR with other improvements, eg median filtering. For example we could initialise height to a fake measurement of zero whilst awaiting the selected height sensor to pass checks.

PR #998, opened 2021-03-31
Terrain estimation when height estimation is based on range finder data

ecmnet:
When height estimation is based on range finder data, the terrain estimation must not use a range finder as source. Either terrain estimation needs to be stopped or uses a different source. This is an open PR to find an appropriate and generic solution to this issue. 

As a first step, terrain estimation is paused as soon as a fallback from vision based to range finder based height estimation occurs. Terrain estimation is continued when vision data is available again. However, this workaround is not generic enough as it only covers the specific use case described.

Refer to https://github.com/PX4/PX4-ECL/pull/994#issuecomment-810309127

### DIFF
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -119,8 +119,9 @@ bool Ekf::update()
 		// control fusion of observation data
 		controlFusionModes();
 
-		// run a separate filter for terrain estimation
-		runTerrainEstimator();
+		// run a separate filter for terrain estimation if not in vision fallback with rangefinder
+		if(!(_control_status.flags.rng_hgt && _params.vdist_sensor_type == VDIST_SENSOR_EV))
+			runTerrainEstimator();
 
 		updated = true;
 
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -119,8 +119,11 @@ bool Ekf::update()
 		// control fusion of observation data
 		controlFusionModes();
 
-		// run a separate filter for terrain estimation if not in vision fallback with rangefinder
-		if(!(_control_status.flags.rng_hgt && _params.vdist_sensor_type == VDIST_SENSOR_EV)) {
+		if(_control_status.flags.rng_hgt) {
+		    // do not update hagl in range finder mode, but keep it valid
+			_time_last_hagl_fuse = _time_last_imu;
+		} else {
+		    // run a separate filter for terrain if other height sources available
 			runTerrainEstimator();
 		}
 
--- a/EKF/ekf.cpp
+++ b/EKF/ekf.cpp
@@ -120,8 +120,9 @@ bool Ekf::update()
 		controlFusionModes();
 
 		// run a separate filter for terrain estimation if not in vision fallback with rangefinder
-		if(!(_control_status.flags.rng_hgt && _params.vdist_sensor_type == VDIST_SENSOR_EV))
+		if(!(_control_status.flags.rng_hgt && _params.vdist_sensor_type == VDIST_SENSOR_EV)) {
 			runTerrainEstimator();
+		}
 
 		updated = true;
 


EKF/ekf.cpp
```diff
@@ -119,8 +119,9 @@ bool Ekf::update()
 		// control fusion of observation data
 		controlFusionModes();
 
-		// run a separate filter for terrain estimation
-		runTerrainEstimator();
+		// run a separate filter for terrain estimation if not in vision fallback with rangefinder
+		if(!(_control_status.flags.rng_hgt && _params.vdist_sensor_type == VDIST_SENSOR_EV))
```
dagar, line 123: Can you add braces?

dagar, line 123:
```suggestion
		if(!(_control_status.flags.rng_hgt && _params.vdist_sensor_type == VDIST_SENSOR_EV)) {
```

ecmnet: Sure

ecmnet:
To avoid drift during fallbacks, terrain estimator should not run for modes where a valid estimation is not possible:

- Rangefinder based height 
- intermediate states where height estimation purely relies on ACC

ecmnet:
@mhkabir, @bresch : Your comments?

priseborough:
I agree that we should not be running the terrain estimator if ground is being used as the height datum for the EKF, ie using range finder. Do you have test evidence that this fixes the issue reported? The change looks sensible.

ecmnet:
The change is tested in the context of vision as primary source to survive reset cycles of the vision subsystem.

ecmnet:
Further tests have shown that this does not work in the case when vision fails before terrain is estimated the first time (initialization?). Any suggestions how to solve this?

priseborough:
> Further tests have shown that this does not work in the case when vision fails before terrain is estimated the first time (initialization?). Any suggestions how to solve this?

@ecmnet Are you able to provide the steps you used to create the input data that resulted in the failure case? I would like to make a unit test for corner cases like this.

ecmnet:
I tested this with a live vehicle using odometry and a rangefinder with propellers unmounted. After simulating a takeoff and waiting for valid terrain estimation, I switched off vision observing z estimation following raw lidar data.

dagar:
ECL has been absorbed into PX4-Autopilot, can you recreate this PR over there?

https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/ecl

ecmnet:
Will do as soon as I get a running version based on v1.12.0. Currently I have the issue, that 'freezing' the ground level takes place too late (about 5secs). I also want to adjust initialisation in the case that the initial primary height source is a range finder .

ecmnet:
> ECL has been absorbed into PX4-Autopilot, can you recreate this PR over there?
> 
> https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/ecl

Link is not working

dagar:
@ecmnet I initially merged it into `src/lib/ecl` (where the submodule lived), then later moved it to within `src/modules/ekf2/EKF`.

https://github.com/PX4/PX4-Autopilot/tree/master/src/modules/ekf2/EKF

ecmnet:
Further tests show, that there is a dependency to `Ekf::controlHeightSensorTimeouts()`: Within the phase the selected height sensor idoes not provide any data and the detection of the timeout, terrain estimation is continued and relies e.g. on RF. As no height source is available, Z position is considered to be (more or less) constant. If the vehicle actually moves in Z direction during this phase, this movement is falsely reflected as movement of the terrain altitude. With the detection of the timeout, the terrain altitude is freezed. The freezed altitude in this case is not correct, depending on the selected timeout time and the speed of the vehicle in Z-direction.

Any suggestions how to solve this?

priseborough:
There may reasons to use range finder for terrain offset and EKF height (assuming terrain is at zero) simultaneously due to differing requirements for filtering out the effect of terrain height variation. We will move this to the PX4-Autopilot issues list and continue the investigation.

PR #999, opened 2021-04-01
Allow RangeFinder fusion in vision height mode (Fix for #994)

ecmnet:
Although `controlHeightSensorTimeouts()` switched to` rng_hgt` mode correctly as soon as an ev timeout happens, range finder height fusion is not performed as `rng_hgt` is not allowed if vision is the primary height source. This PR allows range finder height fusion even if vision is selected as primary height source.

Relates to https://github.com/PX4/PX4-ECL/pull/994

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1079,20 +1079,13 @@ void Ekf::controlHeightFusion()
 			}
 		}
 
-		if (_control_status.flags.rng_hgt && _range_sensor.isDataHealthy()) {
-			// Allow RangeFinder Fusion
-			fuse_height = true;
-		}
-
-		if (_control_status.flags.baro_hgt && _baro_data_ready && !_baro_hgt_faulty) {
-			// switch to baro if there was a reset to baro
-			fuse_height = true;
-		}
-
-		// determine if we should use the vertical position observation
-		if (_control_status.flags.ev_hgt) {
-			fuse_height = true;
-		}
+		if (_control_status.flags.ev_hgt && _ev_data_ready) {
+		     fuse_height = true;
+		} else if (_control_status.flags.rng_hgt && _range_sensor.isDataHealthy()) {
+		      fuse_height = true;
+		} else if (_control_status.flags.baro_hgt && _baro_data_ready && !_baro_hgt_faulty) {
+		      fuse_height = true;
+	    }
 
 		break;
 	}
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1079,6 +1079,11 @@ void Ekf::controlHeightFusion()
 			}
 		}
 
+		if (_control_status.flags.rng_hgt) {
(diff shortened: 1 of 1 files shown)


EKF/control.cpp
```diff
@@ -1079,6 +1079,11 @@ void Ekf::controlHeightFusion()
 			}
 		}
 
+		if (_control_status.flags.rng_hgt) {
```
dagar, line 1082: Only fuse height if _range_sensor.isDataHealthy()?

dagar, line 1082:
```suggestion
		if (_control_status.flags.rng_hgt && _range_sensor.isDataHealthy()) {
```

ecmnet:
@bresch Could you have a look? #994 does not work without this fix.

mhkabir:
A little late to the party here (missed #994 for a while), so please humor me. 

How does fallback from vision to range work, if there is an offset between the vision Z and rangefinder Z? Is the offset compensated for before starting range fusion?

I'm also wondering if would it be better to fall back to range-aided baro instead of pure rangefinder in case flying over ground with obstacles/non-flat ground.

ecmnet:
> A little late to the party here (missed #994 for a while), so please humor me.
> 
> How does fallback from vision to range work if there is an offset between the vision Z and rangefinder Z? Is the offset compensated for before starting range fusion?
> 
> I'm also wondering if would it be better to fall back to range-aided baro instead of pure rangefinder in case flying over ground with obstacles/non-flat ground.

The offset is compensated in #994 already. Purpose of the fallback is to support vision reset cycles / drop outs (e.g. direct sun, etc.) - therefore the assumption of a flat ground seems to be sufficient.

mhkabir (approved):

EKF/control.cpp
```diff
@@ -1079,6 +1079,11 @@ void Ekf::controlHeightFusion()
 			}
 		}
 
+		if (_control_status.flags.rng_hgt && _range_sensor.isDataHealthy()) {
```
mhkabir, line 1082: Can you replace 1082 - 1095 with this block?
```
               if (_control_status.flags.ev_hgt && _ev_data_ready) {
                       fuse_height = true;
               } else if (_control_status.flags.rng_hgt && _range_sensor.isDataHealthy()) {
                       fuse_height = true;
               } else if (_control_status.flags.baro_hgt && _baro_data_ready && !_baro_hgt_faulty) {
                        fuse_height = true;
               }

```

This groups the statements together like for GPS (few lines above), and adds a check for `_ev_data_ready` which fixes the issue you saw here https://github.com/PX4/PX4-ECL/pull/994#discussion_r621816867.

After that you can also revert the change you made on line 792 in #994, and timeouts will work as designed.

mhkabir (changes requested):
See inline comment

mhkabir (approved):
Perfect! Tested via backport and works.

PR #1001, opened 2021-04-06
Allow global position to be published after setting origin

ecmnet:
Sending a set_gps_global_origin message to EKF2 via MAVLINK sets the NED origin as reported by EKF2 but does not switch to publish global position. This would be helpful in use cases where no GPS is available but LPOS is provided by other means.
Reason for this is, that the flag _NED_origin_initialised is not set to true inEkf::setEkfGlobalOrigin. 

Doing so in elf_helper.cpp would work as long as the parameter EKF2_DECL_TYPE is set to 0. Otherwise an invalid look up of the magnetic declination results an invalid EKF2_MAG_DECL value.

### DIFF
--- a/EKF/ekf_helper.cpp
+++ b/EKF/ekf_helper.cpp
@@ -719,6 +719,8 @@ bool Ekf::setEkfGlobalOrigin(const double latitude, const double longitude, cons
 		map_projection_reproject(&_pos_ref, _state.pos(0), _state.pos(1), &current_lat, &current_lon);
 		current_alt = -_state.pos(2) + _gps_alt_ref;
 		current_pos_available = true;
+	} else {
+		_NED_origin_initialised = true;
 	}
 
 	// reinitialize map projection to latitude, longitude, altitude, and reset position
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -1079,20 +1079,13 @@ void Ekf::controlHeightFusion()
 			}
 		}
 
-		if (_control_status.flags.rng_hgt && _range_sensor.isDataHealthy()) {
-			// Allow RangeFinder Fusion
-			fuse_height = true;
-		}
-
-		if (_control_status.flags.baro_hgt && _baro_data_ready && !_baro_hgt_faulty) {
-			// switch to baro if there was a reset to baro
-			fuse_height = true;
-		}
-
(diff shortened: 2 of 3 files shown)


dagar:
This is probably sufficient for initial testing, but ultimately I think we should also handle getting valid GPS after it's been set manually.

https://github.com/PX4/PX4-ECL/blob/5d34d7a24ef72b826c320a3259ee0ec68b1936df/EKF/gps_checks.cpp#L82-L108

priseborough:
It appears that the only change required to achieve the stated objective of this PR are the two lines in ekf_helper.cpp. Can you please remove other fixes not required for the publishing of global position and submit them separately

EKF/ekf_helper.cpp
```diff
@@ -719,6 +719,8 @@ bool Ekf::setEkfGlobalOrigin(const double latitude, const double longitude, cons
 		map_projection_reproject(&_pos_ref, _state.pos(0), _state.pos(1), &current_lat, &current_lon);
 		current_alt = -_state.pos(2) + _gps_alt_ref;
 		current_pos_available = true;
+	} else {
+		_NED_origin_initialised = true;
```
bresch, line 723: I wouldn't set the flag here because the meaning of that `else` is: "if the map is not already initialized" and not if we actually initialized the map with the new origin.
I would suggest to place it above the `return true` below, as this is the place where we know that the map has been initialized correctly with the new origin.

bresch (changes requested):

atyshka:
Any progress on making the changes to get this merged? Would very much appreciate this feature for indoor positioning

dagar:
ECL has been absorbed into PX4-Autopilot, can you recreate this PR over there?

https://github.com/PX4/PX4-Autopilot/tree/master/src/lib/ecl

ecmnet:
I just did a short test with V1.12.0 - it still works fine. Will create a new PR.

priseborough:
We need to be careful that there are not consumers of the global position message that interpret availability of the global position message as indicating validity.

priseborough:
The https://github.com/PX4/PX4-Autopilot/blob/master/msg/vehicle_global_position.msg message has no validity flag for LLH so consumers will assume valid for controller use if present.

TODO: check how dead_reckoning flag is being used.

PR #1005, opened 2021-05-02
EKF: increase allowed difference between flow and gyro ODRs

mhkabir:
In the normal and lowlight mode for the PAW3902/3 sensor, the ODR is 66Hz, while in super lowlight mode it is 50Hz. See plots for the switching in a typical flight.

The IMU ODR vs flow ODR mismatch threshold was too conservative for the 66Hz ODR, since https://github.com/PX4/PX4-ECL/pull/928. The EKF would keep waiting for the data to line up and never fuse measurements.

image
image

### DIFF
--- a/EKF/control.cpp
+++ b/EKF/control.cpp
@@ -361,8 +361,8 @@ void Ekf::controlOpticalFlowFusion()
 		const bool is_magnitude_good = !_flow_sample_delayed.flow_xy_rad.longerThan(_flow_sample_delayed.dt * _flow_max_rate);
 		const bool is_tilt_good = (_R_to_earth(2, 2) > _params.range_cos_max_tilt);
 
-		const float delta_time_min = fmaxf(0.8f * _delta_time_of, 0.001f);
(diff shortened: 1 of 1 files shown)


bresch (approved):
That makes sense, the dt_flow is 15ms @66Hz and two imu samples make a dt of 200ms @100Hz. 0.8 * 200 = 160ms  and 0.7 * 200 = 140ms.

bresch:
@mhkabir While reviewing this, I ended up checking how the gyro compensation was computed and I found something that should be improved:
When there is no gyro attached with the flow sensor, ekf2 integrates the gyro delta_angle to produce the value in `_flow_sample_delayed.gyro_xyz`. Then, when doing the compensation, the angular rate is obtained by dividing using `_flow_sample_delayed.dt` which is the integration time of the flow sensor:
https://github.com/PX4/PX4-ECL/blob/aad48407c21f056550de343a5b0da1c02feb65b9/EKF/optflow_fusion.cpp#L76

However, as you mentioned in this PR, they are often not the same and this could lead to large errors in the angular rate compensation.

PR #1006, opened 2021-05-04
GNSS yaw fusion rework and failsafes

bresch:
This PR reorganizes and adds more failure cases handling for the GNSS heading fusion.

Main design points:
- to start the fusion, the conditions to continue the fusion must be met + some extra ones
- to continue the fusion, some high-level metrics must pass
- the fusion is disabled immediately when the conditions to continue the fusion are not met anymore

The fusion can also be stopped if:
- the GNSS heading data is not received for some time (the fusion can restart when the starting condition are met again)
- the data is constantly rejected by the estimator for some time: 
   - if it is the first time since the condition started -> attempt to reset (only 1 reset is allowed)
   - if a reset already occurred and the starting checks are passing, something else is probably wrong -> never start the fusion again
   - if a reset already occurred and the starting checks are **not** currently passing, it might be a temporary outage -> disable the fusion but let it start again if the starting conditions are passing

Because a diagram is often easier to understand:
GNSS_heading

Unit tests:
- some duration had to be increased because more checks need to pass and the timeouts are based on `reset_timeout_max` (7 seconds)
- reset case covered

Remark: I tried to come up with a high-level generic approach that could be re-usable for all aiding sources. My hope is that all the aiding sources use the same structure so once we're happy with this refactor, I will try to adapt the "control" code of another aiding source to see if a common pattern can be used.


EKF/control.cpp
```diff
@@ -744,30 +747,69 @@ void Ekf::controlGpsYawFusion()
...
+
+		const bool continuing_conditions_passing = !gps_checks_failing;
+
+		const bool is_gps_yaw_data_intermittent = !isRecent(_time_last_gps_yaw_data, 2 * GPS_MAX_INTERVAL);
+
+		const bool starting_conditions_passing = continuing_conditions_passing
+		                                         && _control_status.flags.tilt_align
+		                                         && gps_checks_passing
+							 && !is_gps_yaw_data_intermittent
+		                                         && !_gps_hgt_intermittent;
```
priseborough, line 762: Any specific reason to check height for intermittency here?

EKF/control.cpp
```diff
@@ -744,30 +747,69 @@ void Ekf::controlGpsYawFusion()
...
-	// Check if the data is constantly fused by the estimator,
-	// if not, declare the sensor faulty and stop the fusion
-	// By doing this, another source of yaw aiding is allowed to start
-	if (_control_status.flags.gps_yaw
-	    && isTimedOut(_time_last_gps_yaw_fuse, (uint64_t)5e6)) {
-		_is_gps_yaw_faulty = true;
+	} else if (_control_status.flags.gps_yaw
+	           && isTimedOut(_time_last_gps_yaw_data, _params.reset_timeout_max)) {
+		// No yaw data in the message anymore. Stop until it comes back.
 		stopGpsYawFusion();
```
priseborough, line 813: This is an interesting case and one that can persist for some time when on ground. I have seen logs where the GNSS yaw is acquired, glitches, then stops for up to a minute until the RTK functionality is restored. The code looks like it should fuse _last_static_yaw if _control_status.flags.vehicle_at_rest is true to prevent uncontrolled yaw drift and enable yaw gyro bias learning in this period. This is something we should check when looking at a test log.

priseborough (commented):
Is it possible to get a HW test log to go with this? If you have a copter with a dual antenna setup, you might be able to force a loss of yaw by covering or disconnecting one of the antennas. I would like check that we are fusing _last_static_yaw where possible or a zero innovation as required to prevent uncontrolled covariance growth as loss of RTK yaw when on ground can persist for some time.

EKF/control.cpp
bresch: This is just bad variable naming; `_gps_hgt_intermittent` should just be `_gps_intermittent`

priseborough:
I tried to set up a simple SITL test to verify fusion if static yaw with a log that could be replayed, but have run into an issue where SITL logs generated with SDLOG_PROFILE = 3, EKF2_MULTI_MAG=0, EKF2_MULTI_ IMU = 0, SENS_IMU_MODE =1 and SENS_MAG_MODE = 1 to disable multi EKF generated the following error message on replay:

ERROR [replay] unknown log definition type 81, size 26 (offset 52267)

EKF/control.cpp
bresch: If the mag fusion is set to NONE, then it will enter this case and fuse the last static yaw if on ground:
https://github.com/PX4/PX4-ECL/blob/29243ac5cbb5d27ac71744e88afcd786df6f748d/EKF/mag_control.cpp#L49-L59

Otherwise, it will simply start the desired mag fusion mode

bresch: I added a unit test for this specific case here: https://github.com/PX4/PX4-ECL/pull/1006/commits/400b66d76bee133d50acf0343b7af489a0c1b74a

bresch: However, if there is no mag data and the `MAG_FUSE_TYPE` is not set to `NONE` (i.e.: auto, 3d or heading), the yaw variance is not constrained. This is maybe something we need to address as well in a following PR

PR #1017, opened 2021-06-18
delay time: potentially reduce delay time if airspeed is not used

BazookaJoe1900:
it seems that if some observes are not used, still their delay time parameters are used and might increase the delay buffers.

this pr should check if observes are used and if not, ignore their delay parameters.

1. airspeed - seems that if param_ekf2_arsp_thr is >0 it can be used
2. baro - I couldn't find (if) there is parameter that can imply it used or not. so there is still `if(??)` on the code !
3. aux velocity -  I couldn't find (if) there is parameter that can imply it used or not. so there is still `if(??)` on the code !

### DIFF
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -277,7 +277,7 @@ struct parameters {
 	// airspeed fusion
 	float tas_innov_gate{5.0f};		///< True Airspeed innovation consistency gate size (STD)
 	float eas_noise{1.4f};			///< EAS measurement noise standard deviation used for airspeed fusion (m/s)
-	float arsp_thr{0.0f};			///< Airspeed fusion threshold. A value of zero will deactivate airspeed fusion
+	float arsp_thr{2.0f};			///< Airspeed fusion threshold. A value of zero will deactivate airspeed fusion
 
 	// synthetic sideslip fusion
 	float beta_innov_gate{5.0f};		///< synthetic sideslip innovation consistency gate size in standard deviation (STD)
--- a/EKF/common.h
+++ b/EKF/common.h
@@ -277,6 +277,7 @@ struct parameters {
 	// airspeed fusion
 	float tas_innov_gate{5.0f};		///< True Airspeed innovation consistency gate size (STD)
 	float eas_noise{1.4f};			///< EAS measurement noise standard deviation used for airspeed fusion (m/s)
+	float arsp_thr{0.0f};			///< Airspeed fusion threshold. A value of zero will deactivate airspeed fusion
 
 	// synthetic sideslip fusion
 	float beta_innov_gate{5.0f};		///< synthetic sideslip innovation consistency gate size in standard deviation (STD)
--- a/EKF/estimator_interface.cpp
+++ b/EKF/estimator_interface.cpp
@@ -478,9 +478,14 @@ void EstimatorInterface::setDragData(const imuSample &imu)
 bool EstimatorInterface::initialise_interface(uint64_t timestamp)
 {
 	// find the maximum time delay the buffers are required to handle
-	float max_time_delay_ms = math::max(_params.baro_delay_ms,
-					math::max(_params.auxvel_delay_ms,
-						_params.airspeed_delay_ms));
+	// it's reasonable to assume that barometer is always used, and its delay is low
+	// it's reasonable to assume that aux velocity device has low delay. TODO: check the delay only if the aux device is used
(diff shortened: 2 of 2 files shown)


EKF/estimator_interface.cpp
```diff
@@ -478,9 +478,21 @@ void EstimatorInterface::setDragData(const imuSample &imu)
 bool EstimatorInterface::initialise_interface(uint64_t timestamp)
 {
 	// find the maximum time delay the buffers are required to handle
-	float max_time_delay_ms = math::max(_params.baro_delay_ms,
-					math::max(_params.auxvel_delay_ms,
-						_params.airspeed_delay_ms));
+	float max_time_delay_ms = 0.f;
+	// using Barometric altitude
+	if (??) {
```
bresch, line 483: The default baro delay is 0ms and it will always be small so I don't think it's worth adding a check for it.

EKF/estimator_interface.cpp
```diff
@@ -478,9 +478,21 @@ void EstimatorInterface::setDragData(const imuSample &imu)
...
+		max_time_delay_ms = math::max(_params.baro_delay_ms, max_time_delay_ms);
+	}
+
+	// using aux vel
+	if (??) {
```
bresch, line 488: The best way here would be to add a en-/disable bit in the EKF2_AID_MASK. Currently, the control logic for the "aux vel" fusion is really crude (no start/stop, health check, ...).

bresch (changes requested):
Since the default delay values of baro and aux vel are really small (0ms and 5ms, respectively), I would suggest that we just merge the airspeed check (100ms) and ignore the other ones for now.

EKF/estimator_interface.cpp
BazookaJoe1900: @bresch - Done

EKF/estimator_interface.cpp
BazookaJoe1900: @bresch - Done

BazookaJoe1900:
@bresch - your notes has implemented.
can you re-check?

EKF/estimator_interface.cpp
```diff
@@ -478,9 +478,14 @@ void EstimatorInterface::setDragData(const imuSample &imu)
 bool EstimatorInterface::initialise_interface(uint64_t timestamp)
 {
 	// find the maximum time delay the buffers are required to handle
-	float max_time_delay_ms = math::max(_params.baro_delay_ms,
-					math::max(_params.auxvel_delay_ms,
-						_params.airspeed_delay_ms));
+	// its reasonable to assume that barometer is always used, and its delay is low
+	// its reasonable to assume that aux velocity device has low delay. TODO: check the delay only if the aux device is used
```
bresch, line 482:
```suggestion
	// it's reasonable to assume that barometer is always used, and its delay is low
	// it's reasonable to assume that aux velocity device has low delay. TODO: check the delay only if the aux device is used
```

bresch (changes requested):
Alright, thanks!

bresch (changes requested):
We have to make sure there is a corresponding Firmware PR to set `arsp_thr`, otherwise we'll have issues with all the platforms that have an airspeed sensor. To avoid those issues, please initialize the `arsp_thr` to something like 2 m/s in order keep the same behavior until the parameter value is sent by the firmware.

BazookaJoe1900:
there is need to add EKF2_ARSP_THR to be be ParamExtFloat here:

https://github.com/PX4/PX4-Autopilot/blob/a728431e4350917f3add9814732715101ac99b4d/src/modules/ekf2/EKF2.hpp#L459

so after @priseborough will look at that too, I will do the change there too

EKF/common.h
```diff
@@ -277,6 +277,7 @@ struct parameters {
 	// airspeed fusion
 	float tas_innov_gate{5.0f};		///< True Airspeed innovation consistency gate size (STD)
 	float eas_noise{1.4f};			///< EAS measurement noise standard deviation used for airspeed fusion (m/s)
+	float arsp_thr{0.0f};			///< Airspeed fusion threshold. A value of zero will deactivate airspeed fusion
```
bresch, line 280:
```suggestion
	float arsp_thr{2.0f};			///< Airspeed fusion threshold. A value of zero will deactivate airspeed fusion
```

We don't want to change the default behavior.

bresch (changes requested):

bresch (approved):

priseborough (approved):

